PR #25, opened 2023-09-07
Initial implementation of MLClient and workflow interface

owaiskazi19:
### Description
Adds initial implementation of MLClient and an interface for the building block of the workflow setup. Modified build.gradle to handle snapshot version of the plugin and added dependency of MLClient from ml-commons.

Also, created plugin-security.policy for granting permissions for MLClient.

### Issues Resolved
Fixes https://github.com/opensearch-project/opensearch-ai-flow-framework/issues/20

### DIFF
--- a/src/main/java/org/opensearch/flowframework/workflow/Workflow.java
+++ b/src/main/java/org/opensearch/flowframework/workflow/Workflow.java
@@ -1,0 +1,22 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+package org.opensearch.flowframework.workflow;
+
+/**
+ * Interface for the workflow setup of different building blocks.
+ */
+public interface Workflow {
+
+    /**
+     * Triggers the processing of the building block.
+     * @throws Exception if execution fails
+     */
+    void process() throws Exception;
+
+}
--- a/src/main/java/org/opensearch/flowframework/workflow/Workflow.java
+++ b/src/main/java/org/opensearch/flowframework/workflow/Workflow.java
@@ -8,6 +8,8 @@
  */
 package org.opensearch.flowframework.workflow;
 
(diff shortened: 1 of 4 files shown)


src/main/java/org/opensearch/flowframework/workflow/Workflow.java
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * Interface for the workflow setup of different building blocks.
+ */
+public interface Workflow {
+
+    /**
+     * Triggers the processing of the building block.
+     * @throws Exception if execution fails
+     */
+    void process() throws Exception;
```
owaiskazi19, line 20: @dbwiddis need input on the interface for the building block. I envision that the every building block let's UploadModel would implement this interface something like:

```java
public class UploadModel implements Workflow {

    @Override
    public void process() {
       mlClient.uploadModel()
    }
}

dbwiddis: 1. We probably want a return value of type `CompletableFuture<Workflow>`
2. I'm not overly fond of the name "process" here, as I'm thinking a workflow is a chaining of processes.   I'd prefer `execute()` or possibly  `doExecute()`.
3. I'm working on process chaining in a separate branch with a PR soon.  I don't expect my design to be final either, but let's not let this discussion hold up progress, we'll see the connections and merge/rename the approaches in a later refactor.

dbwiddis (approved):

src/main/java/org/opensearch/flowframework/workflow/Workflow.java
owaiskazi19: Thanks for the suggestions. I have changed the method name to `execute` because `doExecute()` would create little confusion with `prepareRequest` -> `doExecute()`. Also, I have added the return type as `CompletableFuture<Workflow>`.

dbwiddis: Great!  Your comments on my PR #26 on how I can interface with your `execute()` from my code would be helpful.  OK to merge this one, we can always tweak later.

joshpalis (approved):

PR #26, opened 2023-09-08
Topological Sorting and Sequenced Execution

### LINKED ISSUE #17: [FEATURE] Parse a JSON Directed Graph into a process sequence

A portion of the user template will contain nodes and edges representing a process execution sequence.  This needs to be parsed into the actual executable sequence.

Use [Topological Sorting](https://en.wikipedia.org/wiki/Topological_sorting) to place objects representing the nodes into an ordered sequence.  The nodes must include references to predecessor nodes in the graph.  They should include references to dependent nodes if feasible.

The node predecessors will be used to delay execution using an `allOf()` completable future.

### LINKED ISSUE #19: [FEATURE] Provide asynchronous process execution API

In https://github.com/opensearch-project/opensearch-ai-flow-framework/issues/17, a process sequence with predecessor requirements will be generated.  Use this sequence to actually execute process steps when predecessors are complete.

An execution of the process will complete the following steps:
1. Iterate through the entire sequence, calling a method that returns a completable future immediately.
2. The method will take a collection of predecessor step completable futures (generated in step 1) as an argument, and delay further execution until an an allOf() completable future is complete.
3. When the step is actually complete, complete the future, which will be used by a future step.

This is the same type of execution currently used in plugins to execute a sequence of events.  It just needs to be adapted to rely on multiple predecessor steps.

### LINKED ISSUE #29: [FEATURE] Improve Workflow api

I implemented the `Workflow` API committed in #25 as part of my draft PR #26 and identified a few needed improvements.

Looking up and executing these should conceptually be similar to OpenSearch `TransportAction`s.  They all have an `ActionType` (fancy wrapper for a string), and then an action taking a `Request` argument and a `Response` type, which is used internally to return the results asynchronously.  We need counterparts of these same 4 components here:
 - [x] We need a consistent way to tie a string name to a Java object. Some sort of internal name field for each class will allow us to initialize a map somewhere and look these up.
 - [x] Another `<Workflow>` is not the appropriate return type as all we can do with the returned future is execute it... we need two new interfaces here representing input and output.   I think accepting input as an argument and a completable future for the output is probably the best way to do it.
     - I'm thinking a `Map<String, Object>` is probably a good internal representation for these input/output objects (to ease going to/from JSON) but there may be a better one.  It's probably very analogous to XContent. 
 - [x] Do not require the `execute()` method to throw an `Exception`. This requires calling code to catch the exception.  We're already returning a `CompletableFuture`, just complete the future exceptionally on error.
 - [x] I'm not sure `Workflow` is the right class name as that seems to imply the highest level, while this corresponds either to individual process steps, or sub-workflows.  I think `WorkflowStep` might be a good option but there may be better ones.

I think these suggestions are open-ended enough to allow lots of alternatives, I'm just requesting functionality.

Somewhere we'll need a way of finding these java objects to iterate over them to get their names.  This requires some level of object "registration" via an extension point somewhere.  Handling via `createComponents()` is a possibility.

### LINKED ISSUE #39: [BUG] Spotless doesn't replace tabs|spaces with spaces|tabs in JSON
### What is the bug?

JSON files with tabs aren't replaced by spaces like any sane formatter should do.

Or if we prefer tabs, spaces aren't being replaced by tabs.

1. Save a JSON file with tabs (accidentally, of course) and another one with spaces in an IDE configured for 4-space indent.
2. Run `./gradlew spotlessApply`
3. Observe that no changes have occurred, and the result looks different on different editors/environments (e.g., GitHub which for some insane reason loves 8 character tab indent)

Pick one:
 - [ ] No tabs used as indentation in any files representing structured data (like YAML or JSON).
 - [ ] Tabs always used as indentation in any files representing structured data (like YAML or JSON).

### Do you have any screenshots?

Screenshot 2023-09-14 at 5 22 27 PM

https://www.reddit.com/r/programming/comments/p1j1c/tabs_vs_spaces_vs_both/

dbwiddis:
### Description

Uses [Topological Sorting](https://en.wikipedia.org/wiki/Topological_sorting#Kahn's_algorithm) to parse a JSON graph description of nodes and edges into a sequence of processes.   Uses asynchronous execution with CompletableFutures to execute the processes.

Example input:
```json
{
    "sequence": {
        "nodes": [
            {
                "id": "fetch_model"
            },
            {
                "id": "create_ingest_pipeline"
            },
            {
                "id": "create_search_pipeline"
            },
            {
                "id": "create_neural_search_index"
            }
        ],
        "edges": [
            {
                "source": "fetch_model",
                "dest": "create_ingest_pipeline"
            },
            {
                "source": "fetch_model",
                "dest": "create_search_pipeline"
            },
            {
                "source": "create_ingest_pipeline",
                "dest": "create_neural_search_index"
            },
            {
                "source": "create_search_pipeline",
                "dest": "create_neural_search_index"
            }
        ]
    }
}
```
Example output showing sequencing with parallel paths:
```
17:22:34.597 INFO  o.o.f.demo.Demo - Parsing graph to sequence...
17:22:34.634 DEBUG o.o.f.template.TemplateParser - Start node(s): [fetch_model]
17:22:34.635 DEBUG o.o.f.template.TemplateParser - Execution sequence: [fetch_model, create_search_pipeline, create_ingest_pipeline, create_neural_search_index]
17:22:34.636 INFO  o.o.f.demo.Demo - Queueing process [fetch_model]. Can start immediately!
17:22:34.638 INFO  o.o.f.demo.Demo - Queueing process [create_search_pipeline]. Must wait for [fetch_model] to complete first.
17:22:34.638 DEBUG o.o.f.template.ProcessNode - >>> Starting fetch_model.
17:22:34.638 INFO  o.o.f.demo.Demo - Queueing process [create_ingest_pipeline]. Must wait for [fetch_model] to complete first.
17:22:34.639 INFO  o.o.f.demo.Demo - Queueing process [create_neural_search_index]. Must wait for [create_ingest_pipeline, create_search_pipeline] to complete first.
17:22:37.640 DEBUG o.o.f.template.ProcessNode - <<< Completed fetch_model
17:22:37.640 DEBUG o.o.f.template.ProcessNode - >>> Starting create_ingest_pipeline.
17:22:37.640 DEBUG o.o.f.template.ProcessNode - >>> Starting create_search_pipeline.
17:22:40.644 DEBUG o.o.f.template.ProcessNode - <<< Completed create_ingest_pipeline
17:22:42.644 DEBUG o.o.f.template.ProcessNode - <<< Completed create_search_pipeline
17:22:42.644 DEBUG o.o.f.template.ProcessNode - >>> Starting create_neural_search_index.
17:22:44.648 DEBUG o.o.f.template.ProcessNode - <<< Completed create_neural_search_index
17:22:44.648 INFO  o.o.f.demo.Demo - All done!
```

Example showing initialized data:
```json
{
    "sequence": {
        "nodes": [
            {
                "id": "create_index",
                "index_name": "demo"
            },
            {
                "id": "create_another_index",
                "index_name": "second_demo"
            }
        ],
        "edges": [
            {
                "source": "create_index",
                "dest": "create_another_index"
            }
        ]
    }
}
```
And the processing:
```
10:37:55.530 INFO  o.o.f.demo.DataDemo - Parsing graph to sequence...
10:37:55.594 DEBUG o.o.f.template.TemplateParser - Start node(s): [create_index]
10:37:55.595 DEBUG o.o.f.template.TemplateParser - Execution sequence: [create_index, create_another_index]
10:37:55.595 INFO  o.o.f.demo.DataDemo - Queueing process [create_index]. Can start immediately!
10:37:55.599 INFO  o.o.f.demo.DataDemo - Queueing process [create_another_index]. Must wait for [create_index] to complete first.
10:37:55.599 DEBUG o.o.f.template.ProcessNode - >>> Starting create_index.
10:37:55.684 DEBUG o.o.f.demo.CreateIndexWorkflowStep - Initialization sent params: {index=demo}, content: {aliases=[], settings={}, mappings={}}
10:37:57.699 DEBUG o.o.f.template.ProcessNode - <<< Completed create_index
10:37:57.699 DEBUG o.o.f.template.ProcessNode - >>> Starting create_another_index.
10:37:57.700 DEBUG o.o.f.demo.CreateIndexWorkflowStep - Initialization sent params: {index=second_demo}, content: {aliases=[], settings={}, mappings={}}
10:37:57.700 DEBUG o.o.f.demo.CreateIndexWorkflowStep - Previous step sent params: {}, content: {index=demo}
10:37:59.705 DEBUG o.o.f.template.ProcessNode - <<< Completed create_another_index
10:37:59.705 INFO  o.o.f.demo.DataDemo - All done!
```

### Issues Resolved

Fixes #17 
Fixes #19 
Fixes #29
Fixes #39 


src/main/java/org/opensearch/flowframework/template/ProcessNode.java
```diff
@@ -0,0 +1,94 @@
...
+            if (future.isCompletedExceptionally()) {
+                return;
+            }
+            System.out.println(">>> Starting " + this.id);
+            sleep(id.contains("ingest") ? 8000 : 4000);
+            System.out.println("<<< Finished " + this.id);
+            future.complete(this.id);
+        });
+        return this.future;
+    }
```
owaiskazi19, line 68: The `execute` method I have created in the interface is for the building block to use it like:

```java
public class UploadModel implements Workflow {

    @Override
    public void execute() {
       mlClient.uploadModel()
    }
}
```

Since this `execute` method is for process chaining, we don't have to implement the interface `Workflow` here. 
Feel free if you want to change the method's name to something else than `execute` here.

src/main/java/org/opensearch/flowframework/template/ProcessNode.java
```diff
@@ -0,0 +1,94 @@
...
+                try {
+                    waitForPredecessors.orTimeout(30, TimeUnit.SECONDS).get();
+                } catch (InterruptedException | ExecutionException e) {
+                    future.completeExceptionally(e);
+                }
+            }
+            if (future.isCompletedExceptionally()) {
+                return;
+            }
+            System.out.println(">>> Starting " + this.id);
```
owaiskazi19, line 62: Leftover?

src/main/java/org/opensearch/flowframework/template/ProcessNode.java
```diff
@@ -0,0 +1,94 @@
...
+            sleep(id.contains("ingest") ? 8000 : 4000);
```
owaiskazi19, line 63: Here as well?

src/main/java/org/opensearch/flowframework/template/TemplateParser.java
```diff
@@ -0,0 +1,168 @@
...
+            + "                \"source\": \"create_search_pipeline\",\n"
+            + "                \"dest\": \"create_neural_search_index\"\n"
+            // + " }\n,"
+            // + " {\n"
+            // + " \"source\": \"create_neural_search_index\",\n"
+            // + " \"dest\": \"fetch_model\"\n"
+            + "            }\n"
+            + "        ]\n"
+            + "    }\n"
+            + "}";
```
owaiskazi19, line 70: We can have a dummy json file to read from there.

owaiskazi19 (commented):
First pass review

src/main/java/org/opensearch/flowframework/template/ProcessNode.java
dbwiddis: All the println statements will be changed to logging eventually, or just removed. They (and the Demo classes) are there temporarily until the sequence is properly implemented/tested in a test case.  Haven't gotten there yet, this is still in draft.

src/main/java/org/opensearch/flowframework/template/ProcessNode.java
dbwiddis: Execute is fine, but see #29 for some requests.

src/main/java/org/opensearch/flowframework/template/ProcessNode.java
```diff
@@ -0,0 +1,184 @@
...
+    }
+
+    /**
+     * Execute this node in the sequence. Initializes the node's {@link CompletableFuture} and completes it when the process completes.
+     *
+     * @return this node's future. This is returned immediately, while process execution continues asynchronously.
+     */
+    public CompletableFuture<WorkflowData> execute() {
+        this.future = new CompletableFuture<>();
+        CompletableFuture.runAsync(() -> {
```
kotwanikunal, line 125: This is where the thread is leaked. 
>Returns a new CompletableFuture that is asynchronously completed by a task running in the ForkJoinPool.commonPool() after it runs the given action.

I would suggest executing it bounded using a defined executor. You can use - 
```
runAsync(Runnable runnable, Executor executor)
```
The executor can be either passed in or defined as a class variable.

The tests can then do something similar to https://github.com/opensearch-project/OpenSearch/blob/d34b35272fa19bc53cb0f6843a92d2d1ee4e3b47/server/src/test/java/org/opensearch/common/blobstore/stream/read/listener/ReadContextListenerTests.java#L50-L58

dbwiddis: The ProcessNode class currently doesn't take a thread pool as an argument on instantiation.   Set the thread leak scope to NONE as a temporary fix.  Created issue https://github.com/opensearch-project/opensearch-ai-flow-framework/issues/42 to track permanent fix which will allow implementation of this approach for tests and properly using OpenSearch thread pool in production.

dbwiddis:
> Merging #26 (136ceaa) into main (0f0b65d) will increase coverage by 74.26%.

I hereby claim the record for most code coverage improvement in this project ever!

src/main/java/demo/CreateIndexWorkflowStep.java
```diff
@@ -0,0 +1,81 @@
...
+    public CompletableFuture<WorkflowData> execute(List<WorkflowData> data) {
+        CompletableFuture<WorkflowData> future = new CompletableFuture<>();
+        CompletableFuture.runAsync(() -> {
+            String inputIndex = null;
+            boolean first = true;
+            for (WorkflowData wfData : data) {
+                logger.debug(
+                    "{} sent params: {}, content: {}",
+                    first ? "Initialization" : "Previous step",
+                    wfData.getParams(),
```
joshpalis, line 47: Are these parameters/content based on the REST API that each step is related to or are these the parameters/content of the fractal execute API which takes in the use case template and provisions the infrastructure?

src/main/java/demo/CreateIndexWorkflowStep.java
```diff
@@ -0,0 +1,81 @@
...
+                    wfData.getContent()
+                );
+                if (first) {
+                    Map<String, String> params = data.get(0).getParams();
```
joshpalis, line 51: Is it correct to assume that the first item in this `WorkflowData` list contains the user provided data from the use case template?

src/main/java/demo/CreateIndexWorkflowStep.java
```diff
@@ -0,0 +1,81 @@
...
+     * Instantiate this class.
+     */
+    public CreateIndexWorkflowStep() {
+        this.name = "CREATE_INDEX";
+    }
+
+    @Override
+    public CompletableFuture<WorkflowData> execute(List<WorkflowData> data) {
+        CompletableFuture<WorkflowData> future = new CompletableFuture<>();
+        CompletableFuture.runAsync(() -> {
```
joshpalis, line 40: Haven't seen `CompletableFuture.runAsync()` before. Just wanted to understand the rationale for this step and if its necessary to handle each `WorkflowStep` this way

joshpalis (commented):
Initial pass

src/main/java/org/opensearch/flowframework/template/ProcessNode.java
```diff
@@ -0,0 +1,184 @@
...
+            List<CompletableFuture<WorkflowData>> predFutures = predecessors.stream().map(p -> p.getFuture()).collect(Collectors.toList());
+            if (!predecessors.isEmpty()) {
+                CompletableFuture<Void> waitForPredecessors = CompletableFuture.allOf(predFutures.toArray(new CompletableFuture<?>[0]));
```
owaiskazi19, line 128: Why we do need a new future here? We can just add a timeout for `predFutures` to get populated.

src/main/java/org/opensearch/flowframework/template/ProcessNode.java
```diff
@@ -0,0 +1,184 @@
...
+            try {
+                stepFuture.join();
+                future.complete(stepFuture.get());
+                logger.debug("<<< Completed {}", this.id);
+            } catch (InterruptedException | ExecutionException e) {
+                handleException(e);
+            }
+        });
+        return this.future;
+    }
```
owaiskazi19, line 158: Can we pull this method from this POJO class to a new class as this method will have more additions later like validations?

src/main/java/org/opensearch/flowframework/template/TemplateParser.java
```diff
@@ -0,0 +1,164 @@
...
+
+/**
+ * Utility class for parsing templates.
+ */
+public class TemplateParser {
+
+    private static final Logger logger = LogManager.getLogger(TemplateParser.class);
+
+    // Field names in the JSON. Package private for tests.
+    static final String WORKFLOW = "sequence";
```
owaiskazi19, line 40: here and for the below
```suggestion
    private static final String WORKFLOW = "sequence";
```

src/main/java/org/opensearch/flowframework/template/TemplateParser.java
```diff
@@ -0,0 +1,164 @@
...
+    private TemplateParser() {}
+
+    /**
+     * Parse a JSON representation of nodes and edges into a topologically sorted list of process nodes.
+     * @param json A string containing a JSON representation of nodes and edges
+     * @param workflowSteps A map linking JSON node names to Java objects implementing {@link WorkflowStep}
+     * @return A list of Process Nodes sorted topologically.  All predecessors of any node will occur prior to it in the list.
+     */
+    public static List<ProcessNode> parseJsonGraphToSequence(String json, Map<String, WorkflowStep> workflowSteps) {
+        Gson gson = new Gson();
```
owaiskazi19, line 59: We can `ObjectMapper` from OpenSearch here instead of adding a new dependency for parsing json.

src/main/java/org/opensearch/flowframework/template/TemplateParser.java
```diff
@@ -0,0 +1,164 @@
...
+        List<ProcessNode> nodes = new ArrayList<>();
+        List<ProcessSequenceEdge> edges = new ArrayList<>();
+
+        for (JsonElement nodeJson : graph.getAsJsonArray(NODES)) {
+            JsonObject nodeObject = nodeJson.getAsJsonObject();
+            String nodeId = nodeObject.get(NODE_ID).getAsString();
+            // The below steps will be replaced by a generator class that instantiates a WorkflowStep
+            // based on user_input data from the template.
+            WorkflowStep workflowStep = workflowSteps.get(nodeId);
+            // temporary demo POC of getting from a request to input data
```
owaiskazi19, line 73: Can we have a TODO here or issue if present? Anything works.

src/main/java/org/opensearch/flowframework/template/TemplateParser.java
```diff
@@ -0,0 +1,164 @@
...
+                    @Override
+                    public Map<String, Object> getContent() {
+                        // See CreateIndexRequest ParseFields for source of content keys needed
+                        return Map.of("mappings", request.mappings(), "settings", request.settings(), "aliases", request.aliases());
+                    }
+
+                    @Override
+                    public Map<String, String> getParams() {
+                        // See RestCreateIndexAction for source of param keys needed
+                        return Map.of("index", request.index());
```
owaiskazi19, line 89: Why are we having `index` in `getParams()` as it will be also provided in content?

src/main/java/org/opensearch/flowframework/template/TemplateParser.java
```diff
@@ -0,0 +1,164 @@
...
+        // See https://en.wikipedia.org/wiki/Topological_sorting#Kahn's_algorithm
+        // L <- Empty list that will contain the sorted elements
+        List<ProcessNode> sortedNodes = new ArrayList<>();
+        // S <- Set of all nodes with no incoming edge
+        Queue<ProcessNode> sourceNodes = new ArrayDeque<>();
+        nodes.stream().filter(n -> !predecessorEdges.containsKey(n)).forEach(n -> sourceNodes.add(n));
+        if (sourceNodes.isEmpty()) {
+            throw new IllegalArgumentException("No start node detected: all nodes have a predecessor.");
+        }
+        logger.debug("Start node(s): {}", sourceNodes);
```
owaiskazi19, line 138: Don't we want to log just the first node here and not an entire queue?

src/main/java/org/opensearch/flowframework/template/TemplateParser.java
```diff
@@ -0,0 +1,164 @@
...
+                // remove edge e from the graph
+                graph.remove(e);
+                // if m has no other incoming edges then
+                if (!predecessorEdges.get(m).stream().anyMatch(i -> graph.contains(i))) {
+                    // insert m into S
+                    sourceNodes.add(m);
+                }
+            }
+        }
+        if (!graph.isEmpty()) {
```
owaiskazi19, line 158:
```suggestion
        if (!graph.isEmpty() || graph.size() != sortedNodes.size()) {
```

owaiskazi19 (commented):
Thanks for putting this together. Did a thorough review this time :)

src/main/java/demo/CreateIndexWorkflowStep.java
dbwiddis: In the future I'll be adding an executor argument to this to define which thread pool it will run on (see #42). It's much the same as ExecutorService.submit().

The rationale here is that we can launch this separate runnable and it will run in a different thread, and eventually complete `future` when it does its work.  In the meantime the `execute()` method returns the (uncompleted) future so the next step can wait for all the predecessors to complete with an allOf().

src/main/java/demo/CreateIndexWorkflowStep.java
dbwiddis: Yes and yes. :)

First note this is just an example class, @owaiskazi19 is working on actually implementing this in PR #38 

The `WorkflowData` class is a container for params (string to string map) and content (string to object map) which are intended to represent the two portions of a Rest Request you can communicate information (params on the URI, content in the content/data field).

The list that each step will get will have 1 or more of these instances:
 - index 0 comes from initialization, so may come from the template (user data) or the initial API call from a user
 - the remaining elements of the list come from the previous workflow step, where params will probably be empty (but isn't required to be) and the content will be the equivalent of the API response JSON.

src/main/java/org/opensearch/flowframework/template/ProcessNode.java
```diff
@@ -0,0 +1,184 @@
...
+        this.predecessors = Set.copyOf(predecessors);
+    }
+
+    /**
+     * Execute this node in the sequence. Initializes the node's {@link CompletableFuture} and completes it when the process completes.
+     *
+     * @return this node's future. This is returned immediately, while process execution continues asynchronously.
+     */
+    public CompletableFuture<WorkflowData> execute() {
+        this.future = new CompletableFuture<>();
```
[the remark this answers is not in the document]
dbwiddis: Yes, see #42

src/main/java/org/opensearch/flowframework/template/ProcessNode.java
dbwiddis: I'm using this for `allOf()`.  

Another option is to stick them all in a list and call `join()` on them.

Not sure what other option you are proposing.

src/main/java/org/opensearch/flowframework/template/ProcessNode.java
```diff
@@ -0,0 +1,184 @@
...
+        CompletableFuture.runAsync(() -> {
+            List<CompletableFuture<WorkflowData>> predFutures = predecessors.stream().map(p -> p.getFuture()).collect(Collectors.toList());
+            if (!predecessors.isEmpty()) {
+                CompletableFuture<Void> waitForPredecessors = CompletableFuture.allOf(predFutures.toArray(new CompletableFuture<?>[0]));
+                try {
+                    waitForPredecessors.orTimeout(30, TimeUnit.SECONDS).get();
```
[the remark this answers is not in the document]
dbwiddis: This is just a placeholder for now, and it's for the entire workflow.  It may need to be longer.  It probably needs to be part of the template.

src/main/java/org/opensearch/flowframework/template/ProcessNode.java
```diff
@@ -0,0 +1,184 @@
...
+                } catch (InterruptedException | ExecutionException e) {
+                    handleException(e);
+                    return;
+                }
+            }
+            logger.debug(">>> Starting {}.", this.id);
```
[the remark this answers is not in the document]
dbwiddis: Maybe?  We also need to update status somewhere.   I have no strong feelings about the log level here.

src/main/java/org/opensearch/flowframework/template/ProcessNode.java
dbwiddis: Maybe?  Can you be more specific?  Is it a problem leaving it here for now and refactoring later?

src/main/java/org/opensearch/flowframework/template/TemplateParser.java
dbwiddis: See comment 2 lines above, I'm using these in the test classes.

src/main/java/org/opensearch/flowframework/template/TemplateParser.java
dbwiddis: `ObjectMapper` is `@opensearch.internal`.

Can we open a new issue to discuss this?  I don't want to delay this PR to have this discussion.

src/main/java/org/opensearch/flowframework/template/TemplateParser.java
dbwiddis: Already an issue #41

src/main/java/org/opensearch/flowframework/template/TemplateParser.java
dbwiddis: That's where it is in a CreateIndexRequest that this is a temporary placeholder for

src/main/java/demo/CreateIndexWorkflowStep.java
joshpalis: I see, thanks for the clarification. That PR #38 attempts to read the index name from the workflow data content rather than from the parameters. I've followed the same pattern (read everything from content) on my PR #44, but ill update this to read the pipelineId from parameters instead.

src/main/java/org/opensearch/flowframework/template/TemplateParser.java
dbwiddis: We do log the entire queue later.  Sample output in the PR:
```
17:22:34.597 INFO  o.o.f.demo.Demo - Parsing graph to sequence...
17:22:34.634 DEBUG o.o.f.template.TemplateParser - Start node(s): [fetch_model]
17:22:34.635 DEBUG o.o.f.template.TemplateParser - Execution sequence: [fetch_model, create_search_pipeline, create_ingest_pipeline, create_neural_search_index]
```

src/main/java/org/opensearch/flowframework/template/TemplateParser.java
dbwiddis: I don't think this change is needed.  The only time I'd get to the second condition is if `graph.isEmpty()`.  In that case `graph.size()` is 0, and since we exited the above while loop, `sourceNodes.isEmpty()` and its size is therefor also 0, so this check ends up always being `0 != 0` which will never be true.

src/main/java/org/opensearch/flowframework/template/ProcessNode.java
owaiskazi19: I missed `allOf()` due to scrolling. Looks good and let's keep `allOf()` only here as it allows us to continue our flow in a non-blocking way.

src/main/java/org/opensearch/flowframework/template/TemplateParser.java
owaiskazi19: We should be a little strict in introducing new dependencies in the plugin and fetch more from OpenSearch if already present. Yes, we should open an issue to discuss this.

src/main/java/org/opensearch/flowframework/template/TemplateParser.java
owaiskazi19: I meant for our execution API it will be passed in content and not params.

owaiskazi19:
I see this PR fixes #19 as well but I don't see any RestHandler for the execution API. Am I missing something?

src/main/java/org/opensearch/flowframework/template/TemplateParser.java
dbwiddis: These all actually really should be public as they will be part of the template definition.

src/main/java/org/opensearch/flowframework/template/ProcessNode.java
owaiskazi19: There's no problem in leaving anything here right now for the entire PR and refactoring later :)
Can we have an issue to address this later?

src/main/java/org/opensearch/flowframework/template/ProcessNode.java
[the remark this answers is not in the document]
owaiskazi19: Instead of `debug`, I'm more inclined in keeping the level as `info` everywhere.

src/main/java/org/opensearch/flowframework/template/TemplateParser.java
dbwiddis: https://github.com/opensearch-project/opensearch-ai-flow-framework/issues/46

src/main/java/org/opensearch/flowframework/template/TemplateParser.java
dbwiddis: I leave it to you to handle this in #38 however you want to.  The above is placeholder code that will be removed

dbwiddis:
> I see this PR fixes #19 as well but I don't see any RestHandler for the execution API. Am I missing something?

It completes all the items listed in #19's description.

API in this context is not referring to a REST API

src/main/java/org/opensearch/flowframework/template/ProcessNode.java
dbwiddis: I'll handle it as part of #42

owaiskazi19:
> API in this context is not referring to a REST API

We would need a REST API though for frontend to trigger the process sequencing which is already handled in this PR.

src/main/java/org/opensearch/flowframework/template/ProcessNode.java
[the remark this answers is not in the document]
dbwiddis: I opened https://github.com/opensearch-project/opensearch-ai-flow-framework/issues/45 for this

dbwiddis:
> Won't we need a REST API though for frontend to trigger the process sequencing which is already handled in this PR?

Yes, we'll need a REST call to send the template in that we can't do until we have the template design finalized.  I see a few scattered REST API issues, is there a META with all of the required ones listed somewhere?

owaiskazi19:
> Yes, we'll need a REST call to send the template in that we can't do until we have the template design finalized. I see a few scattered REST API issues, is there a META with all of the required ones listed somewhere?

I was under the impression #19 is for the REST execution API. Based on our decided design [here](https://github.com/opensearch-project/OpenSearch/issues/9213#issuecomment-1724640922) we will need 2 APIs:
1. Execute API for frontend/users to trigger the worklow and pass use case template in the payload.
2. Orchestrate API to handle the operations to be done by the workflow.

owaiskazi19 (approved):
Thanks for addressing the feedback patiently. LGTM!

dbwiddis:
> I was under the impression #19 is for the REST execution API. Based on our decided design [here](https://github.com/opensearch-project/OpenSearch/issues/9213#issuecomment-1724640922) we will need 2 APIs:
> 
> 1. Execute API for frontend/users to trigger the worklow and pass use case template in the payload.
> 2. Orchestrate API to handle the operations to be done by the workflow.

So exection is by orchestrate?  That's what triggers the workflow using internal client stuff.

The template APIs would be CRUD.

We should definitely clarify all this in a new meta issue summarizing the various RFC comments.

owaiskazi19:
> So exection is by orchestrate? That's what triggers the workflow using internal client stuff.

Execution is by execution API. The operations to run on the workflow like Search/ingest would be done using orchestrate API.

> The template APIs would be CRUD.

The execution API should be able to pass the created use case template from frontend to backend as payload.

src/main/java/org/opensearch/flowframework/template/ProcessNode.java
```diff
@@ -0,0 +1,189 @@
...
+     * @param input Input required by the node
+     */
+    public ProcessNode(String id, WorkflowStep workflowStep, WorkflowData input) {
+        this.id = id;
+        this.workflowStep = workflowStep;
+        this.input = input;
+    }
+
+    /**
+     * Returns the node's id.
```
jackiehanyang, line 65: is this node id information coming from use-case template or it will be assigned at backend? I don't remember seeing this node id on the use-case template, you might want to confirm with front end people if you need this information from use-case template

dbwiddis: There is currently an `id` field with a somewhat descriptive name in the proposed use case template, which is what I based the JSON for this on. 

It needs to be unique, which we haven't really explicitly stated, but it's the same string that is used to define the source and destination for the edges, which somewhat forces that requirement.

joshpalis (approved):

PR #38, opened 2023-09-13
Adds Create Index Building block 

owaiskazi19:
### Description
Adds Create Index Building block for the workflow setup and logging functionality in the project.

### Issues Resolved
Part of https://github.com/opensearch-project/opensearch-ai-flow-framework/issues/21


build.gradle
```diff
@@ -105,6 +105,7 @@ repositories {
 dependencies {
     implementation "org.opensearch:opensearch:${opensearch_version}"
     implementation 'org.junit.jupiter:junit-jupiter:5.10.0'
+    implementation "org.apache.logging.log4j:log4j-slf4j-impl:2.19.0"
```
dbwiddis, line 108: This does nothing without slf4j-api.  We already inherit log4j2 from OpenSearch so we just need to configure the logger(s).  We should use the OpenSearch config like other plugins.

src/main/java/org/opensearch/flowframework/FlowFrameworkPlugin.java
```diff
@@ -8,11 +8,47 @@
...
+        ScriptService scriptService,
+        NamedXContentRegistry xContentRegistry,
+        Environment environment,
+        NodeEnvironment nodeEnvironment,
+        NamedWriteableRegistry namedWriteableRegistry,
+        IndexNameExpressionResolver indexNameExpressionResolver,
+        Supplier<RepositoriesService> repositoriesServiceSupplier
+    ) {
+        this.client = client;
+        CreateIndexStep createIndexStep = new CreateIndexStep(client);
```
dbwiddis, line 51: Not sure why this needs to be instantiated in advance, particularly as a singleton.  As I'm iterating on draft PR #26 I think we need a generator class for these steps.

src/main/java/org/opensearch/flowframework/workflow/CreateIndexStep.java
```diff
@@ -0,0 +1,73 @@
...
+import org.opensearch.action.admin.indices.create.CreateIndexRequest;
+import org.opensearch.action.admin.indices.create.CreateIndexResponse;
+import org.opensearch.client.Client;
+import org.opensearch.common.xcontent.XContentType;
+
+import java.io.IOException;
+import java.net.URL;
+import java.util.concurrent.CompletableFuture;
+
+public class CreateIndexStep implements WorkflowStep {
```
dbwiddis, line 25: See https://github.com/opensearch-project/opensearch-ai-flow-framework/pull/26/files#diff-8862e1a64f0bf9a0c3816e3b988f2faa628d1dae4e5ece3faab12a28025734b6 for ideas.

src/main/java/org/opensearch/flowframework/workflow/CreateIndex/CreateIndexStep.java
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    @Override
+    public CompletableFuture<WorkflowData> execute(List<WorkflowData> data) {
+        CompletableFuture<WorkflowData> future = new CompletableFuture<>();
+        ActionListener<CreateIndexResponse> actionListener = new ActionListener<>() {
+
+            @Override
+            public void onResponse(CreateIndexResponse createIndexResponse) {
+                logger.info("created index:{}");
```
dbwiddis, line 46:
```suggestion
                logger.info("created index: {}", createIndexResponse.index());
```
The `{}` in a log statement corresponds to an object param after it, that it evaluates the toString of.

src/main/java/org/opensearch/flowframework/workflow/CreateIndex/CreateIndexStep.java
```diff
@@ -0,0 +1,94 @@
...
+                future.completeExceptionally(e);
+            }
+        };
+
+        String index = null;
+
+        for (WorkflowData workflowData : data) {
+            // Fetch index from content i.e. request body of execute API
+            Map<String, Object> content = workflowData.getContent();
+            index = (String) content.get("index");
```
dbwiddis, line 67: As written, this will iterate all of the previous steps' data, with later ones overwriting earlier ones.  This may not be what you want.

The first element in `data` will be the initialization from the template parsing, most likely to contain any user-provided info.  The later elements are any immediate predecessor steps but ordering is not guaranteed.

TLDR: you might want a 'break;` here if index is not `null`.

src/main/java/org/opensearch/flowframework/workflow/CreateIndex/CreateIndexStep.java
owaiskazi19: I missed that adding `index`:(

src/main/java/org/opensearch/flowframework/FlowFrameworkPlugin.java
owaiskazi19: Will update this once https://github.com/opensearch-project/opensearch-ai-flow-framework/pull/47 is in if that works?

src/main/java/org/opensearch/flowframework/workflow/CreateIndex/CreateIndexStep.java
```diff
@@ -0,0 +1,115 @@
...
+                break;
+            }
+        }
+
+        // TODO:
+        // 1. Create settings based on the index settings received from content
+
+        try {
+            CreateIndexRequest request = new CreateIndexRequest(index).mapping(
+                getIndexMappings("mappings/" + type + ".json"),
```
joshpalis, line 88: Not blocking for this PR, but in my opinion index mappings should be provided by the user, since the user may want control over what fields they need. Having a static mapping is fine for now as we work towards building support for default use cases, but we'll definitely need some common mechanism to create these request bodies given the provided workflow data. This would apply to pipeline configurations and index settings as well, pretty much for any content that requires us to map fields with the user inputs.

src/main/java/org/opensearch/flowframework/workflow/CreateIndex/CreateIndexStep.java
```diff
@@ -0,0 +1,115 @@
...
+            Map<String, Object> content = workflowData.getContent();
+            index = (String) content.get("index-name");
+            type = (String) content.get("type");
+            if (index != null && type != null && settings != null) {
+                break;
+            }
+        }
+
+        // TODO:
+        // 1. Create settings based on the index settings received from content
```
joshpalis, line 84: Since this initial implementation is specifically for a KNN index, we'll need at least a `knn=true` within the index settings.

src/test/resources/template/datademo.json
```diff
@@ -3,11 +3,13 @@
         "nodes": [
             {
                 "id": "create_index",
-                "index_name": "demo"
+                "index_name": "demo",
+                "type": "knn"
```
dbwiddis, line 7: I'm actually deleting my "datademo" json file in #47 and the "demo" JSON is in full template style.

Please write your own JSON test class if you need one, but I don't see where you use this one, so best to just undo this file change.

dbwiddis (commented):
Initial review.  A few substantive and a few non-substantive requests.

src/main/java/org/opensearch/flowframework/workflow/CreateIndex/CreateIndexStep.java
owaiskazi19: I wouldn't say this is just for k-nn index because the mapping file is for `knn` in this PR. Based on your comment above if the mapping is provided by the user, then we do need to get the settings from the user as well or create dynamic settings based on the mappings provided for any type of index.

src/main/java/org/opensearch/flowframework/workflow/CreateIndex/CreateIndexStep.java
```diff
@@ -0,0 +1,115 @@
...
+
+        try {
+            CreateIndexRequest request = new CreateIndexRequest(index).mapping(
+                getIndexMappings("mappings/" + type + ".json"),
+                XContentType.JSON
+            );
+            client.admin().indices().create(request, actionListener);
+        } catch (Exception e) {
```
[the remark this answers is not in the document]
owaiskazi19: Agreed, we should stop the workflow and call our exceptions here. Will iterate on this after #43 is merged.

src/main/java/org/opensearch/flowframework/workflow/CreateIndex/CreateIndexStep.java
dbwiddis: I think we can have a mix here... static mappings for well-known things, user-defined mappings otherwise.  We'd just need to have some sort of syntax for the template, e.g., if user puts `{{knn}}` it will use that static mapping, or they can provide their own.

src/test/java/org/opensearch/flowframework/workflow/CreateIndex/CreateIndexStepTests.java
```diff
@@ -0,0 +1,105 @@
...
+
+    @Override
+    public void setUp() throws Exception {
+        super.setUp();
+
+        inputData = new WorkflowData() {
+
+            @Override
+            public Map<String, Object> getContent() {
+                // See CreateIndexRequest ParseFields for source of content keys needed
```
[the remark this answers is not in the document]
owaiskazi19: From the execute API build in #49 , all the fields will be part of the content. This comment was a leftover. Removed it.

src/test/java/org/opensearch/flowframework/workflow/CreateIndex/CreateIndexStepTests.java
```diff
@@ -0,0 +1,105 @@
...
+
+        ArgumentCaptor<ActionListener> actionListenerCaptor = ArgumentCaptor.forClass(ActionListener.class);
+        CompletableFuture<WorkflowData> future = createIndexStep.execute(List.of(inputData));
+        assertFalse(future.isDone());
+        verify(indicesAdminClient, times(1)).create(any(CreateIndexRequest.class), actionListenerCaptor.capture());
+
+        actionListenerCaptor.getValue().onFailure(new Exception());
+
+        assertTrue(future.isDone());
+        assertThrows(Exception.class, () -> future.get().getContent());
```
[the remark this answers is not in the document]
owaiskazi19: That's a good point.

src/test/resources/template/datademo.json
owaiskazi19: I made this change to test locally but then I populated workflowData in test, so yes can remove these changes.

dbwiddis (approved):

joshpalis (approved):

src/main/java/org/opensearch/flowframework/workflow/CreateIndex/CreateIndexStep.java
```diff
@@ -0,0 +1,115 @@
...
+        String index = null;
+        String type = null;
+        Settings settings = null;
+
+        for (WorkflowData workflowData : data) {
+            Map<String, Object> content = workflowData.getContent();
+            index = (String) content.get("index-name");
+            type = (String) content.get("type");
+            if (index != null && type != null && settings != null) {
+                break;
```
jackiehanyang, line 79: what's the point of having this break check here? We don't assign value to the `setting` variable within the loop, so it remains `null`. Isn't this means the loop will always process all items, even if it's unnecessary?

src/main/java/org/opensearch/flowframework/workflow/CreateIndex/CreateIndexStep.java
```diff
@@ -0,0 +1,115 @@
...
+        };
+
+        String index = null;
+        String type = null;
+        Settings settings = null;
+
+        for (WorkflowData workflowData : data) {
+            Map<String, Object> content = workflowData.getContent();
+            index = (String) content.get("index-name");
+            type = (String) content.get("type");
```
jackiehanyang, line 77: nit: consider catching NPE when using map.get().

src/main/java/org/opensearch/flowframework/workflow/CreateIndex/CreateIndexStep.java
dbwiddis: I interpreted this as a placeholder for future functionality, see the TODO below.

jackiehanyang (approved):

src/main/java/org/opensearch/flowframework/workflow/CreateIndex/CreateIndexStep.java
owaiskazi19: Let me take this in the followup PR.

PR #43, opened 2023-09-18
Add global context index and indices handler

jackiehanyang:
### Description
- Creating an indices handler to create index while if it's absent, or upgrade the mapping if it's needed. 
- Adding use-case template index mapping
- Adding a class to throw our own custom exceptions. 

Integration tests will be added later when the setup for integration test CI is ready. 

### Issues Resolved
https://github.com/opensearch-project/opensearch-ai-flow-framework/issues/51


src/main/java/org/opensearch/flowframework/constant/CommonValue.java
```diff
@@ -0,0 +1,87 @@
...
+ */
+package org.opensearch.flowframework.constant;
+
+public class CommonValue {
+
+    public static Integer NO_SCHEMA_VERSION = 0;
+    public static final String META = "_meta";
+    public static final String SCHEMA_VERSION_FIELD = "schema_version";
+    public static final Integer GLOBAL_CONTEXT_INDEX_SCHEMA_VERSION = 1;
+    public static final String GLOBAL_CONTEXT_INDEX_MAPPING =
```
dbwiddis, line 17: Can we put this in `src/main/resources` and read it into this constant off the class path?  See [here](https://github.com/opensearch-project/anomaly-detection/blob/713d556d980ccb144eb6ae952cebc7b3963ed28c/src/main/java/org/opensearch/ad/settings/AnomalyDetectorSettings.java#L192-L194).

build.gradle
```diff
@@ -105,6 +105,7 @@ repositories {
 dependencies {
     implementation "org.opensearch:opensearch:${opensearch_version}"
     implementation 'org.junit.jupiter:junit-jupiter:5.10.0'
+    implementation 'org.projectlombok:lombok:1.18.22'
```
dbwiddis, line 108: We should have a separate discussion/issue/RFC on whether we want to use a code generation tool like Lombok, and figure out what modifications we need to make to our CI to support it.

src/main/java/org/opensearch/flowframework/exception/FlowFrameworkException.java
```diff
@@ -0,0 +1,20 @@
...
+ */
+package org.opensearch.flowframework.exception;
+
+public class FlowFrameworkException extends RuntimeException {
+    /**
+     * Constructor with error message.
+     *
+     * @param message message of the exception
+     */
+    public FlowFrameworkException(String message) {
```
dbwiddis, line 17: I like the idea of having our own custom exceptions.  Given that these will eventually need to be processed in a RestResponse, can we consider including the appropriate return code as part of constructing this exception?  For example, "index not found" should return `RestStatus.NOT_FOUND` etc.

dbwiddis (commented):
Initial pass

src/main/java/org/opensearch/flowframework/constant/CommonValue.java
jackiehanyang: I'm fine with putting this index mapping in either path, I think it's just different style of code. But I do prefer to put it in under `src/main/java` cause it can save us from setting up methods to read from `src/main/resources`, something like `getMapping()` ML-Commons is using this approach as well, see [here](https://github.com/opensearch-project/ml-commons/blob/main/common/src/main/java/org/opensearch/ml/common/CommonValue.java#L45)

build.gradle
jackiehanyang: removed lombok dependency for now. It's not needed in this pr. I will open a discuss and figure out the XContent question before adding this dependency

build.gradle
```diff
@@ -117,8 +117,12 @@ dependencies {
 
 test {
     include '**/*Tests.class'
+    systemProperty 'tests.security.manager', 'false'
```
joshpalis, line 120: Trying to understand these build.gradle settings better, does this disable security for tests?

src/main/java/org/opensearch/flowframework/constant/CommonValue.java
```diff
@@ -0,0 +1,91 @@
...
+    public static final Integer GLOBAL_CONTEXT_INDEX_SCHEMA_VERSION = 1;
+    public static final String GLOBAL_CONTEXT_INDEX_MAPPING = "{\n  "
+        + "   \"dynamic\": false,\n"
+        + "   \"_meta\": {\n"
+        + "       \"schema_version\": "
+        + GLOBAL_CONTEXT_INDEX_SCHEMA_VERSION
+        + "\n"
+        + "   },\n"
+        + "   \"properties\": {\n"
+        + "       \"pipeline_id\": {\n"
```
joshpalis, line 28: let's change this to `workflow_id` to avoid confusion with ingest/search pipelines

src/main/java/org/opensearch/flowframework/indices/FlowFrameworkIndicesHandler.java
```diff
@@ -0,0 +1,176 @@
...
+import static org.opensearch.flowframework.constant.CommonValue.NO_SCHEMA_VERSION;
+import static org.opensearch.flowframework.constant.CommonValue.SCHEMA_VERSION_FIELD;
+
+/**
+ * Flow Framework Indices Handler
+ */
+public class FlowFrameworkIndicesHandler {
+
+    private static final Logger logger = LogManager.getLogger(FlowFrameworkIndicesHandler.class);
+    private static final Map<String, Object> indexSettings = Map.of("index.auto_expand_replicas", "0-1");
```
joshpalis, line 39: What's the justification for this `auto_expand_replicas` setting, just trying to understand why this is necessary

src/main/java/org/opensearch/flowframework/indices/FlowFrameworkIndicesHandler.java
```diff
@@ -0,0 +1,176 @@
...
+    /**
+     * General method for initiate flow framework indices or update index mapping if it's needed
+     * @param index flow framework index
+     * @param listener action listener
+     */
+    public void initFlowFrameworkIndexIfAbsent(FlowFrameworkIndex index, ActionListener<Boolean> listener) {
+        String indexName = index.getIndexName();
+        String mapping = index.getMapping();
+
+        try (ThreadContext.StoredContext threadContext = client.threadPool().getThreadContext().stashContext()) {
```
joshpalis, line 78: From what I understand about security, we only need to stash the thread context if this particular index is managed by the security plugin's list of protected indices. This allows the plugin to do CRUD operations on this protected index. Is the plan to add all flow framework indices to the security plugin's list of protected indices, and if not, why stash the thread context?

src/main/java/org/opensearch/flowframework/indices/FlowFrameworkIndicesHandler.java
```diff
@@ -0,0 +1,176 @@
...
+                        internalListener.onResponse(false);
+                    }
+                }, e -> {
+                    logger.error("Failed to create index " + indexName, e);
+                    internalListener.onFailure(e);
+                });
+                CreateIndexRequest request = new CreateIndexRequest(indexName).mapping(mapping).settings(indexSettings);
+                client.admin().indices().create(request, actionListener);
+            } else {
+                logger.debug("index:{} is already created", indexName);
```
joshpalis, line 95: I'd prefer that we keep logging at the `info` or `error` level for success and failure cases respectively

joshpalis (commented):
Initial pass

src/main/java/org/opensearch/flowframework/indices/FlowFrameworkIndicesHandler.java
```diff
@@ -0,0 +1,176 @@
...
+        IndexMetadata indexMetaData = clusterService.state().getMetadata().indices().get(indexName);
+        if (indexMetaData == null) {
+            listener.onResponse(Boolean.FALSE);
+            return;
+        }
+        Integer oldVersion = NO_SCHEMA_VERSION;
+        Map<String, Object> indexMapping = indexMetaData.mapping().getSourceAsMap();
+        Object meta = indexMapping.get(META);
+        if (meta != null && meta instanceof Map) {
+            @SuppressWarnings("unchecked")
```
owaiskazi19, line 167: Why do we have suppress warnings here?

owaiskazi19 (changes requested):
Need tests! Since we both kind of worked on creating the index. I will create a new PR to have the generic way of creating the index and later you can pass the Global Context mapping file to it to create GC index.

src/main/java/org/opensearch/flowframework/indices/FlowFrameworkIndicesHandler.java
dbwiddis: I suspect because all we know is that `meta` is an `Object` that an instance of `Map` but we don't know the types of its key and value when they are cast.

src/main/java/org/opensearch/flowframework/constant/CommonName.java
```diff
@@ -0,0 +1,17 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+package org.opensearch.flowframework.constant;
```
dbwiddis, line 9: Curious about this choice of package name. Given the class names, seems `common` might be better?

src/main/java/org/opensearch/flowframework/indices/FlowFrameworkIndicesHandler.java
```diff
@@ -0,0 +1,176 @@
...
+                        logger.info("create index:{}", indexName);
+                        internalListener.onResponse(true);
+                    } else {
+                        internalListener.onResponse(false);
+                    }
+                }, e -> {
+                    logger.error("Failed to create index " + indexName, e);
+                    internalListener.onFailure(e);
+                });
+                CreateIndexRequest request = new CreateIndexRequest(indexName).mapping(mapping).settings(indexSettings);
```
dbwiddis, line 92: does the mapping here require a mediatype?  Looking at the javadoc it seems it expects JSON but needs a special format with a `_doc` key.   We shold be consistent between our various implementations so all our mapping files look consistent.

src/main/java/org/opensearch/flowframework/indices/FlowFrameworkIndicesHandler.java
dbwiddis: I personally don't mind extra debug or even trace logging.

src/main/java/org/opensearch/flowframework/indices/FlowFrameworkIndicesHandler.java
```diff
@@ -0,0 +1,176 @@
...
+            } else {
+                logger.debug("index:{} is already created", indexName);
+                if (indexMappingUpdated.containsKey(indexName) && !indexMappingUpdated.get(indexName).get()) {
+                    shouldUpdateIndex(indexName, index.getVersion(), ActionListener.wrap(r -> {
+                        if (r) {
+                            // return true if update index is needed
+                            client.admin()
+                                .indices()
+                                .putMapping(
+                                    new PutMappingRequest().indices(indexName).source(mapping, XContentType.JSON),
```
dbwiddis, line 103: I think this requires a `MediaType` argument instead of `XContentType`.

PR #44, opened 2023-09-18
Adds Create Ingest Pipeline Step

joshpalis:
### Description
Adds CreateIngestPipelineStep for workflow set up. In this current state, this step is hardcoded to parse the necessary information from the `WorkflowData` related to the `text_embedding` processor. 

### Issues Resolved
https://github.com/opensearch-project/opensearch-ai-flow-framework/issues/24


src/main/java/org/opensearch/flowframework/workflow/CreateIngestPipelineStep.java
```diff
@@ -0,0 +1,178 @@
...
+        BytesReference configuration = null;
+
+        // Extract required content from workflow data and generate the ingest pipeline configuration
+        for (WorkflowData workflowData : data) {
+
+            Map<String, String> parameters = workflowData.getParams();
+            Map<String, Object> content = workflowData.getContent();
+            logger.debug("Previous step sent params: {}, content: {}", parameters, content);
+
+            if (content.containsKey(PIPELINE_ID_FIELD)) {
```
dbwiddis, line 76: This style works, but six if statements in a row and multiple repeated all-caps constants start to strain readability.  Consider iterating over the entrySet and making assignments based on the keys you found, something like:
```java
for (Entry<String, Object> e : content.entrySet()) {
    switch (e.getKey()) {
        case PIPELINE_ID_FIELD:
            pipelineId = (String) e.getValue();
            break;
        case ....
    }
}
```
You could also get fancy with lambdas and use `ArrayList.forEach()` to do the iteration around this, something like:
```java
data.forEach(map -> {
    map.entrySet().forEach(e -> {
        switch (e.getKey()) {
            case PIPELINE_ID_FIELD:
                pipelineId = (String) e.getValue();
                break;
            case ....
        }
    });
});
```

src/main/java/org/opensearch/flowframework/workflow/CreateIngestPipelineStep.java
```diff
@@ -0,0 +1,178 @@
...
+                modelId = (String) content.get(MODEL_ID_FIELD);
+            }
+            if (content.containsKey(INPUT_FIELD)) {
+                inputFieldName = (String) content.get(INPUT_FIELD);
+            }
+            if (content.containsKey(OUTPUT_FIELD)) {
+                outputFieldName = (String) content.get(OUTPUT_FIELD);
+            }
+
+            if (Stream.of(pipelineId, description, modelId, inputFieldName, outputFieldName).allMatch(x -> x != null)) {
```
dbwiddis, line 92: I've never seen this for null checking, I like it!  

However, I don't see an "else" clause here, should we throw an exception if we don't have all the required fields?

src/main/java/org/opensearch/flowframework/workflow/CreateIngestPipelineStep.java
```diff
@@ -0,0 +1,178 @@
...
+     */
+    private XContentBuilder buildIngestPipelineRequestContent(
+        String description,
+        String modelId,
+        String inputFieldName,
+        String outputFieldName
+    ) throws IOException {
+        return XContentFactory.jsonBuilder()
+            .startObject()
+            .field("description", description)
```
dbwiddis, line 167: Let's make all of these constants like `DESCRIPTION_FIELD` since you'll need a corresponding `parse()` method relying on them.  See my template parsing methods in #47 for examples.

dbwiddis (commented):
Initial review.  Looks good so far!

src/main/java/org/opensearch/flowframework/FlowFrameworkPlugin.java
```diff
@@ -8,11 +8,48 @@
...
+        ScriptService scriptService,
+        NamedXContentRegistry xContentRegistry,
+        Environment environment,
+        NodeEnvironment nodeEnvironment,
+        NamedWriteableRegistry namedWriteableRegistry,
+        IndexNameExpressionResolver indexNameExpressionResolver,
+        Supplier<RepositoriesService> repositoriesServiceSupplier
+    ) {
+        this.client = client;
+        CreateIngestPipelineStep createIngestPipelineStep = new CreateIngestPipelineStep(client);
```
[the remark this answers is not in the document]
joshpalis: Yes this is just temporary to gain access to the client. Ill move this once we have that factory class

src/main/java/org/opensearch/flowframework/workflow/CreateIngestPipelineStep.java
joshpalis: The exception is thrown below when we check if the configuration is null or not, but ill move this around so its more apparent

src/main/java/org/opensearch/flowframework/FlowFrameworkPlugin.java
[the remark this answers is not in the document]
dbwiddis: Same comment I made on #38 for the same reason... OK for now, if #47 gets merged first you can remove it or if this PR gets merged first I'll make it part of #47.

src/main/java/org/opensearch/flowframework/workflow/CreateIngestPipelineStep.java
```diff
@@ -0,0 +1,199 @@
...
+                        break;
+                    case MODEL_ID_FIELD:
+                        modelId = (String) content.get(MODEL_ID_FIELD);
+                        break;
+                    case INPUT_FIELD:
+                        inputFieldName = (String) content.get(INPUT_FIELD);
+                        break;
+                    case OUTPUT_FIELD:
+                        outputFieldName = (String) content.get(OUTPUT_FIELD);
+                        break;
```
dbwiddis, line 103: You need a `default` case, even if it's just "ignore extra keys".

src/main/java/org/opensearch/flowframework/workflow/CreateIngestPipelineStep.java
```diff
@@ -0,0 +1,199 @@
...
+     *  "description" : "<description>",
+     *  "processors" : [
+     *      {
+     *          "<type>" : {
+     *              "model_id" : "<model_id>",
+     *              "field_map" : {
+     *                  "<input_field_name>" : "<output_field_name>"
+     *          }
+     *      }
+     *  ]
```
dbwiddis, line 166: I only did one level of nested JSON objects in my template parsing in #47 for "generic" requests assuming the "processor" field is a string-to-string map.  I could modify it to read arbitrarily nested string-to-object maps, but I'm not sure that's a good idea. 

Is this processor format consistent enough that we can generalize this "temporary" parsing when reading in a template and always do it for a "processors" key?

src/test/java/org/opensearch/flowframework/workflow/CreateIngestPipeline/CreateIngestPipelineStepTests.java
```diff
@@ -0,0 +1,155 @@
...
+                    "some description",
+                    "type",
+                    "text_embedding",
+                    "model_id",
+                    "model_id",
+                    "input_field_name",
+                    "inputField",
+                    "output_field_name",
+                    "outputField"
+                );
```
dbwiddis, line 63: (Style) I made the same comment on #38 but this is even a more unreadable example.  Yes, `Map.of()` permits this for up to 10 pairs, but it's very confusing.

Can you change this to
```
return Map.ofEntries(
    Map.entry("id", "pipelineId"),
    Map.entry("description", "some description"),
    ...
)
```

src/test/java/org/opensearch/flowframework/workflow/CreateIngestPipeline/CreateIngestPipelineStepTests.java
```diff
@@ -0,0 +1,155 @@
...
+            }
+
+            @Override
+            public Map<String, String> getParams() {
+                return Map.of();
+            }
```
dbwiddis, line 69: I made the same comment on #38.  The interface already includes a default empty map so you don't need to override.

src/test/java/org/opensearch/flowframework/workflow/CreateIngestPipeline/CreateIngestPipelineStepTests.java
```diff
@@ -0,0 +1,155 @@
...
+        ArgumentCaptor<ActionListener> actionListenerCaptor = ArgumentCaptor.forClass(ActionListener.class);
+        CompletableFuture<WorkflowData> future = createIngestPipelineStep.execute(List.of(inputData));
+
+        assertFalse(future.isDone());
+
+        // Mock put pipeline request execution and return true
+        verify(clusterAdminClient, times(1)).putPipeline(any(PutPipelineRequest.class), actionListenerCaptor.capture());
+        actionListenerCaptor.getValue().onResponse(new AcknowledgedResponse(true));
+
+        assertTrue(future.isDone());
```
dbwiddis, line 107: Also test `!future.isCompletedExceptionally()`

src/test/java/org/opensearch/flowframework/workflow/CreateIngestPipeline/CreateIngestPipelineStepTests.java
```diff
@@ -0,0 +1,155 @@
...
+        ArgumentCaptor<ActionListener> actionListenerCaptor = ArgumentCaptor.forClass(ActionListener.class);
+        CompletableFuture<WorkflowData> future = createIngestPipelineStep.execute(List.of(inputData));
+
+        assertFalse(future.isDone());
+
+        // Mock put pipeline request execution and return false
+        verify(clusterAdminClient, times(1)).putPipeline(any(PutPipelineRequest.class), actionListenerCaptor.capture());
+        actionListenerCaptor.getValue().onFailure(new Exception());
+
+        assertTrue(future.isDone());
```
dbwiddis, line 124: Change this to test `future.isCompletedExceptionally()`

src/test/java/org/opensearch/flowframework/workflow/CreateIngestPipeline/CreateIngestPipelineStepTests.java
```diff
@@ -0,0 +1,155 @@
...
+        assertThrows(Exception.class, () -> future.get());
```
dbwiddis, line 125: This is not the same `Exception` you threw on line 122.  This is an `ExecutionException` with the line 122 exception as the cause.

Recommend testing `ExecutionException` here at a minimum.  Optionally add a message to the line 122 exception and test the getCause message matches.

src/test/java/org/opensearch/flowframework/workflow/CreateIngestPipeline/CreateIngestPipelineStepTests.java
```diff
@@ -0,0 +1,155 @@
...
+
+    public void testMissingData() {
+        CreateIngestPipelineStep createIngestPipelineStep = new CreateIngestPipelineStep(client);
+
+        // Data with missing input and output fields
+        WorkflowData incorrectData = new WorkflowData() {
+
+            @Override
+            public Map<String, Object> getContent() {
+                return Map.of("id", "pipelineId", "description", "some description", "type", "text_embedding", "model_id", "model_id");
```
dbwiddis, line 136: See above on readability and `Map.ofEntries()`

src/test/java/org/opensearch/flowframework/workflow/CreateIngestPipeline/CreateIngestPipelineStepTests.java
```diff
@@ -0,0 +1,155 @@
...
+            }
+
+            @Override
```
dbwiddis, line 139: see above on just deleting this

src/test/java/org/opensearch/flowframework/workflow/CreateIngestPipeline/CreateIngestPipelineStepTests.java
```diff
@@ -0,0 +1,155 @@
...
+                return Map.of();
+            }
+        };
+
+        CompletableFuture<WorkflowData> future = createIngestPipelineStep.execute(List.of(incorrectData));
+        assertTrue(future.isDone());
+
+        try {
+            future.get();
+        } catch (Exception e) {
```
dbwiddis, line 150: I really don't like broadly catching exceptions like this.  You know the exact types of exceptions to expect here.

dbwiddis (commented):
Some similar comments to #38

src/main/java/org/opensearch/flowframework/workflow/CreateIngestPipelineStep.java
joshpalis: All pipeline configuration formats follow a similar pattern :
```
{
    "description" : "",
    "processors": [
        {
             "processor1" : {
                 // required inputs
             }
         },
         {
             "processor2" : {
                 // required inputs
             }
         },
         ....
         ...
    ]
}
```

We can definitely generalize the creation of the configuration as long as the workflow data has a list of processors and their required inputs. The processors themselves vary in terms of format, so doing it this way would require some xcontent parser for each type of processor, since we'll need to write to specific fields.

src/test/java/org/opensearch/flowframework/workflow/CreateIngestPipeline/CreateIngestPipelineStepTests.java
joshpalis: sure ill go ahead and make this change

dbwiddis (approved):

src/test/java/org/opensearch/flowframework/workflow/CreateIngestPipeline/CreateIngestPipelineStepTests.java
dbwiddis: Alternately merge #54 and use the much simpler syntax. :)

src/main/java/org/opensearch/flowframework/workflow/CreateIngestPipelineStep.java
```diff
@@ -0,0 +1,201 @@
...
+    private final Client client;
+
+    /**
+     * Instantiates a new CreateIngestPipelineStep
+     *
+     * @param client The client to create a pipeline and store workflow data into the global context index
+     */
+    public CreateIngestPipelineStep(Client client) {
+        this.clusterAdminClient = client.admin().cluster();
+        this.client = client;
```
owaiskazi19, line 61: We can remove this from the constructor.

src/test/java/org/opensearch/flowframework/workflow/CreateIngestPipeline/CreateIngestPipelineStepTests.java
```diff
@@ -0,0 +1,143 @@
...
+        // Mock put pipeline request execution and return false
+        verify(clusterAdminClient, times(1)).putPipeline(any(PutPipelineRequest.class), actionListenerCaptor.capture());
+        actionListenerCaptor.getValue().onFailure(new Exception("Failed to create ingest pipeline"));
+
+        assertTrue(future.isDone() && future.isCompletedExceptionally());
+        try {
+            future.get();
+        } catch (ExecutionException e) {
+            assertTrue(e.getMessage().contains("Failed to create ingest pipeline"));
+        }
```
owaiskazi19, line 113: We should test the above like the below and remove the try/catch block in the test as we want to test the exception rather catching it.
```java
assertTrue(future.isCompletedExceptionally());
ExecutionException ex = assertThrows(ExecutionException.class, () -> future.get().getContent());
assertTrue(ex.getCause() instanceof Exception);
assertEquals("Failed to create an index", ex.getCause().getMessage());
```

src/test/java/org/opensearch/flowframework/workflow/CreateIngestPipeline/CreateIngestPipelineStepTests.java
```diff
@@ -0,0 +1,143 @@
...
+
+        CompletableFuture<WorkflowData> future = createIngestPipelineStep.execute(List.of(incorrectData));
+        assertTrue(future.isDone() && future.isCompletedExceptionally());
+
+        try {
+            future.get();
+        } catch (ExecutionException e) {
+            assertTrue(e.getMessage().contains("Failed to create ingest pipeline, required inputs not found"));
+        }
+    }
```
owaiskazi19, line 141: Same as above.

owaiskazi19 (commented):
Overall LGTM with few comments.

src/main/java/org/opensearch/flowframework/workflow/CreateIngestPipelineStep.java
```diff
@@ -0,0 +1,201 @@
...
+        String inputFieldName = null;
+        String outputFieldName = null;
+        BytesReference configuration = null;
+
+        // Extract required content from workflow data and generate the ingest pipeline configuration
+        for (WorkflowData workflowData : data) {
+
+            Map<String, String> parameters = workflowData.getParams();
+            Map<String, Object> content = workflowData.getContent();
+            logger.info("Previous step sent params: {}, content: {}", parameters, content);
```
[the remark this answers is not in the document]
joshpalis: sure Ill go ahead and remove this

src/main/java/org/opensearch/flowframework/workflow/CreateIngestPipelineStep.java
joshpalis: Sure, once the global context implementation has been added, we can revisit this

src/test/java/org/opensearch/flowframework/workflow/CreateIngestPipeline/CreateIngestPipelineStepTests.java
joshpalis: Sure this looks like a better way to test this

src/main/java/org/opensearch/flowframework/workflow/CreateIngestPipelineStep.java
```diff
@@ -0,0 +1,191 @@
...
+            .startObject(type)
+            .field(MODEL_ID_FIELD, modelId)
+            .startObject(FIELD_MAP)
+            .field(inputFieldName, outputFieldName)
+            .endObject()
+            .endObject()
+            .endObject()
+            .endArray()
+            .endObject();
+    }
```
owaiskazi19, line 190: I'm not sure if this is the right place to build the XContent for the pipeline. I would say let's create a separate model class but we can iterate on this later.

joshpalis: https://github.com/opensearch-project/opensearch-ai-flow-framework/issues/55

owaiskazi19 (approved):

Issue #45, opened 2023-09-18
[FEATURE] Include timeout settings as part of the user template


dbwiddis:
### Is your feature request related to a problem?

Workflow steps (and the whole workflow) need timeouts. These should be easily configurable by the user in case they expect slow API responses from complex AI models.

### What solution would you like?

A standard field in the template:
 - in user input for global timeout
 - in node section of workflow for individual step timeout

### What alternatives have you considered?
 - guessing and hardcoding values
 - storing them in OpenSearch settings
 - grabbing them from the API calls

Issue #24, opened 2023-09-07
[FEATURE] Create an ingest pipeline workflow step


joshpalis:
### Is your feature request related to a problem?
Implement workflow set up interface to facilitate the creation of an ingest pipeline given the pipeline name, list of processors and their associated inputs.

This first implementation for the semantic search use case should construct an ingestion pipeline using the `text_embedding` pre-processor which handles transforming `IngestDocument`s into vector embeddings prior to being indexed into a KNN index and stores the pipeline ID within the `GlobalContext` index. This pipeline ID will then be used to set as the `default_pipeline` of the KNN index.

The implementation should parse the use case template to retrieve the following inputs for the `text_embedding` pre-processor:
- input_field_name
- output_field_name

The implementation should leverage the provided client to read the `model_id` from the global context.

Using this information, create the `PutPipelineRequest` payload:
```
PUT _ingest/pipeline/neural-pipeline
{
  "description": "An example neural search pipeline",
  "processors" : [
    {
      "text_embedding": {
        "model_id": "<model_id>",
        "field_map": {
           "<input_field_name>": "<output_field_name>"
        }
      }
    }
  ]
}
```

### Do you have any additional context?
Creation of the ingest pipeline should be done via a `PutPipelineRequest` using the transport client. Reference : https://github.com/opensearch-project/geospatial/blob/main/src/main/java/org/opensearch/geospatial/action/upload/geojson/PipelineManager.java#L92

owaiskazi19:
@joshpalis this issue shouldn't just talk about neural pipeline but any ingest pipeline which we can create. There should be a type associated while creating the pipeline. If it's `neural` and the processor is  `text_embedding` the above pipeline would be created then.

PR #47, opened 2023-09-19
Add WorkflowStep Factory and implement XContent-based Template Parsing

### LINKED ISSUE #41: [FEATURE] Create factory class for Workflow Steps

Continuation of work started in #26.

While parsing a template, we create a `ProcessNode` that will manage the actual execution of the `WorkflowStep` with some optional initial `WorkflowData`.  There is a placeholder conditional doing this for a demo "create index" step, but this needs to be more scalable.

1. Add two fields to the nodes in the template:
    - one identifies the class type to generate.  
    - one identifies input data needed.
2. Pass this class type string to a factory class that creates an instance of the corresponding class.
3. Parse the input data field (if any) to generate a WorkflowData anonymous subclass.
4. Use these elements when instantiating a ProcessNode.

Several more complex ways of handling it such as using injection, reflection, pre-parsing the template to populate a map, etc.  Since multiple nodes may share the same class type, singletons or map lookups won't work.

We may iterate over this design and modify it later, but it seems the best direction to start wtih.

### LINKED ISSUE #42: [FEATURE] Make Thread Pool available to Process Node

The `ProcessNode.execute()` method introduced in #26 executes a workflow step asynchronously.  This execution should share resources with the OpenSearch thread pool.

PR #38 will be adding the create components extension point which includes the Thread Pool.  This pool needs to be accessible in some way to the Process Node class.  

Presently the `TemplateParser` class uses static methods, but this class can probably be turned into a singleton, instantiated as part of the createComponents method execution with a copy of the thread pool, that it can pass on to the ProcessNode in its constructor.

There may be other ways to pass around the threadpool in a static or singleton class somewhere.

Using the OpenSearch Thread Pool is pretty much mandatory for properly sharing resources and tracking thread stats / hot threads, etc.

### LINKED ISSUE #46: [DISCUSS] How should we parse the Template JSON?
Coming from https://github.com/opensearch-project/opensearch-ai-flow-framework/pull/26/files#r1329332685

In my initial template parsing implementation, I used the [GSON](https://github.com/google/gson) library.  It's Apache 2.0 open source.  The latest version has [no security vulnerabilites](https://security.snyk.io/package/maven/com.google.code.gson:gson).

Given that we will be using clients, another option is the Jakarta [`JsonParser`](https://jakarta.ee/specifications/jsonp/2.1/).  It's licensed under EPLv2 (since Eclipse took over javax) and GPLv2 with classpath exception (from the Oracle days)  so we'd need to add to our NOTICE file if we used it.  It's probably "closer to the Java spec" though and would still be `javax` if Oracle didn't claim trademark on it.

@owaiskazi19 proposed using [`ObjectMapper`](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/mapper/ObjectMapper.java) from OpenSearch, but that class is marked `@opensearch.internal` and is a lot more heavyweight.  It does give us more XContent style flexibility though.  I think that there may be API versions that use that under the hood, such as `org.opensearch.core.xcontent.XContentParser`.  

Jackson also has an ObjectMapper class (and frequent CVEs) and we should not use it.  Which discussion let me to this issue which led me to GSON as a recommendation for plugins: https://github.com/opensearch-project/OpenSearch/issues/5504 which led to https://github.com/opensearch-project/opensearch-plugins/issues/211 which stated:
> Based on the suggestions in the linked issue and my own research, a combination of GSON for JSON functionality and SnakeYAML for YAML functionality appears to be the best way forward.

Other one I'd consider is [json-simple](https://code.google.com/archive/p/json-simple/). It's also AL2.0 but it's very lightweight, small, no dependencies, and [no vulnerabilities](https://security.snyk.io/package/maven/com.googlecode.json-simple:json-simple). If we're just using it for parsing (like my implementation in #26) and not doing fancy interaction with java objects, it's probably the best choice.

Open to any other ideas.  Would like to get #26 merged with GSON for now and we can replace it later once this question is resolved.

### LINKED ISSUE #50: [BUG] ProcessNodeTest is flaky on CI
### What is the bug?
The below test passes on local but fails on CI.
```
org.opensearch.flowframework.template.ProcessNodeTests > testNode FAILED
    java.lang.AssertionError
        at __randomizedtesting.SeedInfo.seed([9B313DC0540DBCB3:C8934CDC5B151A88]:0)
        at org.junit.Assert.fail(Assert.java:87)
        at org.junit.Assert.assertTrue(Assert.java:42)
        at org.junit.Assert.assertTrue(Assert.java:53)
        at org.opensearch.flowframework.template.ProcessNodeTests.testNode(ProcessNodeTests.java:56)
  1> [2023-09-22T00:46:59,576][INFO ][o.o.f.t.ProcessNodeTests ] [testNode] before test
  1> [2023-09-22T00:46:59,767][INFO ][o.o.f.t.ProcessNode      ] [[Thread-6]] >>> Starting A.
  1> [2023-09-22T00:46:59,805][INFO ][o.o.f.t.ProcessNodeTests ] [testNode] after test
        at __randomizedtesting.SeedInfo.seed([9B313DC0540DBCB3:C8934CDC5B151A88]:0)
```

This fails on CI when checking whether future is complete on 
```
        assertTrue(f.isDone());
```

1. Run tests on Github CI
```
./gradlew ':test' --tests "org.opensearch.flowframework.template.ProcessNodeTests.testNode" -Dtests.seed=9B313DC0540DBCB3 -Dtests.security.manager=false -Dtests.locale=en-US -Dtests.timezone=Europe/Moscow -Druntime.java=17
```

Test should pass

### LINKED ISSUE #55: [FEATURE] Create an XContent parser for Processors in pipelines.

Template parsing implemented in #47 assumes only one layer of nesting (e.g., a map object in `Map<String, Object>` will always be a `Map<String, String>`.  This assumption isn't always true as shown in #44 where some manual content parsing was done.

Create an XContent parser for Processors in pipelines.
            
Representing the processor JSON (below) in a string in the User Template.  This would work but gets very ugly very fast as all the quotes need to be escaped.

> All pipeline configuration formats follow a similar pattern :
> ```
> {
>     "description" : "",
>     "processors": [
>         {
>              "processor1" : {
>                  // required inputs
>              }
>          },
>          {
>              "processor2" : {
>                  // required inputs
>              }
>          },
>          ....
>          ...
>     ]
> }
> ```
> 
> We can definitely generalize the creation of the configuration as long as the workflow data has a list of processors and their required inputs. The processors themselves vary in terms of format, so doing it this way would require some xcontent builder for each type of processor, since we'll need to write to specific fields.

_Originally posted by @joshpalis in https://github.com/opensearch-project/opensearch-ai-flow-framework/pull/44#discussion_r1334813698_

dbwiddis:
### Description

1. Creates a `WorkflowStepFactory` class to map a template "step type" to the correct instance
2. Parses the entire Use Case Template into XContent
3. Process the user inputs section for the workflow into WorkflowData params
4. Processes the user inputs section for each node into WorkflowData content
5. Refactors code to separate parsing functionality (WorkflowNode et. al.) from execution functionality (ProcessNode et. al.).  Uses the WorfklowStep and Workflow Data when creating the ProcessNode.
6. Implements an executor service to use the OpenSearch thread pool for the async execution.  
7. Added tests.
8. Fixed the ProcessNodeTests bug

### Issues Resolved

Fixes #41 
Fixes #42
Fixes #46
Fixes #50  
Fixes #55


dbwiddis:
Flipping back to draft to integrate the commits from #24 and #44 into the factory class (and delete my demo version).

dbwiddis:
And flipping back to ready for review, after rebasing and also completing tasks from #42 and fixing #50.

src/main/java/org/opensearch/flowframework/workflow/Workflow.java
```diff
@@ -0,0 +1,165 @@
...
+ * This represents an object in the workflows section of a {@link Template}.
+ */
+public class Workflow implements ToXContentObject {
+
+    /** The template field name for workflow user params */
+    public static final String USER_PARAMS_FIELD = "user_params";
+    /** The template field name for workflow nodes */
+    public static final String NODES_FIELD = "nodes";
+    /** The template field name for workflow steps, an alternative name for nodes */
+    public static final String STEPS_FIELD = "steps";
```
joshpalis, line 37: Can `nodes` and `steps` be used interchangeably? If so, why not just use `nodes`.

joshpalis (commented):
initial pass, still need to go over the test classes

src/main/java/org/opensearch/flowframework/template/Template.java
```diff
@@ -0,0 +1,394 @@
...
+    /** The template field name for template compatibility with OpenSearch versions */
+    public static final String COMPATIBILITY_FIELD = "compatibility";
+    /** The template field name for template user inputs */
+    public static final String USER_INPUTS_FIELD = "user_inputs";
+    /** The template field name for template workflows */
+    public static final String WORKFLOWS_FIELD = "workflows";
+
+    private final String name;
+    private final String description;
+    private final String useCase; // probably an ENUM actually
```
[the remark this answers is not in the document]
dbwiddis: Until the enums exist in this repo I can't parse to them.  Once they are added this is an easy `.valueOf()` change.

src/main/java/org/opensearch/flowframework/template/Template.java
```diff
@@ -0,0 +1,394 @@
...
+        for (Entry<String, Workflow> e : workflows.entrySet()) {
+            xContentBuilder.field(e.getKey(), e.getValue(), params);
+        }
+        xContentBuilder.endObject();
+
+        return xContentBuilder.endObject();
+    }
+
+    /**
+     * Parse raw json content into a workflow node instance.
```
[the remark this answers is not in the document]
dbwiddis: I always miss something in copypaste.

src/main/java/org/opensearch/flowframework/workflow/Workflow.java
dbwiddis: Why not allow either?  I'm ok, just was looking at the sample template that had both.

joshpalis: Sample template has been updated, as per @ohltyler let's stick to `nodes` so we are not using language that infers there cannot be graphs.

src/main/java/org/opensearch/flowframework/template/WorkflowNode.java
```diff
@@ -0,0 +1,185 @@
...
+            String fieldName = parser.currentName();
+            parser.nextToken();
+            switch (fieldName) {
+                case ID_FIELD:
+                    id = parser.text();
+                    break;
+                case TYPE_FIELD:
+                    type = parser.text();
+                    break;
+                case INPUTS_FIELD:
```
joshpalis, line 109: Perhaps we can have handling specifically for pipeline configurations within this `INPUTS_FIELDS`. We can update the use case template to generalize pipeline/processor configurations like so:

```
"processors" : [
    {
    type : "<processor_type>",
    params : [
        "field1" : "value",
        "field2" : "value",
        "field3" : "value",
    ]
}
]
```

We can then map the processor type to some template and pass the params there to build the configuration

dbwiddis: Agreed, we just need a predictable way to identify when and where in the template tree we will use these.

The parser is written for the specific json committed with this PR, we can continue to update and iterate on that.

dbwiddis: Just had a thought, if this is part of another plugin, it might already have its NamedXContent loaded into opensearch and available for us to parse with.

dbwiddis:
To ease review, a general summary of the main part of this PR:
 - Parsing.  Starts with Template, which also parses workflow, which also parses WorkflowEdge and WorkflowNode.
 - Process Sorting: takes the WorflowNodes and WorkflowEdges and creates the sorted ProcessNode sequence with links to predecessor ProcessNodes (no change from previous code, just separated out the parsing part)
 - WorkflowStepFactory: takes the node "type" field and generates a unique instance of the correct workflow step

Additional changes:
 - Made the sorter and step factory singletons taking the needed arguments that will go into the process nodes and workflow steps
 - Passed an executor service around to have better control of threading.  This can be changed in the future away from the generic one.
 - Misc. cleanup of warnings (casting, deprecation)

src/main/java/org/opensearch/flowframework/template/WorkflowNode.java
dbwiddis: Yep, there's a `ProcessorInfo` class [that's already implemented this](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/ingest/ProcessorInfo.java)

joshpalis: `ProcessorInfo` just gives us the list of processor types, not the configuration itself.

src/main/java/org/opensearch/flowframework/workflow/Workflow.java
owaiskazi19: Let's just have `nodes`?

src/main/java/org/opensearch/flowframework/workflow/Workflow.java
```diff
@@ -0,0 +1,165 @@
...
+            }
+
+        }
+        if (nodes.isEmpty()) {
+            throw new IOException("A workflow must have at least one node.");
+        }
+        if (edges.isEmpty()) {
+            // infer edges from sequence of nodes
+            // Start iteration at 1, will skip for a one-node array
+            for (int i = 1; i < nodes.size(); i++) {
```
owaiskazi19: Style/Nit: How about starting from `0` and then doing `i` and `i+1` respectively below?

src/main/java/org/opensearch/flowframework/workflow/WorkflowStepFactory.java
```diff
@@ -0,0 +1,96 @@
...
+ * Generates instances implementing {@link WorkflowStep}.
+ */
+public class WorkflowStepFactory {
+
+    private static WorkflowStepFactory instance = null;
+
+    private final Map<String, WorkflowStep> stepMap = new HashMap<>();
+
+    /**
+     * Create the singleton instance of this class. Throws an {@link IllegalStateException} if already created.
```
owaiskazi19, line 30: I'm just trying to think instead of creating singleton of this class, if we can use some extension point from Plugin.java to achieve the singleton behavior.

owaiskazi19 (commented):
Initial Pass! Thanks for adding XContent functionality to the project

src/main/java/org/opensearch/flowframework/FlowFrameworkPlugin.java
```diff
@@ -48,9 +46,9 @@ public Collection<Object> createComponents(
         IndexNameExpressionResolver indexNameExpressionResolver,
         Supplier<RepositoriesService> repositoriesServiceSupplier
     ) {
-        this.client = client;
-        CreateIngestPipelineStep createIngestPipelineStep = new CreateIngestPipelineStep(client);
-        CreateIndexStep createIndexStep = new CreateIndexStep(client);
-        return ImmutableList.of(createIngestPipelineStep, createIndexStep);
+        WorkflowStepFactory workflowStepFactory = WorkflowStepFactory.create(client);
+        WorkflowProcessSorter workflowProcessSorter = WorkflowProcessSorter.create(workflowStepFactory, threadPool.generic());
```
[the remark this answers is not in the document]
dbwiddis: I agree, we should pass threadpool, or better yet just use the injected version of it in a transport action.  I briefly went down that road but in the interest of getting the bug (#50) fixed it turned out doing tests with the thread pool version was far more complicated than I wanted to deal with, as I think that'll require IT rather than unit test.

src/main/java/org/opensearch/flowframework/template/ProcessNode.java
```diff
@@ -152,34 +134,20 @@ public CompletableFuture<WorkflowData> execute() {
             }
             CompletableFuture<WorkflowData> stepFuture = this.workflowStep.execute(input);
             try {
-                stepFuture.join();
+                stepFuture.orTimeout(15, TimeUnit.SECONDS).join();
```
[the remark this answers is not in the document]
dbwiddis: Already an issue on this, #45

src/main/java/org/opensearch/flowframework/template/Template.java
```diff
@@ -0,0 +1,394 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
+
+import static org.opensearch.core.xcontent.XContentParserUtils.ensureExpectedToken;
+
+/**
+ * The Template is the central data structure which configures workflows. This object is used to parse JSON communicated via REST API.
+ */
+public class Template implements ToXContentObject {
```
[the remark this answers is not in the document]
dbwiddis: I have all the xContent parsing classes under `o.o.f.template` right now.  I don't mind a package name change.

src/main/java/org/opensearch/flowframework/template/Template.java
```diff
@@ -0,0 +1,394 @@
...
+     * @return a YAML representation of the template.
+     */
+    public String toYaml() {
+        try {
+            XContentBuilder builder = YamlXContent.contentBuilder();
+            return this.toXContent(builder, EMPTY_PARAMS).toString();
+        } catch (IOException e) {
+            return "error: couldn't create YAML: " + e.getMessage();
+        }
+    }
```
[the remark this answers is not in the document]
dbwiddis: Sorta yeah.  So is the toJson. But why not?  It's a different flavor of toString that is useful. I really do like the yaml output to "validate" the xcontent parsing visually.

src/main/java/org/opensearch/flowframework/template/Template.java
```diff
@@ -0,0 +1,394 @@
...
+            + operations
+            + ", templateVersion="
+            + templateVersion
+            + ", compatibilityVersion="
+            + compatibilityVersion
+            + ", userInputs="
+            + userInputs
+            + ", workflows="
+            + workflows
+            + "]";
```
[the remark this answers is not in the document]
dbwiddis: I just used the IDE auto-generated toString here.  I'm not picky about what goes into it, and frankly I'd probably never use it (preferring toJson or toYaml).

src/main/java/org/opensearch/flowframework/template/WorkflowNode.java
```diff
@@ -0,0 +1,185 @@
...
+        XContentBuilder xContentBuilder = builder.startObject();
+        xContentBuilder.field(ID_FIELD, this.id);
+        xContentBuilder.field(TYPE_FIELD, this.type);
+
+        xContentBuilder.startObject(INPUTS_FIELD);
+        for (Entry<String, Object> e : inputs.entrySet()) {
+            xContentBuilder.field(e.getKey());
+            if (e.getValue() instanceof String) {
+                xContentBuilder.value(e.getValue());
+            } else if (e.getValue() instanceof Map<?, ?>) {
```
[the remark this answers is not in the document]
dbwiddis: Instanceof doesn't take generic types (see "type erasure" in your favorite search engine), it's either this or `Map` with no types, which gives a warning.

src/main/java/org/opensearch/flowframework/workflow/Workflow.java
dbwiddis: Starting at 1 avoids adding a `-1` to `nodes.size()` which, with the spacing added by spotless, is 4 more bytes of space in the source code. Storage space is expensive! Back in 1960 a byte cost nearly $1!  It's cheaper now at $0.000076 per byte.

They're equivalent, and I prefer this one as I'm really iterating pairs and whether I pick the first or second of the pair isn't that important.  Happy to switch, though, I expect they compile to the same byte code.

src/main/java/org/opensearch/flowframework/workflow/WorkflowStepFactory.java
dbwiddis: I've added it to createcontents so it is injected as a singleton and available everywhere that is also injected.

The problem is that we haven't yet implemented getActions which injects the transport actions that will use this.

So yes this is temporary and should eventually be replaced by `@Inject` annotation.  I'm trying to not make this PR any bigger than it already is.

src/main/java/org/opensearch/flowframework/FlowFrameworkPlugin.java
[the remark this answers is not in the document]
owaiskazi19: Are you planning to take care of passing threadpool in #50 ?

src/main/java/org/opensearch/flowframework/model/PipelineProcessor.java
```diff
@@ -0,0 +1,101 @@
...
+     */
+    public String type() {
+        return type;
+    }
+
+    /**
+     * Get the processor params
+     * @return the params
+     */
+    public Map<String, String> params() {
```
owaiskazi19, line 98: Shouldn't we keep the type of params as `<String, Object>`?

owaiskazi19 (approved):
Looks good overall with a minor comment/question.

src/main/java/org/opensearch/flowframework/model/PipelineProcessor.java
joshpalis: Pipeline processor params are just the user provided inputs, so a `Map<String,String>` makes sense. The directional idea here is to take these params and build the configuration, mapping the provided fields and values to some template. We'll probably have to create a set of classes that implement `ToXContentObject` for each type of pipeline processor and pass these params there.

joshpalis (approved):

joshpalis:
Thanks @dbwiddis for adding the Pipeline Processor parsers!

src/main/java/org/opensearch/flowframework/model/PipelineProcessor.java
dbwiddis: The problem with an object value is you need to define what it is.  We could allow nested maps but at some point you have to stop the recursion.

PR #52, opened 2023-09-23
Adds Register and Deploy Model Step for remote model

owaiskazi19:
### Description
Adds Register, Deploy Model Step for the remote model. 

Also, adds GetTask step for the future work of registering a local model.
TODO:
1. Figure out a way to handle the verify method call in the tests.

### Issues Resolved
Part of https://github.com/opensearch-project/opensearch-ai-flow-framework/issues/21


src/test/java/org/opensearch/flowframework/workflow/DeployModelStepTests.java
```diff
@@ -0,0 +1,82 @@
...
+            ActionListener<MLDeployModelResponse> actionListener = invocation.getArgument(2);
+            MLDeployModelResponse output = new MLDeployModelResponse(taskId, mlTaskType, status);
+            actionListener.onResponse(output);
+            return null;
+        }).when(machineLearningNodeClient).deploy(eq("modelId"), actionListenerCaptor.capture());
+
+        CompletableFuture<WorkflowData> future = deployModel.execute(List.of(inputData));
+
+        // TODO: Find a way to verify the below
+        // verify(machineLearningNodeClient).deploy(eq(MLRegisterModelInput.class), actionListenerCaptor.capture());
```
owaiskazi19, line 77: @dbwiddis any idea here? Approaches I tried:
1 `MockedStatic` with mockito-inline 
2. Created ArgumentCaptor only for `MLDeployModelResponse`.
Although, I can see the method getting invoked in the debugger. Keeping it as a TODO here for now.

joshpalis (commented):
Initial pass, still need to go over the test cases

src/main/java/org/opensearch/flowframework/FlowFrameworkPlugin.java
```diff
@@ -51,7 +53,10 @@ public Collection<Object> createComponents(
         IndexNameExpressionResolver indexNameExpressionResolver,
         Supplier<RepositoriesService> repositoriesServiceSupplier
     ) {
-        WorkflowStepFactory workflowStepFactory = new WorkflowStepFactory(client);
+        Settings settings = environment.settings();
+        // TODO: Creating NodeClient is a temporary fix until we get the NodeClient from the provision API
+        NodeClient nodeClient = new NodeClient(settings, threadPool);
+        WorkflowStepFactory workflowStepFactory = new WorkflowStepFactory(client, nodeClient);
```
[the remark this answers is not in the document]
owaiskazi19: It's of type `Client` [here](https://github.com/opensearch-project/opensearch-ai-flow-framework/pull/52/files#diff-9d60d6086ce87e2240d7d6bcf8d2a8b79c73400fc6ec6865f74fc571e03bf1afR44). Typecasting it to `(NodeClient) client` fails the test. Anyway, it's temporary until we get the NodeClient from provision API.

src/main/java/org/opensearch/flowframework/workflow/DeployModelStep.java
```diff
@@ -0,0 +1,86 @@
...
+        };
+
+        String modelId = null;
+
+        for (WorkflowData workflowData : data) {
+            if (workflowData != null) {
+                Map<String, Object> content = workflowData.getContent();
+
+                for (Map.Entry<String, Object> entry : content.entrySet()) {
+                    if (entry.getKey() == MODEL_ID) {
```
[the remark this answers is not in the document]
owaiskazi19: Are you suggesting to use switch case for just one case?

src/main/java/org/opensearch/flowframework/workflow/DeployModelStep.java
```diff
@@ -0,0 +1,86 @@
...
+            public void onFailure(Exception e) {
+                logger.error("Model deployment failed");
+                deployModelFuture.completeExceptionally(e);
+            }
+        };
+
+        String modelId = null;
+
+        for (WorkflowData workflowData : data) {
+            if (workflowData != null) {
```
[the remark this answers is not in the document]
owaiskazi19: When tested the above step with Demo.java, workflowData has one null element and the next one has `getContent` and `getParams`. This is coming from WorkflowData. @dbwiddis might have more context on null.

src/main/java/org/opensearch/flowframework/workflow/RegisterModelStep.java
```diff
@@ -0,0 +1,154 @@
...
+        String modelName = null;
+        String modelVersion = null;
+        String modelGroupId = null;
+        String connectorId = null;
+        String description = null;
+        MLModelFormat modelFormat = null;
+        MLModelConfig modelConfig = null;
+
+        for (WorkflowData workflowData : data) {
+            if (workflowData != null) {
```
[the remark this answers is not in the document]
owaiskazi19: again, answered above :)

src/main/java/org/opensearch/flowframework/workflow/DeployModelStep.java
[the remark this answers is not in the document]
dbwiddis: `DemoWorkflowStep` line 40 is `future.complete(null);`.  That will not be encountered in prodcution.  The Demo should be changed to return `Workflow.EMPTY` and a null check is not needed here.

Somewhere I wrote some comments about not depending on the Demo code. It was written quickly and lazily. :-)

src/main/java/org/opensearch/flowframework/FlowFrameworkPlugin.java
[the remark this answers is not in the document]
dbwiddis: > It's of type `Client` [here](https://github.com/opensearch-project/opensearch-ai-flow-framework/pull/52/files#diff-9d60d6086ce87e2240d7d6bcf8d2a8b79c73400fc6ec6865f74fc571e03bf1afR44). 

And that is called from Node.java passing this object: `private final NodeClient client;`.

> Typecasting it to `(NodeClient) client` fails the plugin test. 

Can't you typecast the param you're sending to the `WorkflowStepFactory` constructor?

> Anyway, it's temporary until we get the NodeClient from provision API.

Not clear how this is relevant.  When this Plugin is loaded in OpenSearch, `client` will refer to the `NodeClient` instance initialized on the `Node`.

src/main/java/org/opensearch/flowframework/workflow/DeployModelStep.java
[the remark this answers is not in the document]
dbwiddis: > Are you suggesting to use switch case for just one case?

Will this always be only one case?

If so, `if` is fine.  If there's a possibility of more fields being added in the future, maintaining a consistent pattern will help future developers.

My guess is the JIT compiler will produce identical bytecode in either case.

src/main/java/org/opensearch/flowframework/workflow/RegisterModelStep.java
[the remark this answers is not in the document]
dbwiddis: Re-commenting here.   If you're getting `null` workflow data from anywhere, that's a bug.  `WorkflowData.EMPTY` exists for precisely the case where you have nothing to send.

We should not need null checks.  If we are encountering null here, change whatever is sending it to send `WorkflowData.EMPTY` instead of null.

src/test/java/org/opensearch/flowframework/workflow/DeployModelStepTests.java
dbwiddis: Hard to answer without live debugging.  In general my debugging suggestion is to try verify with `any(Object.class)` for the arguments.  If this fails, the problem is not in this line, but you're encountering a failure in `deploy()`.   

Toss up a new issue to debug this so we can get this merged.

owaiskazi19: This was the 3rd option I tried and forgot to mention. I'll create a new issue and link it here.

dbwiddis: Looked into this.  The `machineLearningNodeClient` you are verifying is the local mock from this test class.  However, when you call `deployModel.execute()` you are creating a new client with this line:
```java
MachineLearningNodeClient machineLearningNodeClient = MLClient.createMLClient(nodeClient);
```

src/main/java/org/opensearch/flowframework/FlowFrameworkPlugin.java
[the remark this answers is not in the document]
owaiskazi19: > Can't you typecast the param you're sending to the `WorkflowStepFactory` constructor?
I can. The tests are failing with the below error
```
org.opensearch.flowframework.workflow.WorkflowProcessSorterTests > classMethod FAILED
    java.lang.ClassCastException: class org.opensearch.client.Client$MockitoMock$zUodJfPz cannot be cast to class org.opensearch.client.node.NodeClient (org.opensearch.client.Client$MockitoMock$zUodJfPz and org.opensearch.client.node.NodeClient are in unnamed module of loader 'app')
        at __randomizedtesting.SeedInfo.seed([6CC035B077521807]:0)
        at org.opensearch.flowframework.workflow.WorkflowStepFactory.populateMap(WorkflowStepFactory.java:43)
        at org.opensearch.flowframework.workflow.WorkflowStepFactory.<init>(WorkflowStepFactory.java:37)
        at org.opensearch.flowframework.workflow.WorkflowProcessSorterTests.setup(WorkflowProcessSorterTests.java:67)

File: src/main/java/org/opensearch/flowframework/workflow/DeployModelStep.java (hunk 4 above)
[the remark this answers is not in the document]
Comment (owaiskazi19) [reply src/main/java/org/opensearch/flowframework/workflow/DeployModelStep.java]: That might be the reason. I fixed the null error.

File: src/main/java/org/opensearch/flowframework/FlowFrameworkPlugin.java (hunk 2 above)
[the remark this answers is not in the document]
Comment (dbwiddis) [reply src/main/java/org/opensearch/flowframework/FlowFrameworkPlugin.java]: > java.lang.ClassCastException: class org.opensearch.client.Client$MockitoMock$zUodJfPz cannot be cast to class org.opensearch.client.node.NodeClient 

I think the issue is that you are mocking `Client.class` rather than `NodeClient.class`.  Mockito doesn't seem to allow downcasting (and rightly so).

File: src/test/java/org/opensearch/flowframework/workflow/DeployModelStepTests.java (hunk 1 above)
Comment (owaiskazi19) [reply src/test/java/org/opensearch/flowframework/workflow/DeployModelStepTests.java]: Thanks for looking into this. Let me create a follow up PR for the above.

File: src/test/java/org/opensearch/flowframework/workflow/DeployModelStepTests.java [hunk 6]
```diff
@@ -0,0 +1,82 @@
...
+            ActionListener<MLDeployModelResponse> actionListener = invocation.getArgument(2);
+            MLDeployModelResponse output = new MLDeployModelResponse(taskId, mlTaskType, status);
+            actionListener.onResponse(output);
+            return null;
+        }).when(machineLearningNodeClient).deploy(eq("modelId"), actionListenerCaptor.capture());
+
+        CompletableFuture<WorkflowData> future = deployModel.execute(List.of(inputData));
+
+        // TODO: Find a way to verify the below
+        // verify(machineLearningNodeClient).deploy(eq("modelId"), actionListenerCaptor.capture());
```
joshpalis, line 77: If you change line 72 to :
`}).when(machineLearningNodeClient).deploy(any(), any());`

You can verify number of invocations like so : `verify(<mock>, times(<number of invocations>)).deploy(any(), actionListenerCapture.capture())`

And then assert the captured values
`assertEquals(<expected>, actionListenerCaptor.getValues().<MLDeployModelResponseGetterMethod()>)`

src/test/java/org/opensearch/flowframework/workflow/RegisterModelStepTests.java
```diff
@@ -0,0 +1,108 @@
...
+            ActionListener<MLRegisterModelResponse> actionListener = invocation.getArgument(2);
+            MLRegisterModelResponse output = new MLRegisterModelResponse(taskId, status, modelId);
+            actionListener.onResponse(output);
+            return null;
+        }).when(machineLearningNodeClient).register(any(MLRegisterModelInput.class), actionListenerCaptor.capture());
+
+        CompletableFuture<WorkflowData> future = registerModelStep.execute(List.of(inputData));
+
+        // TODO: Find a way to verify the below
+        // verify(machineLearningNodeClient).register(any(MLRegisterModelInput.class), actionListenerCaptor.capture());
```
joshpalis, line 102: same here

joshpalis (approved):
Looks good to me, just added a few suggestions regarding testing

src/test/java/org/opensearch/flowframework/workflow/DeployModelStepTests.java
owaiskazi19: Thanks for looking into it. Already tried the above approach and couple of other, it returns the below:
```java
  2> Wanted but not invoked:
    machineLearningNodeClient.deploy(
        "modelId",
        <Capturing argument: ActionListener>
    );
    -> at org.opensearch.flowframework.workflow.DeployModelStepTests.testDeployModel(DeployModelStepTests.java:79)
```

I have added the approaches tried [here](https://github.com/opensearch-project/opensearch-ai-flow-framework/pull/52#discussion_r1346548526). The issue is with mocking the [static](https://github.com/opensearch-project/opensearch-ai-flow-framework/blob/main/src/main/java/org/opensearch/flowframework/client/MLClient.java#L28) method.  It needs be mock like below:
```java
 try (MockedStatic<MLClient> mlClientMockedStatic = Mockito.mockStatic(MLClient.class)) {
      when(MLClient.getMLClient(any(NodeClient.class))).thenReturn(machineLearningNodeClient);
```
Which requires `mockito-inline` dependency and to avoid using that we have to use the below plugin `org.mockito.plugins.MockMaker ` with the extension `mock-maker-inline`. Tried both way which returns the below error
```
Caused by:
                org.mockito.exceptions.base.MockitoInitializationException: 
                Could not initialize inline Byte Buddy mock maker.
```
for which I added `net.bytebuddy` dependency  which again fails with the below

```
 Failed to load interface org.mockito.plugins.MockMaker implementation declared in java.lang.CompoundEnumeration@1f268988

File: src/test/java/org/opensearch/flowframework/workflow/RegisterModelStepTests.java (hunk 7 above)
Comment (owaiskazi19) [reply src/test/java/org/opensearch/flowframework/workflow/RegisterModelStepTests.java:102]: Answered above :)

File: src/main/java/org/opensearch/flowframework/client/MLClient.java [hunk 8]
```diff
@@ -22,12 +22,12 @@ private MLClient() {}
     /**
      * Creates machine learning client.
      *
-     * @param nodeClient node client of OpenSearch.
+     * @param client client of OpenSearch.
      * @return machine learning client from ml-commons.
      */
-    public static MachineLearningNodeClient createMLClient(NodeClient nodeClient) {
+    public static MachineLearningNodeClient createMLClient(Client client) {
```
jackiehanyang, line 28: what's the motivation of this client change?

src/main/java/org/opensearch/flowframework/workflow/DeployModelStep.java
```diff
@@ -0,0 +1,83 @@
...
+import java.util.concurrent.CompletableFuture;
+
+/**
+ * Step to deploy a model
+ */
+public class DeployModelStep implements WorkflowStep {
+    private static final Logger logger = LogManager.getLogger(DeployModelStep.class);
+
+    private Client client;
+    private static final String MODEL_ID = "model_id";
```
jackiehanyang, line 30: nit: consider moving constants to comon/CommonValue.java file

src/main/java/org/opensearch/flowframework/workflow/DeployModelStep.java
```diff
@@ -0,0 +1,83 @@
...
+                deployModelFuture.completeExceptionally(e);
+            }
+        };
+
+        String modelId = null;
+
+        for (WorkflowData workflowData : data) {
+            Map<String, Object> content = workflowData.getContent();
+            for (Map.Entry<String, Object> entry : content.entrySet()) {
+                if (entry.getKey() == MODEL_ID) {
```
jackiehanyang, line 69: move MODEL_ID to the front and use `.equals` for string comparison to avoid potential NPE
```suggestion
                if (MODEL_ID.equals(entry.getKey())) {
```

src/main/java/org/opensearch/flowframework/workflow/DeployModelStep.java
```diff
@@ -0,0 +1,83 @@
...
+                    modelId = (String) content.get(MODEL_ID);
+                }
+
+            }
+        }
```
jackiehanyang, line 74: I think this loop for fetching model_id can be optimized. How about break out of the loop once we've found the model_id to avoid unnecessary iterations. Also, the inner loop ins't needed since we can just query the map directly. 
```suggestion
        for (WorkflowData workflowData : data) {
            if (workflowData.getContent().containsKey(MODEL_ID)) {
                modelId = (String) workflowData.getContent().get(MODEL_ID);
                break;
            }
        }
```

src/main/java/org/opensearch/flowframework/client/MLClient.java
owaiskazi19: MLClient should take care of making the client as NodeClient
```java
@Generated
    public MachineLearningNodeClient(Client client) {
        this.client = client;
    }

File: src/main/java/org/opensearch/flowframework/workflow/DeployModelStep.java [hunk 12]
```diff
@@ -0,0 +1,81 @@
...
+
+        String modelId = null;
+
+        for (WorkflowData workflowData : data) {
+            if (workflowData.getContent().containsKey(MODEL_ID)) {
+                modelId = (String) workflowData.getContent().get(MODEL_ID);
+                break;
+            }
+        }
+        machineLearningNodeClient.deploy(modelId, actionListener);
```
jackiehanyang, line 73: There's a possibility that the `model_id` remains null, how ml-commons is handling this case? I think this might result in a NPE or other unintended consequences. Maybe consider adding a check for null and fail early to handle it appropriately?

jackiehanyang (approved):
overall lgtm! I'm also ok if you want to address the `model_id` might remains null when calling `deploy` action in the following pr as well :)

PR #65, opened 2023-09-30
Introduce global-context index and related operations

jackiehanyang:
### Description
- Add global-context index as an enum and its mapping
- Add global-context index handler that contains methods like initIfAbsent, put, and update
- Add custom exception class
- Update Template parser to add `responses` and `resources_created` fields
- Disable checkStyleTest

### Issues Resolved
https://github.com/opensearch-project/opensearch-ai-flow-framework/issues/51


src/main/java/org/opensearch/flowframework/workflow/ProcessNode.java
```diff
@@ -130,7 +131,12 @@ public CompletableFuture<WorkflowData> execute() {
                     return;
                 }
             }
-            CompletableFuture<WorkflowData> stepFuture = this.workflowStep.execute(input);
+            CompletableFuture<WorkflowData> stepFuture = null;
+            try {
+                stepFuture = this.workflowStep.execute(input);
+            } catch (IOException e) {
+                throw new RuntimeException(e);
+            }
```
dbwiddis, line 139: This section is entirely rewritten in #66, you may want to leave this class alone in your PR.

src/test/java/org/opensearch/flowframework/indices/GlobalContextHandlerTests.java
```diff
@@ -0,0 +1,108 @@
...
+
+        globalContextHandler.storeResponseToGlobalContext(documentId, updatedFields, listener);
+
+        ArgumentCaptor<UpdateRequest> requestCaptor = ArgumentCaptor.forClass(UpdateRequest.class);
+        verify(client).update(requestCaptor.capture(), any());
+
+        assertEquals(GLOBAL_CONTEXT_INDEX, requestCaptor.getValue().index());
+        assertEquals(documentId, requestCaptor.getValue().id());
+    }
+}
```
dbwiddis, line 108: Be sure to run spotless before committing, you're missing an EOF.

src/main/java/org/opensearch/flowframework/exception/FlowFrameworkException.java
```diff
@@ -0,0 +1,40 @@
...
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+package org.opensearch.flowframework.exception;
+
+/**
+ * Representation of Flow Framework Exceptions
+ */
+public class FlowFrameworkException extends RuntimeException {
```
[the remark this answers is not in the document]
jackiehanyang: Sorry forgot to include that change in this pr. Update the class with RestStatus, and returning proper RestStatus when using this exception class.

src/main/java/org/opensearch/flowframework/function/ThrowingSupplier.java
```diff
@@ -0,0 +1,26 @@
...
+package org.opensearch.flowframework.function;
+
+/**
+ * A supplier that can throw checked exception
+ *
+ * @param <T> method parameter type
+ * @param <E> Exception type
+ */
+@FunctionalInterface
+public interface ThrowingSupplier<T, E extends Exception> {
```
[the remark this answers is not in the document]
jackiehanyang: The purpose of adding this class is to throw checked exceptions in static initialization blocks where Java doesn't allow to do it, enum constructors in this case. I prefer to not use `Callable` because it introduces an exception when we use `Callable.call()`. Therefore, in order to capture any exception thrown by `mappingCallable.call()`, we need to assign a default value to the `mapping`, like `null`, or logging it. I think it's more complicated to handle. But using this `ThrowingSupplier`, we don't need to care about all this. I see AD is using the same method for their Index Enum class. ML-Commons is not using any wrapper because they don't load index mapping from a json file, so they don't need worry about handle exceptions in an enum class

src/main/java/org/opensearch/flowframework/function/ThrowingSupplierWrapper.java
```diff
@@ -0,0 +1,43 @@
...
+    public static <T> Supplier<T> throwingSupplierWrapper(ThrowingSupplier<T, Exception> throwingSupplier) {
+
+        return () -> {
+            try {
+                return throwingSupplier.get();
+            } catch (Exception ex) {
+                throw new RuntimeException(ex);
+            }
+        };
+    }
```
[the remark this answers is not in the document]
jackiehanyang: responded this suggested change above.

src/main/java/org/opensearch/flowframework/indices/FlowFrameworkIndex.java
```diff
@@ -0,0 +1,50 @@
...
+import static org.opensearch.flowframework.constant.CommonValue.GLOBAL_CONTEXT_INDEX;
+import static org.opensearch.flowframework.constant.CommonValue.GLOBAL_CONTEXT_INDEX_VERSION;
+
+/**
+ * An enumeration of Flow Framework indices
+ */
+public enum FlowFrameworkIndex {
+    GLOBAL_CONTEXT(
+        GLOBAL_CONTEXT_INDEX,
+        ThrowingSupplierWrapper.throwingSupplierWrapper(GlobalContextHandler::getGlobalContextMappings),
```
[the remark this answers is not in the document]
jackiehanyang: responded this suggested change above.

src/main/java/org/opensearch/flowframework/workflow/CreateIndexStep.java
```diff
@@ -96,15 +117,124 @@ public String getName() {
...
     }
+
+    /**
+     * Check if we should update index based on schema version.
+     * @param indexName index name
+     * @param newVersion new index mapping version
+     * @param listener action listener, if update index is needed, will pass true to its onResponse method
+     */
+    @VisibleForTesting
+    protected void shouldUpdateIndex(String indexName, Integer newVersion, ActionListener<Boolean> listener) {
```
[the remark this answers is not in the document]
jackiehanyang: removed @VisibleForTesting and made method private

src/main/java/org/opensearch/flowframework/function/ThrowingSupplier.java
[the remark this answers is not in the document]
dbwiddis: Right, all the wrapping handles the use in static initialization (see later comments with wrapped callable suggestion).  Both work, it's probably more of a usage/style question where a "supplier" tends to be retrieving a value that exists somewhere close/fast while a "callable" needs to call out somewhere else to fetch a value and may be slower, and I feel that was the case here.

src/main/java/org/opensearch/flowframework/workflow/CreateIndexStep.java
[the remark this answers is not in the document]
dbwiddis: It's ok to keep it (or change it to) package private if you need to for testing, but if you can test it while private, by testing a public method, that's preferred.

build.gradle
```diff
@@ -11,6 +11,7 @@ apply plugin: 'com.diffplug.spotless'
 apply from: 'formatter/formatting.gradle'
 // for javadocs and checks spotless doesn't do
 apply plugin: 'checkstyle'
+checkstyleTest.enabled = false
```
dbwiddis, line 14: The checkstyle failures were because you are using star imports, which we intentionally forbid for [Reasons](https://www.javadude.com/posts/20040522-import-on-demand-is-evil/).

dbwiddis: You can suppress test path in the config, see https://github.com/opensearch-project/opensearch-ai-flow-framework/blob/main/config/checkstyle/checkstyle_suppressions.xml where javadocs are suppressed rather than turning the whole thing off.  If we really want to allow * in test we could add it there.

But with modern IDEs I really don't think it's necessary. When writing code I put the * in there, and then let my IDE sort out what to actually import.

dbwiddis: I'm also fine with you turning it off for draft while you develop and fixing it at the end, I recognize this PR is a work in progress!

src/main/java/org/opensearch/flowframework/indices/GlobalContextHandler.java
```diff
@@ -0,0 +1,119 @@
...
+     * Get global-context index mapping
+     * @return global-context index mapping
+     * @throws IOException if mapping file cannot be read correctly
+     */
+    public static String getGlobalContextMappings() throws IOException {
+        return getIndexMappings(GLOBAL_CONTEXT_INDEX_MAPPING);
+    }
+
+    private void initGlobalContextIndexIfAbsent(ActionListener<Boolean> listener) {
+        createIndexStep.initIndexIfAbsent(FlowFrameworkIndex.GLOBAL_CONTEXT, listener);
```
joshpalis, line 66: Curious why the functionality to create our system indices is within the `CreateIndexStep`, I would assume the workflow steps are only used during template execution to set up the infrastructure of the builder's application. Perhaps lets have a separate index utility class that both the `CreateIndexStep` and `GlobalContextHandler` can use.

joshpalis: Maybe something like the [`IndexManagement`](https://github.com/opensearch-project/anomaly-detection/blob/main/src/main/java/org/opensearch/timeseries/indices/IndexManagement.java) class in AD : we can keep all the logic we need for managing the lifecycle of our plugin's system indices in a single class.

jackiehanyang: I had a separate class like `IndexManagement` in my previous pr. But after @owaiskazi19 `CreateIndexStep` pr, Owais suggested me to move methods in the `IndexManagement` file under this file. I don't have a strong preference, but I closed my previous pr for this refactoring reason

src/main/java/org/opensearch/flowframework/exception/FlowFrameworkException.java
```diff
@@ -0,0 +1,51 @@
...
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+package org.opensearch.flowframework.exception;
+
+import javax.ws.rs.core.Response;
```
dbwiddis, line 11: Let's use the same status type as we'll be using in the Rest Handlers.
```suggestion
import org.opensearch.core.rest.RestStatus;
```

src/test/java/org/opensearch/flowframework/workflow/CreateIndexStepTests.java
```diff
@@ -89,4 +112,34 @@ public void testCreateIndexStepFailure() throws ExecutionException, InterruptedE
...
+    //
+    // ActionListener<Boolean> listener = mock(ActionListener.class);
+    // createIndexStep.initIndexIfAbsent(index, listener);
+    //
+    // ArgumentCaptor<PutMappingRequest> captor = ArgumentCaptor.forClass(PutMappingRequest.class);
+    // verify(indicesAdminClient).putMapping(captor.capture());
+    //
+    // PutMappingRequest capturedRequest = captor.getValue();
+    // assertEquals(index.getIndexName(), capturedRequest.indices()[0]);
+    // }
```
dbwiddis, line 143: Is this PR still a draft/WIP?  What are all these commented-out lines?

src/main/java/org/opensearch/flowframework/indices/GlobalContextHandler.java
joshpalis: Additionally, we'll want to register the global context index with OpenSearch using this [`SystemIndexPlugin`](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/plugins/SystemIndexPlugin.java). We'll have to make the `FlowFrameworkPlugin` class extend this.

More information on how to correctly handle system indices is called out [here](https://github.com/opensearch-project/job-scheduler/issues/466)

src/main/resources/mappings/global-context.json
```diff
@@ -0,0 +1,69 @@
...
+        },
+        "compatibility": {
+          "type": "integer"
+        }
+      }
+    },
+    "user_inputs": {
+      "type": "nested",
+      "properties": {
+        "model_id": {
```
amitgalitz, line 42: Should we remove these specific fields, we might not want to have fields here that are specific to a use case, and make the fields for the GC mapping completely use-case independent

build.gradle
```diff
@@ -107,6 +107,7 @@ dependencies {
     implementation "org.opensearch:opensearch:${opensearch_version}"
     implementation 'org.junit.jupiter:junit-jupiter:5.10.0'
     implementation "com.google.guava:guava:32.1.2-jre"
+    implementation 'javax.ws.rs:javax.ws.rs-api:2.1.1'
```
dbwiddis, line 110: I think this dependency is only here for the `Response.Status` used for the exception, but you can use `org.opensearch.core.rest.RestStatus` for that and remove this dependency.
```suggestion
```

src/main/java/org/opensearch/flowframework/exception/FlowFrameworkException.java
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * Representation of Flow Framework Exceptions
+ */
+public class FlowFrameworkException extends RuntimeException {
```
dbwiddis, line 16: By inheritance this class implements `Serializable` so it needs a `serialVersionUID`.
```suggestion
public class FlowFrameworkException extends RuntimeException {

    private static final long serialVersionUID = 1L;
```

src/test/java/org/opensearch/flowframework/indices/GlobalContextHandlerTests.java
```diff
@@ -0,0 +1,109 @@
...
+        when(client.admin()).thenReturn(adminClient);
+    }
+
+    public void testPutTemplateToGlobalContext() throws IOException {
+        Template template = mock(Template.class);
+        when(template.toXContent(any(XContentBuilder.class), eq(ToXContent.EMPTY_PARAMS))).thenAnswer(invocation -> {
+            XContentBuilder builder = invocation.getArgument(0);
+            return builder;
+        });
+        ActionListener<IndexResponse> listener = mock(ActionListener.class);
```
dbwiddis, line 79:
```suggestion
        @SuppressWarnings("unchecked")
        ActionListener<IndexResponse> listener = mock(ActionListener.class);
```
(And also line 99)

src/test/java/org/opensearch/flowframework/workflow/CreateIndexStepTests.java
```diff
@@ -89,4 +120,67 @@ public void testCreateIndexStepFailure() throws ExecutionException, InterruptedE
         assertTrue(ex.getCause() instanceof Exception);
         assertEquals("Failed to create an index", ex.getCause().getMessage());
     }
+
+    public void testInitIndexIfAbsent_IndexNotPresent() {
+        when(metadata.hasIndex(FlowFrameworkIndex.GLOBAL_CONTEXT.getIndexName())).thenReturn(false);
+
+        ActionListener<Boolean> listener = mock(ActionListener.class);
```
dbwiddis, line 127:
```suggestion
                @SuppressWarnings("unchecked")
                ActionListener<Boolean> listener = mock(ActionListener.class);
```
And the other mocks in this class of ActionListener and Map.

src/test/java/org/opensearch/flowframework/workflow/CreateIndexStepTests.java
```diff
@@ -89,4 +120,67 @@ public void testCreateIndexStepFailure() throws ExecutionException, InterruptedE
...
+        mockMetaMapping.put(SCHEMA_VERSION_FIELD, 1);
+        mockMapping.put(META, mockMetaMapping);
+        MappingMetadata mockMappingMetadata = mock(MappingMetadata.class);
+        when(mockIndexMetadata.mapping()).thenReturn(mockMappingMetadata);
+        when(mockMappingMetadata.getSourceAsMap()).thenReturn(mockMapping);
+
+        createIndexStep.initIndexIfAbsent(index, listener);
+
+        ArgumentCaptor<PutMappingRequest> putMappingRequestArgumentCaptor = ArgumentCaptor.forClass(PutMappingRequest.class);
+        ArgumentCaptor<ActionListener> listenerCaptor = ArgumentCaptor.forClass(ActionListener.class);
```
dbwiddis, line 161:
```suggestion
        @SuppressWarnings({ "unchecked", "deprecation" })
        ArgumentCaptor<ActionListener<AcknowledgedResponse>> listenerCaptor = ArgumentCaptor.forClass(ActionListener.class);        
```

src/main/java/org/opensearch/flowframework/workflow/CreateIndexStep.java
```diff
@@ -96,15 +119,132 @@ public String getName() {
         return NAME;
     }
 
+    /**
```
owaiskazi19, line 122: It's fine to keep the above CreateIndexRequest code (L82:86) here and later we can follow the same path like ml-commons for CreateIndexStep to create an index.

src/main/java/org/opensearch/flowframework/workflow/CreateIndexStep.java
```diff
@@ -96,15 +119,132 @@ public String getName() {
...
+                                        } else {
+                                            internalListener.onFailure(
+                                                new FlowFrameworkException(
+                                                    "Failed to update index: " + indexName,
+                                                    Response.Status.INTERNAL_SERVER_ERROR
+                                                )
+                                            );
+                                        }
+                                    }, exception -> {
+                                        logger.error("Failed to update index " + indexName, exception);
```
owaiskazi19, line 188: Here and for the rest
```suggestion
                                        logger.error(() ->  new ParameterizedMessage("Failed to update index {}", indexName), exception);
```

src/main/java/org/opensearch/flowframework/workflow/CreateIndexStep.java
```diff
@@ -96,15 +119,132 @@ public String getName() {
...
+                        logger.error("Failed to update index mapping", e);
+                        internalListener.onFailure(e);
+                    }));
+                } else {
+                    // No need to update index if it's already updated.
+                    internalListener.onResponse(true);
+                }
+            }
+        } catch (Exception e) {
+            logger.error("Failed to init index " + indexName, e);
```
owaiskazi19, line 207: ParamterizedMessage can be used it

src/main/java/org/opensearch/flowframework/workflow/ProcessNode.java
owaiskazi19: Ignoring this file for review based on the above comment.

src/main/resources/mappings/global-context.json
owaiskazi19: `model_id` would be generated after we register a model and won't be a user input. We can remove this.

src/test/java/org/opensearch/flowframework/indices/GlobalContextHandlerTests.java
```diff
@@ -0,0 +1,109 @@
...
+        doAnswer(invocation -> {
+            ActionListener<Boolean> callback = invocation.getArgument(1);
+            callback.onResponse(true);
+            return null;
+        }).when(createIndexStep).initIndexIfAbsent(any(), any());
+
+        globalContextHandler.putTemplateToGlobalContext(template, listener);
+
+        ArgumentCaptor<IndexRequest> requestCaptor = ArgumentCaptor.forClass(IndexRequest.class);
+        verify(client).index(requestCaptor.capture(), any());
```
owaiskazi19, line 90: Nit/Style: I like to add `times` with verify for specific number of invocations :)
```suggestion
        verify(client, times(1)).index(requestCaptor.capture(), any());
```

src/test/java/org/opensearch/flowframework/workflow/CreateIndexStepTests.java
```diff
@@ -10,54 +10,88 @@
...
 
-        when(adminClient.indices()).thenReturn(indicesAdminClient);
+        when(client.threadPool()).thenReturn(threadPool);
+        when(threadPool.getThreadContext()).thenReturn(threadContext);
         when(client.admin()).thenReturn(adminClient);
+        when(adminClient.indices()).thenReturn(indicesAdminClient);
+        when(clusterService.state()).thenReturn(ClusterState.builder(new ClusterName("test cluster")).build());
+        when(metadata.indices()).thenReturn(Map.of(GLOBAL_CONTEXT_INDEX, indexMetadata));
 
+        createIndexStep = new CreateIndexStep(clusterService, client);
```
owaiskazi19, line 90: Can we mock CreateIndexStep completely?

owaiskazi19 (commented):
Overall looks with few comments

src/main/java/org/opensearch/flowframework/workflow/CreateIndexStep.java
dbwiddis: I (tentatively) disagree with this suggestion, at least for now.  @owaiskazi19 can you tell me what the difference here is?

owaiskazi19: Something we used to do while working on core. A better way to write the parameterized message. Nothing specific.

dbwiddis: Got it. I think that may have been necessary with earlier log4j syntax but it's not any more. 

I agree the concatenation should be replaced, but I think `logger.error("Failed to update index {}", indexName, exception);` should produce exactly the same thing.

src/main/resources/mappings/global-context.json
amitgalitz: I think we should remove all fields that are specific to a use case from the GC mapping. We can't sustainably add more and more fields as we have more and more use cases

src/main/java/org/opensearch/flowframework/indices/GlobalContextHandler.java
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    /**
+     * add document insert into global context index
+     * @param template the use-case template
+     * @param listener action listener
+     */
+    public void putTemplateToGlobalContext(Template template, ActionListener<IndexResponse> listener) {
```
joshpalis, line 74: Can we additionally have a method that allows for updating an entry given a document Id (`workflow_id`). This is for use in the `CreateWorkflow` API, which allows for updating a stored use case template. I envision this as a complete replacement of the use case template in the GC entry, keeping the same document ID.

src/main/java/org/opensearch/flowframework/function/ThrowingSupplierWrapper.java
```diff
@@ -0,0 +1,43 @@
...
+import java.util.function.Supplier;
+
+/**
+ * Wrapper for throwing checked exception inside places that does not allow to do so
+ */
+public class ThrowingSupplierWrapper {
+    /*
+     * Private constructor to avoid Jacoco complaining about public constructor
+     * not covered: https://tinyurl.com/yetc7tra
+     */
```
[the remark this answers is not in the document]
jackiehanyang: Sounds good, removed unnecessary java doc here

src/main/java/org/opensearch/flowframework/model/Template.java
```diff
@@ -49,6 +49,10 @@ public class Template implements ToXContentObject {
     public static final String USER_INPUTS_FIELD = "user_inputs";
     /** The template field name for template workflows */
     public static final String WORKFLOWS_FIELD = "workflows";
+    /** The template field name for template responses */
+    public static final String RESPONSES_FIELD = "responses";
```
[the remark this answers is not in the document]
jackiehanyang: Updated the name from `responses` to `user_outputs` to match with the `user_inputs` above

src/main/java/org/opensearch/flowframework/model/Template.java
```diff
@@ -132,6 +144,18 @@ public XContentBuilder toXContent(XContentBuilder builder, Params params) throws
         }
         xContentBuilder.endObject();
 
+        xContentBuilder.startObject(RESPONSES_FIELD);
+        for (Entry<String, Object> e : responses.entrySet()) {
+            xContentBuilder.field(e.getKey(), e.getValue());
+        }
```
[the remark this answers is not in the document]
jackiehanyang: I think it's a good idea to use generic type when try to prevent unknown value during parsing stage. But the responses/user_outputs map here has the exact same data structure as user_inputs. So I'm just keeping it the same as what user_inputs field is using. I will document this in a global context followup issue future reference.  Will attache this pr to the issue later

src/main/java/org/opensearch/flowframework/workflow/CreateIndexStep.java
```diff
@@ -96,15 +117,124 @@ public String getName() {
...
+                                                    internalListener.onFailure(exception);
+                                                }));
+                                        } else {
+                                            internalListener.onFailure(new FlowFrameworkException("Failed to update index: " + indexName));
+                                        }
+                                    }, exception -> {
+                                        logger.error("Failed to update index " + indexName, exception);
+                                        internalListener.onFailure(exception);
+                                    })
+                                );
```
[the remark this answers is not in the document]
jackiehanyang: As discussed with Owais and Amit, we're going to keep this method like this for now. Refactor will be done later when we have more indices onboard, for example, state index

src/main/java/org/opensearch/flowframework/workflow/ProcessNode.java
jackiehanyang: rebased with latest change

src/main/resources/mappings/global-context.json
jackiehanyang: Updated `user_inputs` mapping like this to keep it align with the final iteration of use case template. Will mention this in the global context followup issue to see if we want to further update the mapping to make it a generic text field

```
    "user_inputs": {
      "type": "nested",
      "properties": {
        "index_name": {
          "type": "keyword"
        },
        "index_setting": {
          "type": "keyword"
        }
      }
    },
```

build.gradle
jackiehanyang: Removed this dependency earlier but it stayed in my local, didn't get a chance to push it until I was offcall :)

src/main/java/org/opensearch/flowframework/indices/GlobalContextHandler.java
jackiehanyang: I will mention this in the global context followup issue and add this in the next PR. Let's not iterate based on Issues instead of PRs :)

src/test/java/org/opensearch/flowframework/workflow/CreateIndexStepTests.java
jackiehanyang: Not sure what you mean exactly.. If we're write unit tests for class `CreateIndexStep`, shouldn't we instantiate the actual class, mock its dependencies, and test its behavior?

src/main/java/org/opensearch/flowframework/workflow/CreateIndexStep.java
```diff
@@ -96,15 +119,132 @@ public String getName() {
...
+     */
+    public void initIndexIfAbsent(FlowFrameworkIndex index, ActionListener<Boolean> listener) {
+        String indexName = index.getIndexName();
+        String mapping = index.getMapping();
+
+        try (ThreadContext.StoredContext threadContext = client.threadPool().getThreadContext().stashContext()) {
+            ActionListener<Boolean> internalListener = ActionListener.runBefore(listener, () -> threadContext.restore());
+            if (!clusterService.state().metadata().hasIndex(indexName)) {
+                ActionListener<CreateIndexResponse> actionListener = ActionListener.wrap(r -> {
+                    if (r.isAcknowledged()) {
```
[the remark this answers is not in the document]
jackiehanyang: annotation cannot be placed in some places as block-level annotations are not directly supported in Java. Will mentioned in the following up issue and get it taken care when we separate CreateIndex for system index and custom index

dbwiddis (approved):
LGTM! Thanks for addressing all the comments.

joshpalis (approved):
Thanks @jackiehanyang for addressing all the comments

PR #66, opened 2023-09-30
Add timeout for node execution

### LINKED ISSUE #42: [FEATURE] Make Thread Pool available to Process Node

### LINKED ISSUE #45: [FEATURE] Include timeout settings as part of the user template

dbwiddis:
### Description

Adds a timeout setting for each node's processing of the Workflow step's `execute()` method in the `ProcessNode`.

Some steps should complete quickly (e.g., create index) while others may take more time to complete (upload/deploy model).

Presently implementing that in the user template (See #45 where I proposed this and nobody objected), with a default to 10s if omitted.
 - It is not required to be entered in the template. It's an option.
 - This location is temporary, it can be moved anywhere else with a simple code change. See #67 for one proposal where to move it.
 - Open to any other ideas, but...
 - I want to keep functionality similar to existing REST APIs where users can add a "timeout" param.  We have multiple APIs and need to enable multiple timeouts that the user can change.
 - Selecting a timeout of 0 means no timeout.

I considered a global timeout, but decided against it as the "sum of node timeouts" effectively enforces that.  I also considered a timeout by workflow type, tied to the default timeouts for their implemented APIs.

Other changes needed along the way:
 - Switched from executor to thread pool to pass around from create components and used for scheduled execution
 - Got rid of test thread leaks by using thread pool properly
 - Added test for plugin class
 - Got rid of singleton step factory and  process sorter classes since they didn't play nice with tests (a singleton is still injected when we eventually get to transport actions)
 - Added missing javadocs creating warnings for JDK18+

### Issues Resolved

Fixes #42 properly
Fixes #45 

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,5 +1,5 @@
 codecov:
-  require_ci_to_pass: yes
+  require_ci_to_pass: true
 
 # ignore files in demo package
 ignore:
@@ -12,5 +12,8 @@ coverage:
   status:
     project:
       default:
-        target: 70%    # the required coverage value
-        threshold: 1%  # the leniency in hitting the target
+        target: auto
+        threshold: 2%  # project coverage can drop
+    patch:
+      default:
+        target: 70%    # required diff coverage value
--- a/src/main/java/org/opensearch/flowframework/workflow/ProcessNode.java
+++ b/src/main/java/org/opensearch/flowframework/workflow/ProcessNode.java
@@ -17,7 +17,6 @@ import org.opensearch.threadpool.ThreadPool;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeoutException;
 import java.util.stream.Collectors;
 
@@ -111,6 +110,14 @@ public class ProcessNode {
     }
 
     /**
+     * Returns the timeout value of this node in the workflow. A value of {@link TimeValue#ZERO} means no timeout.
+     * @return The node's timeout value.
+     */
+    public TimeValue nodeTimeout() {
+        return nodeTimeout;
+    }
+
+    /**
      * Execute this node in the sequence.
      * Initializes the node's {@link CompletableFuture} and completes it when the process completes.
      *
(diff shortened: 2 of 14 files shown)


.codecov.yml
```diff
@@ -12,5 +12,8 @@ coverage:
   status:
     project:
       default:
-        target: 70%    # the required coverage value
-        threshold: 1%  # the leniency in hitting the target
+        target: auto
```
owaiskazi19, line 15: Don't we want to keep it as 70?

src/main/java/org/opensearch/flowframework/workflow/ProcessNode.java
```diff
@@ -90,64 +102,73 @@ public CompletableFuture<WorkflowData> future() {
...
                 }
-            }
-            CompletableFuture<WorkflowData> stepFuture = this.workflowStep.execute(input);
-            try {
-                stepFuture.orTimeout(15, TimeUnit.SECONDS).join();
-                logger.info(">>> Finished {}.", this.id);
+
+                ScheduledCancellable delayExec = null;
+                if (this.nodeTimeout.compareTo(TimeValue.ZERO) > 0) {
+                    delayExec = threadPool.schedule(() -> {
```
owaiskazi19, line 149: For my understanding. Why we are not using `scheduleWithFixedDelay`here?

src/main/java/org/opensearch/flowframework/workflow/WorkflowProcessSorter.java
```diff
@@ -91,7 +74,12 @@ public List<ProcessNode> sortProcessNodes(Workflow workflow) {
                 .map(e -> idToNodeMap.get(e.source()))
                 .collect(Collectors.toList());
 
-            ProcessNode processNode = new ProcessNode(node.id(), step, data, predecessorNodes, executor);
+            TimeValue nodeTimeout = Setting.parseTimeValue(
+                (String) node.inputs().getOrDefault(NODE_TIMEOUT_FIELD, NODE_TIMEOUT_DEFAULT_VALUE),
+                TimeValue.ZERO,
+                String.join(".", node.id(), INPUTS_FIELD, NODE_TIMEOUT_FIELD)
+            );
```
owaiskazi19, line 81: Should we create a utility method for the above?

src/test/java/org/opensearch/flowframework/FlowFrameworkPluginTests.java
```diff
@@ -8,8 +8,40 @@
...
+
+    @Override
+    public void tearDown() throws Exception {
+        ThreadPool.terminate(threadPool, 500, TimeUnit.MILLISECONDS);
+        super.tearDown();
+    }
+
+    public void testPlugin() throws IOException {
+        try (FlowFrameworkPlugin ffp = new FlowFrameworkPlugin()) {
+            assertEquals(2, ffp.createComponents(client, null, threadPool, null, null, null, null, null, null, null, null).size());
```
owaiskazi19, line 44: Why the size is 2 here?

owaiskazi19 (commented):
Looks good overall

.codecov.yml
dbwiddis: > Don't we want to keep it as 70?

Current coverage after this PR is merged would be 91.97%.  70 is way too low a threshold for the overall project.  Changing it to `auto` uses current project coverage as a baseline, with the threshold an acceptable level of lowering it.

I have added 70 in the section below as a minimum requirement for "patch", which means 70% of new lines must be covered.

See https://docs.codecov.com/docs/commit-status#patch-status

src/main/java/org/opensearch/flowframework/workflow/ProcessNode.java
dbwiddis: `scheduleWithFixedDelay` schedules something to run at intervals.  It would be appropriate for a "try this API until it is successful" request.  `schedule` is a one-shot.   See https://github.com/opensearch-project/OpenSearch/blob/5aa18636706597351aacfa7fc51364cd54ba2c04/server/src/main/java/org/opensearch/threadpool/Scheduler.java#L106-L135

src/test/java/org/opensearch/flowframework/FlowFrameworkPluginTests.java
owaiskazi19: Ah! It's from the actual `createComponent`. You can test it something like:
```java
Collection<Object> components = plugin.createComponents(params here)
assertEquals(1, components.size());
assertEquals(threadPool, OpenSearchResources.INSTANCE.getThreadPool());

owaiskazi19 (approved):
I am aware that we are trying to map timeouts similar to REST handler timeouts for nodes but we should think about the right place for them rather taking it as an input from the user. For now, this change looks good.

joshpalis (approved):
Thanks Dan for the changes. Lgtm, though I'm curious if the sum of all the node timeouts should be the timeout value of the `provision` API. I think it could be an option, but perhaps it might be better to have the workflow execution run **after** the `workflow_id` response is sent back to the user.

src/main/java/org/opensearch/flowframework/workflow/WorkflowProcessSorter.java
dbwiddis: Updated with a util with more customized error message.

src/test/java/org/opensearch/flowframework/FlowFrameworkPluginTests.java
dbwiddis: > Ah! It's from the actual `createComponent`. You can test it something like:
> 
> ```java
> Collection<Object> components = plugin.createComponents(params here)
> assertEquals(2, components.size());
> ```

Isn't that exactly what I'm doing?

dbwiddis:
> I am aware that we are trying to map timeouts similar to REST handler timeouts for nodes but we should think about the right place for them rather taking it as an input from the user. For now, this change looks good.

I'll move them to the step JSON definition file described in #67

PR #84, opened 2023-10-11
Separated out tests in a different GHA workflow

owaiskazi19:
### Description
Separated out test, integTest and yamRestTest in a different GHA workflow

### Issues Resolved
Part of https://github.com/opensearch-project/opensearch-ai-flow-framework/issues/73

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -48,8 +48,3 @@ jobs:
     - name: Build
       run: |
           ./gradlew check -x test -x integTest -x yamlRestTest
-    - name: Upload Coverage Report
-      if: matrix.os == 'ubuntu-latest'
-      uses: codecov/codecov-action@v3
-      with:
-        file: ./build/reports/jacoco/test/jacocoTestReport.xml
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -1,4 +1,4 @@
-name: Build and Test Plugin
+name: Build Plugin
 
 on:
   push:
@@ -34,7 +34,7 @@ jobs:
           - windows-latest
         java:
           - 17
-    name: Build and Test Plugin Template
+    name: Build
     if: github.repository == 'opensearch-project/opensearch-ai-flow-framework'
     runs-on: ${{ matrix.os }}
 
@@ -45,9 +45,9 @@ jobs:
       with:
         java-version: ${{ matrix.java }}
         distribution: temurin
-    - name: Build and Run Tests
+    - name: Build
       run: |
-          ./gradlew check
+          ./gradlew check -x test -x integTest -x yamlRestTest
     - name: Upload Coverage Report
       if: matrix.os == 'ubuntu-latest'
       uses: codecov/codecov-action@v3
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -33,7 +33,9 @@ jobs:
           - macOS-latest
           - windows-latest
         java:
+          - 11
           - 17
+          - 20
     name: Build
     if: github.repository == 'opensearch-project/opensearch-ai-flow-framework'
     runs-on: ${{ matrix.os }}
@@ -47,4 +49,4 @@ jobs:
         distribution: temurin
     - name: Build
       run: |
-          ./gradlew check -x test -x integTest -x yamlRestTest
+          ./gradlew build -x test -x integTest -x yamlRestTest
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,7 +13,6 @@ jobs:
       matrix:
         os:
           - ubuntu-latest
-          - macOS-latest
           - windows-latest
         java:
           - 17
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,6 +13,7 @@ jobs:
       matrix:
         os:
           - ubuntu-latest
+          - macOS-latest
           - windows-latest
         java:
           - 17
@@ -21,10 +22,10 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - name: Set up JDK 17
+      - name: Set up JDK ${{ matrix.java }}
         uses: actions/setup-java@v3
         with:
-          java-version: 17
+          java-version: ${{ matrix.java }}
           distribution: temurin
 
       - name: Test
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,6 +11,10 @@ jobs:
   test:
     strategy:
       matrix:
+        test:
+          - test jacocoTestReport
+          - integTest
+          - yamlRestTest
         os:
           - ubuntu-latest
           - macOS-latest
@@ -29,19 +33,7 @@ jobs:
           distribution: temurin
 
       - name: Test
-        run: ./gradlew test
-
-      - name: YamlRestTest
-        run:
-          ./gradlew yamlRestTest
-
-      - name: integTest
-        run:
-          ./gradlew integTest
-
-      - name: Run Tests with Coverage
-        run:
-          ./gradlew test jacocoTestReport --info
+        run: ./gradlew ${{ matrix.test }}
 
       - name: Upload Coverage Report
         if: matrix.os == 'ubuntu-latest'
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -39,6 +39,10 @@ jobs:
         run:
           ./gradlew integTest
 
+      - name: Run Tests with Coverage
+        run:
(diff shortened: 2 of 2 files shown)


dbwiddis (dismissed):
LGTM.

See also [this link](https://github.com/opensearch-project/geospatial/blob/059b10da7b9cc6cc9035c7ce8dba1f93955851f4/build.gradle#L203-L207) for a potential fix for the flaky integ test.

.github/workflows/CI.yml
```diff
@@ -45,9 +45,9 @@ jobs:
...
         java-version: ${{ matrix.java }}
         distribution: temurin
-    - name: Build and Run Tests
+    - name: Build
       run: |
-          ./gradlew check
+          ./gradlew check -x test -x integTest -x yamlRestTest
     - name: Upload Coverage Report
       if: matrix.os == 'ubuntu-latest'
       uses: codecov/codecov-action@v3
```
dbwiddis, line 53: Actually... codecov may depend on the tests being run. Does this still work?

owaiskazi19: It didn't run on this PR. That's something to think about. Do you think I should try moving codecov to tests.yml?

dbwiddis: Maybe. I'm still not clear on the motivation to split everything out from "check" which did it all anyway.

dbwiddis: Also since each step in the tests.yml does its own checkout, I'm not sure if the coverage would be cumulative for all the steps.

dbwiddis (approved):

.github/workflows/CI.yml
owaiskazi19: Yaay! Codecov worked!

owaiskazi19: The motivation behind is to run the tests and build in parallel and reduce the CI time overall when our plugin will become more heavy.

dbwiddis: > run the tests and build in parallel

Hmm.  

Full gradle check gives a total time of 50.570s.  
Gradle check without the tests takes 9.819s.
Running the tests separately gives 42.721s, because you're still running them sequentially.

You could get them to run in parallel by putting them up under the highest level `jobs:` label:

| Task | Duration |
| --- | --- |
| :integTest | 14.761s |  
| :test | 14.051s |  
| :yamlRestTest | 10.518s

Note jacoco is only considering unit tests, so you could (for now) put jacoco in that job and run the other two jobs without jacoco.
```groovy
jacocoTestReport {
    dependsOn test
    reports {
        xml.required = true
    }
}
```
There are ways to get jacoco to combine others: https://docs.gradle.org/7.4-rc-1/userguide/test_report_aggregation_plugin.html but I'm not sure that's needed.

.github/workflows/test.yml
```diff
@@ -0,0 +1,50 @@
...
+
+      - name: Run Tests with Coverage
+        run:
+          ./gradlew test jacocoTestReport --info
+
+      - name: Upload Coverage Report
+        if: matrix.os == 'ubuntu-latest'
+        uses: codecov/codecov-action@v3
+        with:
+          file: ./build/reports/jacoco/test/jacocoTestReport.xml
```
dbwiddis, line 50:
```suggestion
  test:
    if: github.repository == 'opensearch-project/opensearch-ai-flow-framework'
    strategy:
      matrix:
        test:
          - test jacocoTestReport
          - integTest
          - yamlRestTest
        os:
          - ubuntu-latest
          - macOS-latest
          - windows-latest
        java:
          - 17
    runs-on: ${{ matrix.os }}
    steps:
      - uses: actions/checkout@v4
      - name: Set up JDK ${{ matrix.java }}
        uses: actions/setup-java@v3
        with:
          java-version: ${{ matrix.java }}
          distribution: temurin
      - name: Test
        run: ./gradlew ${{ matrix.test }}
      - name: Upload Coverage Report
        if: matrix.os == 'ubuntu-latest' && matrix.test == 'test jacocoTestReport'
        uses: codecov/codecov-action@v3
        with:
          file: ./build/reports/jacoco/test/jacocoTestReport.xml
```

owaiskazi19: You could have raised the PR :)

.github/workflows/CI.yml
```diff
@@ -45,11 +45,6 @@ jobs:
...
-    - name: Build and Run Tests
+    - name: Build
       run: |
-          ./gradlew check
-    - name: Upload Coverage Report
-      if: matrix.os == 'ubuntu-latest'
-      uses: codecov/codecov-action@v3
-      with:
-        file: ./build/reports/jacoco/test/jacocoTestReport.xml
+          ./gradlew check -x test -x integTest -x yamlRestTest
```
amitgalitz, line 50: just wondering, why not use gradlew build here?

owaiskazi19: We can. I don't have any strong opinion here.

.github/workflows/test.yml
```diff
@@ -0,0 +1,42 @@
...
+        test:
+          - test jacocoTestReport
+          - integTest
+          - yamlRestTest
+        os:
+          - ubuntu-latest
+          - macOS-latest
+          - windows-latest
+        java:
+          - 17
```
amitgalitz, line 23: I remember in AD we had to test on 11, 17, 20. Should we do that here too? I think default for 2.x is jdk 11 still but could be wrong

owaiskazi19: @dbwiddis WDYT? Since you worked on the CI initially.

amitgalitz: https://opensearch.org/docs/latest/install-and-configure/install-opensearch/index/#java-compatibility this is what we have documented and i think java 20 is going to be for 3.x

.github/workflows/test.yml
```diff
@@ -0,0 +1,42 @@
...
+    types: [opened, synchronize, reopened]
+
+jobs:
+  test:
+    strategy:
+      matrix:
+        test:
+          - test jacocoTestReport
+          - integTest
+          - yamlRestTest
```
amitgalitz, line 17: also just for understanding, what are yamlRestTests for?

.github/workflows/CI.yml
amitgalitz: I am good with either, `./gradlew build` means we also run assemble which runs the `:compileJava, :processResources` which `check` does also but this does it into a jar.

.github/workflows/test.yml
owaiskazi19: It's for testing the rest-api-spec [yaml](https://github.com/opensearch-project/OpenSearch/tree/ea1cc9d054b2d7cf0046438ecc57777e29112c2b/rest-api-spec/src/main/resources/rest-api-spec) files. We don't have right now for our plugin, but we would in the future.

.github/workflows/test.yml
dbwiddis: It's probably not the right way to go, just pointing out how to achieve what you're trying to do :)

.github/workflows/test.yml
owaiskazi19: Pushed the change.

.github/workflows/CI.yml
owaiskazi19: Changed it to build.

dbwiddis: I think we should step back and figure out all the dependencies and figure out what we are trying to optimize here.  With the current PR we're shifting the tests somewhere else (still in series so not really gaining much) but duplicating compile/build stuff in parallel.

My general preference would be for CI to keep all the check stuff and `test` and code coverage bits, and just spin off the integ tests and yaml tests into another job in parallel (or just put them in the CI yaml at the top level under "jobs").

Lots of different ways to do it, haven't really been convinced we have a problem that needs solving yet :)

.github/workflows/test.yml
dbwiddis: We should probably match whatever OpenSearch tests.

.github/workflows/CI.yml
owaiskazi19: I agree with you. I will go ahead and close the PR. We can always come back and open it back when the plugin becomes heavy.

.github/workflows/test.yml
dbwiddis: Here's an example test https://github.com/opensearch-project/OpenSearch/blob/1447d75e5e9ea042dd4bb645d71f25aea1fb42a8/modules/geo/src/yamlRestTest/resources/rest-api-spec/test/geo_shape/10_basic.yml ... it tests a search API with specific input and makes sure it gets the expected output.

owaiskazi19:
Closing this PR, as the plugin is quite lightweight now and doesn't take much time to process. This PR has some good discussion for the CI. We can always revisit this later later.

PR #90, opened 2023-10-12
Fixed testPluginInstalled test

owaiskazi19:
### Description
`org.opensearch.flowframework.FlowFrameworkPluginIT.testPluginInstalled` test is failing on Windows CI

### Issues Resolved
Fixes https://github.com/opensearch-project/opensearch-ai-flow-framework/issues/87

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -162,7 +162,24 @@ task integTest(type: RestIntegTestTask) {
 }
 tasks.named("check").configure { dependsOn(integTest) }
 
+def _numNodes = findProperty('numNodes') as Integer ?: 1
+
 integTest {
+
+    doFirst {
+        // Tell the test JVM if the cluster JVM is running under a debugger so that tests can
+        // use longer timeouts for requests.
+        def isDebuggingCluster = getDebug() || System.getProperty("test.debug") != null
+        systemProperty 'cluster.debug', isDebuggingCluster
+        // Set number of nodes system property to be used in tests
+        systemProperty 'cluster.number_of_nodes', "${_numNodes}"
+        // There seems to be an issue when running multi node run or integ tasks with unicast_hosts
+        // not being written, the waitForAllConditions ensures it's written
+        getClusters().forEach { cluster ->
+            cluster.waitForAllConditions()
+        }
+    }
+
     // The --debug-jvm command-line option makes the cluster debuggable; this makes the tests debuggable
     if (System.getProperty("test.debug") != null) {
         jvmArgs '-agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=*:5005'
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -162,18 +162,53 @@ task integTest(type: RestIntegTestTask) {
...
     }
 }
 
 testClusters.integTest {
-    testDistribution = "INTEG_TEST"
+    testDistribution = "ARCHIVE"
+    // Cluster shrink exception thrown if we try to set numberOfNodes to 1, so only apply if > 1
+    if (_numNodes > 1) numberOfNodes = _numNodes
+    // When running integration tests it doesn't forward the --debug-jvm to the cluster anymore
+    // i.e. we have to use a custom property to flag when we want to debug elasticsearch JVM
```
dbwiddis, line 194:
```suggestion
    // i.e. we have to use a custom property to flag when we want to debug OpenSearch JVM
```

build.gradle
```diff
@@ -162,18 +162,53 @@ task integTest(type: RestIntegTestTask) {
...
+
     // The --debug-jvm command-line option makes the cluster debuggable; this makes the tests debuggable
     if (System.getProperty("test.debug") != null) {
         jvmArgs '-agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=*:5005'
     }
 }
 
 testClusters.integTest {
-    testDistribution = "INTEG_TEST"
+    testDistribution = "ARCHIVE"
```
dbwiddis, line 190: For my own education, what does this do?

owaiskazi19: This denotes the archive(zip, tar) distributions for test clusters.

build.gradle
owaiskazi19: Oops

build.gradle
joshpalis: also curious, what's the difference between the two?

owaiskazi19: It's just the type of distribution. Other plugins are testing (zip, tar) because we create that in our min distribution. So I'm following the same. [Ref](https://github.com/opensearch-project/OpenSearch/blob/main/buildSrc/src/main/java/org/opensearch/gradle/testclusters/TestDistribution.java#L35)

build.gradle
```diff
@@ -162,18 +162,53 @@ task integTest(type: RestIntegTestTask) {
...
+        def debugPort = 5005
+        nodes.forEach { node ->
+            node.jvmArgs("-agentlib:jdwp=transport=dt_socket,server=n,suspend=y,address=*:${debugPort}")
+            debugPort += 1
+        }
+    }
 
     // This installs our plugin into the testClusters
     plugin(project.tasks.bundlePlugin.archiveFile)
+    nodes.each { node ->
```
joshpalis, line 206: Curious if this is necessary for our plugin, I recognized this from the AD [build.gradle](https://github.com/opensearch-project/anomaly-detection/blob/cfb5fa07e5ddb67e3755a897622a70bee4edcf5c/build.gradle#L371), and from the documentation it seems this was only necessary to re-order the plugin list so that Job Scheduler was installed first.

owaiskazi19: Removed it. If we needed, we can add later.

dbwiddis (approved):

amitgalitz (approved):

PR #94, opened 2023-10-14
Run integ tests in parallel with build

dbwiddis:
### Description

Inspired by discussion on #84 (Thanks @owaiskazi19)

1. ~Removes the repo conditional on spotless and javadoc checks, so you don't need to submit the PR to get alerted about those failures.~
2. Separates the integTest and yamlRestTest tasks into a separate job that runs in parallel with the remaining gradle checks.  Right now they both take about the same time to complete.  If Integ tests start growing longer we can split these two out further.
3. Adds more JDKs to the build matrix.  **Note: I think 3 JDKs on 3 OS's for both test and integ is probably too much, probably should reduce them.**

image

image

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -1,4 +1,4 @@
-name: Build and Test Plugin
+name: CI
 
 on:
   push:
@@ -9,14 +9,12 @@ on:
 
 jobs:
   spotless:
-    if: github.repository == 'opensearch-project/opensearch-ai-flow-framework'
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
       - name: Spotless Check
         run: ./gradlew spotlessCheck
   javadoc:
-    if: github.repository == 'opensearch-project/opensearch-ai-flow-framework'
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
@@ -25,19 +23,14 @@ jobs:
       - name: Javadoc Check
         run: ./gradlew javadoc
   build:
+    if: github.repository == 'opensearch-project/opensearch-ai-flow-framework'
     needs: [spotless, javadoc]
     strategy:
       matrix:
-        os:
-          - ubuntu-latest
-          - macOS-latest
-          - windows-latest
-        java:
-          - 17
-    name: Build and Test Plugin Template
-    if: github.repository == 'opensearch-project/opensearch-ai-flow-framework'
+        os: [ubuntu-latest, macos-latest, windows-latest]
+        java: [11, 17, 20]
+    name: Test JDK${{ matrix.java }}, ${{ matrix.os }}
     runs-on: ${{ matrix.os }}
-
     steps:
     - uses: actions/checkout@v4
     - name: Set up JDK ${{ matrix.java }}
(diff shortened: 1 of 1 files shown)


joshpalis (approved):

.github/workflows/CI.yml
```diff
@@ -9,14 +9,12 @@ on:
...
 jobs:
   spotless:
-    if: github.repository == 'opensearch-project/opensearch-ai-flow-framework'
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
       - name: Spotless Check
         run: ./gradlew spotlessCheck
   javadoc:
-    if: github.repository == 'opensearch-project/opensearch-ai-flow-framework'
```
owaiskazi19, line 19: We should have the repo check or else it will run on the fork as before?

.github/workflows/CI.yml
```diff
@@ -47,9 +40,28 @@ jobs:
         distribution: temurin
     - name: Build and Run Tests
       run: |
-          ./gradlew check
+          ./gradlew check -x integTest -x yamlRestTest
```
owaiskazi19, line 43: Based on our discussion on #84, would code coverage just run for unit tests now?

.github/workflows/CI.yml
dbwiddis: > We should have the repo check or else it will run on the fork as before?

I specifically removed it as I think it's a good thing to run spotless and javadoc on anyone's branch before they submit a PR.

dbwiddis:
> To discuss the 3 JDKs mentioned in the PR description. The below JDK version are needed for the respective branches/development purpose.
> 
> ```
> 11 -> 2.x
> 20 -> 3.0
> 17 -> for our development
> ```
> 
> We can run 17 on just ubuntu though.

So this should be 17 (ubuntu only) and 20 on main, and 17 (ubuntu only) and 11 on backport?

.github/workflows/CI.yml
owaiskazi19: Looking at the number of emails we receive from all the repos. I would suggest to have this check please :)

dbwiddis: You don't like lots of emails?
Screenshot 2023-10-16 at 10 30 39 AM

owaiskazi19:
> So this should be 17 (ubuntu only) and 20 on main, and 17 (ubuntu only) and 11 on backport?

If we are running 11 on just backport/2.x, that way we won't be able to catch any issues on main which are not supported on JDK11.

.github/workflows/CI.yml
dbwiddis: My thought was that this will only generate an email if it fails, and if it fails on your branch before the PR sending the submitter an email it's better than failing on the PR and sending all the maintainers an email.  But I'll go ahead and put it back.

owaiskazi19: How many labels more I need now? :(
Screenshot 2023-10-16 at 10 34 29 AM

dbwiddis:
image

owaiskazi19 (approved):
Thanks for this!

jackiehanyang (approved):

Issue #88, opened 2023-10-12
[META] End to end template API user experience


dbwiddis:
The APIs to save and provision a workflow as part of https://github.com/opensearch-project/opensearch-ai-flow-framework/issues/49 need to be fully integrated with existing and needed validation checks and exception handling.  The goal is to provide fast-failure validation information to a user giving them enough information to correct an invalid template, and also provide appropriate feedback to a user when a workflow fails, identifying enough detail to correct the issue.

The following steps need to be completed:

### P0 Tasks
 - [x] Update existing workflow parsing code to replace `IllegalArgumentException` with a [`FlowFrameworkException`](https://github.com/opensearch-project/opensearch-ai-flow-framework/blob/main/src/main/java/org/opensearch/flowframework/exception/FlowFrameworkException.java) including an appropriate `RestStatus`.
 - [x] Update existing workflow process execution code to handle the cause of exceptionally completed futures (`TimeoutException` or other failures) with a [`FlowFrameworkException`](https://github.com/opensearch-project/opensearch-ai-flow-framework/blob/main/src/main/java/org/opensearch/flowframework/exception/FlowFrameworkException.java) including an appropriate `RestStatus`.
 - [x] Update the `RestResponse` for the API method(s) to handle these exceptions, and build a response to the user containing an appropriately detailed error message.
 - [x] Add input/output data validation checks to the workflow parsing.  https://github.com/opensearch-project/opensearch-ai-flow-framework/issues/18 and https://github.com/opensearch-project/opensearch-ai-flow-framework/issues/67. These may involve updating the workflow step factory.
 - [x] #105
 - [x] https://github.com/opensearch-project/opensearch-ai-flow-framework/issues/108
 - [x] https://github.com/opensearch-project/opensearch-ai-flow-framework/issues/123
 - [x] https://github.com/opensearch-project/opensearch-ai-flow-framework/issues/131
 - [x] https://github.com/opensearch-project/opensearch-ai-flow-framework/issues/149
 - [x] Handle optional params for Register Model, Deploy Model and Create Connector similar to done in Model Group Step
 - [x] https://github.com/opensearch-project/opensearch-ai-flow-framework/issues/22
 - [x] https://github.com/opensearch-project/opensearch-ai-flow-framework/issues/158
 - [x] https://github.com/opensearch-project/opensearch-ai-flow-framework/issues/75
 - [x] Combine RegisterLocalModelStep and GetMLTaskStep (Covered by PR #220)
[...]

PR #100, opened 2023-10-16
Fixing ProvisionWorkflowTransportAction, handling case in which a document does not exist

joshpalis:
### Description
Fixing bug, handling case in which a GC entry does not exist prior to attempting to parse the use case template document from the response source

### Issues Resolved
Part of #88 

### DIFF
--- a/src/main/java/org/opensearch/flowframework/transport/ProvisionWorkflowTransportAction.java
+++ b/src/main/java/org/opensearch/flowframework/transport/ProvisionWorkflowTransportAction.java
@@ -141,33 +141,38 @@ public class ProvisionWorkflowTransportAction extends HandledTransportAction<Wor
      * @param workflowListener The listener that updates the status of a workflow execution
      */
     private void executeWorkflow(Workflow workflow, ActionListener<String> workflowListener) {
-
-        List<ProcessNode> processSequence = workflowProcessSorter.sortProcessNodes(workflow);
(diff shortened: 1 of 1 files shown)


dbwiddis (approved):

src/main/java/org/opensearch/flowframework/transport/ProvisionWorkflowTransportAction.java
```diff
@@ -84,6 +84,16 @@ protected void doExecute(Task task, WorkflowRequest request, ActionListener<Work
...
                 context.restore();
 
+                if (!response.isExists()) {
+                    listener.onFailure(
+                        new FlowFrameworkException(
+                            "Failed to retrieve template (" + workflowId + ") from global context.",
+                            RestStatus.NOT_FOUND
+                        )
+                    );
+                    return;
```
owaiskazi19, line 94: Nit: Can you add test for this case?

owaiskazi19 (approved):
Approved with one comment

src/main/java/org/opensearch/flowframework/transport/ProvisionWorkflowTransportAction.java
joshpalis: The provision workflow transport action will be covered in our integration tests (Part of https://github.com/opensearch-project/opensearch-ai-flow-framework/issues/88)

PR #106, opened 2023-10-19
Updates exception handling with FlowFrameworkException and fix Global Context mapping

joshpalis:
### Description
This PR updates the workflow parsing, workflow execution and rest action exception handling to throw `FlowFrameworkException`s with the correct REST Status code. 

Example invalid create workflow request with missing name :
```
curl -i -XPOST "localhost:9200/_plugins/_flow_framework/workflow?pretty" -H "Content-Type:application/json" --data '{"description":"My semantic search use case","use_case":"SEMANTIC_SEARCH","operations":["PROVISION"],"version":{"template":"1.0.0","compatibility":["2.9.0","3.0.0"]},"user_inputs":{"index_name":"my-knn-index","index_settings":{}},"workflows":{"provision":{"nodes":[{"id":"create_index","type":"create_index","inputs":{"type":"knn","index-name":"my-knn-index","settings":""}},{"id":"create_ingest_pipeline","type":"create_ingest_pipeline","inputs":{"id":"my-ingest-pipeline","description":"some description","processors":[{"type":"text_embedding","params":{"model_id":"my-existing-model-id","input_field_name":"text_passage","output_field_name":"text_embedding"}}]}}],"edges":[{"source":"create_index","dest":"create_ingest_pipeline"}]}}}'

HTTP/1.1 400 Bad Request
content-length: 86

{
  "error" : "Request failed with exception: [A template object requires a name.]"
}
```

Note : Opening this PR up for second opinions regarding the various rest status codes for each exception. I've opted to return 400 for parsing, rest request format errors. Client exception rest status codes are determined via the OpenSearch [ExceptionsHelper::status](https://github.com/opensearch-project/OpenSearch/blob/1e28738b8c966011bf1ae1f00431f0377761cb0a/libs/core/src/main/java/org/opensearch/ExceptionsHelper.java#L95) method

Additionally this PR modified the global context mapping so that `user_outputs`, `workflows`, and `resourcesCreated` are of type `object` instead of `text`. This allows us to use `Template::toXContent` and `Template::parse` directly when reading and writing to the global context index

### Issues Resolved
Part of https://github.com/opensearch-project/opensearch-ai-flow-framework/issues/88


dbwiddis (approved):
LGTM.

I'm okay with 400 (Bad Request) but technically it means bad syntax. From [RFC 2068](https://www.rfc-editor.org/rfc/rfc2068):
>    The request could not be understood by the server due to malformed
>    syntax. The client SHOULD NOT repeat the request without
>    modifications.

Consider using 422 (Unprocessable Entity) for the cases where it's valid JSON but a bad value (equivalent of IllegalArgumetnException).  From [RFC 2518](https://www.rfc-editor.org/rfc/rfc2518):

>   The 422 (Unprocessable Entity) status code means the server
>    understands the content type of the request entity (hence a
>    415(Unsupported Media Type) status code is inappropriate), and the
>    syntax of the request entity is correct (thus a 400 (Bad Request)
>    status code is inappropriate) but was unable to process the contained
>    instructions.  For example, this error condition may occur if an XML
>    request body contains well-formed (i.e., syntactically correct), but
>    semantically erroneous, XML instructions.

For this PR I'd suggest 422 for the cases where you parse a valid edge source/destination but it doesn't match a node, for example (but based on `ExceptionsHelper` I'm still OK with 400.)

joshpalis:
Closing this PR in favor of https://github.com/opensearch-project/opensearch-ai-flow-framework/pull/137

PR #118, opened 2023-10-26
Added register a group model step

owaiskazi19:
### Description
Added Registering a group model

### Issues Resolved
Fixes https://github.com/opensearch-project/opensearch-ai-flow-framework/issues/83

### DIFF
--- a/src/main/java/org/opensearch/flowframework/model/Template.java
+++ b/src/main/java/org/opensearch/flowframework/model/Template.java
@@ -25,26 +25,18 @@ import java.util.Map;
 import java.util.Map.Entry;
 
 import static org.opensearch.core.xcontent.XContentParserUtils.ensureExpectedToken;
+import static org.opensearch.flowframework.common.CommonValue.COMPATIBILITY_FIELD;
+import static org.opensearch.flowframework.common.CommonValue.DESCRIPTION_FIELD;
+import static org.opensearch.flowframework.common.CommonValue.NAME_FIELD;
+import static org.opensearch.flowframework.common.CommonValue.TEMPLATE_FIELD;
+import static org.opensearch.flowframework.common.CommonValue.USE_CASE_FIELD;
+import static org.opensearch.flowframework.common.CommonValue.VERSION_FIELD;
+import static org.opensearch.flowframework.common.CommonValue.WORKFLOWS_FIELD;
 
 /**
  * The Template is the central data structure which configures workflows. This object is used to parse JSON communicated via REST API.
  */
 public class Template implements ToXContentObject {
-
-    /** The template field name for template name */
-    private static final String NAME_FIELD = "name";
-    /** The template field name for template description */
-    private static final String DESCRIPTION_FIELD = "description";
-    /** The template field name for template use case */
-    private static final String USE_CASE_FIELD = "use_case";
-    /** The template field name for template version information */
-    private static final String VERSION_FIELD = "version";
-    /** The template field name for template version */
-    private static final String TEMPLATE_FIELD = "template";
-    /** The template field name for template compatibility with OpenSearch versions */
-    private static final String COMPATIBILITY_FIELD = "compatibility";
-    /** The template field name for template workflows */
-    private static final String WORKFLOWS_FIELD = "workflows";
 
     private final String name;
     private final String description;
--- a/src/main/java/org/opensearch/flowframework/model/Template.java
+++ b/src/main/java/org/opensearch/flowframework/model/Template.java
@@ -32,19 +32,19 @@ import static org.opensearch.core.xcontent.XContentParserUtils.ensureExpectedTok
 public class Template implements ToXContentObject {
 
     /** The template field name for template name */
-    public static final String NAME_FIELD = "name";
+    private static final String NAME_FIELD = "name";
     /** The template field name for template description */
-    public static final String DESCRIPTION_FIELD = "description";
+    private static final String DESCRIPTION_FIELD = "description";
     /** The template field name for template use case */
-    public static final String USE_CASE_FIELD = "use_case";
+    private static final String USE_CASE_FIELD = "use_case";
     /** The template field name for template version information */
-    public static final String VERSION_FIELD = "version";
+    private static final String VERSION_FIELD = "version";
     /** The template field name for template version */
-    public static final String TEMPLATE_FIELD = "template";
+    private static final String TEMPLATE_FIELD = "template";
     /** The template field name for template compatibility with OpenSearch versions */
-    public static final String COMPATIBILITY_FIELD = "compatibility";
+    private static final String COMPATIBILITY_FIELD = "compatibility";
     /** The template field name for template workflows */
-    public static final String WORKFLOWS_FIELD = "workflows";
+    private static final String WORKFLOWS_FIELD = "workflows";
 
     private final String name;
     private final String description;
--- a/src/main/java/org/opensearch/flowframework/workflow/ModelGroupStep.java
+++ b/src/main/java/org/opensearch/flowframework/workflow/ModelGroupStep.java
@@ -111,25 +111,25 @@ public class ModelGroupStep implements WorkflowStep {
             registerModelGroupFuture.completeExceptionally(
                 new FlowFrameworkException("Model group name is not provided", RestStatus.BAD_REQUEST)
             );
-        }
+        } else {
+            MLRegisterModelGroupInputBuilder builder = MLRegisterModelGroupInput.builder();
+            builder.name(modelGroupName);
+            if (description != null) {
+                builder.description(description);
+            }
+            if (backendRoles != null && backendRoles.size() > 0) {
+                builder.backendRoles(backendRoles);
+            }
+            if (modelAccessMode != null) {
+                builder.modelAccessMode(modelAccessMode);
+            }
+            if (isAddAllBackendRoles != null) {
+                builder.isAddAllBackendRoles(isAddAllBackendRoles);
+            }
+            MLRegisterModelGroupInput mlInput = builder.build();
 
-        MLRegisterModelGroupInputBuilder builder = MLRegisterModelGroupInput.builder();
-        builder.name(modelGroupName);
-        if (description != null) {
-            builder.description(description);
+            mlClient.registerModelGroup(mlInput, actionListener);
         }
-        if (backendRoles != null && backendRoles.size() > 0) {
-            builder.backendRoles(backendRoles);
-        }
-        if (modelAccessMode != null) {
-            builder.modelAccessMode(modelAccessMode);
-        }
-        if (isAddAllBackendRoles != null) {
-            builder.isAddAllBackendRoles(isAddAllBackendRoles);
-        }
-        MLRegisterModelGroupInput mlInput = builder.build();
-
-        mlClient.registerModelGroup(mlInput, actionListener);
 
         return registerModelGroupFuture;
     }
--- a/src/main/java/org/opensearch/flowframework/workflow/ModelGroupStep.java
+++ b/src/main/java/org/opensearch/flowframework/workflow/ModelGroupStep.java
@@ -29,7 +29,7 @@ import java.util.concurrent.CompletableFuture;
 
 import static org.opensearch.flowframework.common.CommonValue.ADD_ALL_BACKEND_ROLES;
 import static org.opensearch.flowframework.common.CommonValue.BACKEND_ROLES_FIELD;
-import static org.opensearch.flowframework.common.CommonValue.DESCRIPTION;
+import static org.opensearch.flowframework.common.CommonValue.DESCRIPTION_FIELD;
 import static org.opensearch.flowframework.common.CommonValue.MODEL_ACCESS_MODE;
 import static org.opensearch.flowframework.common.CommonValue.NAME_FIELD;
 
@@ -92,8 +92,8 @@ public class ModelGroupStep implements WorkflowStep {
                     case NAME_FIELD:
                         modelGroupName = (String) content.get(NAME_FIELD);
                         break;
-                    case DESCRIPTION:
-                        description = (String) content.get(DESCRIPTION);
+                    case DESCRIPTION_FIELD:
+                        description = (String) content.get(DESCRIPTION_FIELD);
                         break;
                     case BACKEND_ROLES_FIELD:
                         backendRoles = getBackendRoles(content);
--- a/src/main/java/org/opensearch/flowframework/workflow/ModelGroupStep.java
+++ b/src/main/java/org/opensearch/flowframework/workflow/ModelGroupStep.java
@@ -82,7 +82,7 @@ public class ModelGroupStep implements WorkflowStep {
         String description = null;
         List<String> backendRoles = new ArrayList<>();
         AccessMode modelAccessMode = null;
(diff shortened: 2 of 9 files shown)


src/main/java/org/opensearch/flowframework/workflow/ModelGroupStep.java
```diff
@@ -0,0 +1,113 @@
...
+                logger.error("Failed to register model group");
+                registerModelGroupFuture.completeExceptionally(new FlowFrameworkException(e.getMessage(), ExceptionsHelper.status(e)));
+            }
+        };
+
+        String modelGroupName = null;
+        String description = null;
+
+        for (WorkflowData workflowData : data) {
+            Map<String, Object> content = workflowData.getContent();
```
amitgalitz, line 80: I think model group also takes an access_mode

src/main/java/org/opensearch/flowframework/workflow/ModelGroupStep.java
```diff
@@ -0,0 +1,113 @@
...
+            }
+        }
+
+        if (Stream.of(modelGroupName, description).allMatch(x -> x != null)) {
+            MLRegisterModelGroupInput mlInput = MLRegisterModelGroupInput.builder().name(modelGroupName).description(description).build();
+
+            mlClient.registerModelGroup(mlInput, actionListener);
+        } else {
+            registerModelGroupFuture.completeExceptionally(
+                new FlowFrameworkException("Required fields are not provided", RestStatus.BAD_REQUEST)
```
amitgalitz, line 102: nit: can be done later but should we move common error messages like these to a common file so we have consistent grammar, punctuation and easier for testing if we change anything.

src/main/java/org/opensearch/flowframework/workflow/ModelGroupStep.java
owaiskazi19: Yes, and those are optional. From the documentation [here](https://opensearch.org/docs/latest/ml-commons-plugin/extensibility/index/#step-1-register-a-model-group), I see just name and descriptions as mandatory fields.

src/main/java/org/opensearch/flowframework/workflow/ModelGroupStep.java
```diff
@@ -0,0 +1,113 @@
...
+
+        ActionListener<MLRegisterModelGroupResponse> actionListener = new ActionListener<>() {
+            @Override
+            public void onResponse(MLRegisterModelGroupResponse mlRegisterModelGroupResponse) {
+                logger.info("Model group registration successful");
+                registerModelGroupFuture.complete(
+                    new WorkflowData(
+                        Map.ofEntries(
+                            Map.entry("model_group_id", mlRegisterModelGroupResponse.getModelGroupId()),
+                            Map.entry("model_group_status", mlRegisterModelGroupResponse.getStatus())
```
amitgalitz, line 63: Is that status here referring only basically to response we get back 200/400 for example?

src/main/java/org/opensearch/flowframework/workflow/ModelGroupStep.java
amitgalitz: Description is also optional based on [this](https://opensearch.org/docs/latest/ml-commons-plugin/model-access-control/#request-fields). 

Also there are cases it seems like that if `access_mode` = `restricted` then we have to have backend_roles or add_all_backend_roles. I assume for this case we will just let handle ml-commons handle exceptions right?

src/main/java/org/opensearch/flowframework/workflow/ModelGroupStep.java
owaiskazi19: No this is the status of register model group. See [this](https://github.com/opensearch-project/ml-commons/blob/main/common/src/main/java/org/opensearch/ml/common/MLTaskState.java#L25).

src/main/java/org/opensearch/flowframework/workflow/ModelGroupStep.java
owaiskazi19: Handled the optional params using builder [here](https://github.com/opensearch-project/opensearch-ai-flow-framework/pull/118/files#diff-d1418e86d6fb899a62733f59aea086fb12c25a2fb08c9d21e420af35652f4b40R118). Thanks for the suggestion here.

src/main/java/org/opensearch/flowframework/workflow/ModelGroupStep.java
owaiskazi19: Made the exception error specific to this step.

joshpalis (approved):

src/main/java/org/opensearch/flowframework/model/Template.java
```diff
@@ -32,19 +32,19 @@
 public class Template implements ToXContentObject {
 
     /** The template field name for template name */
-    public static final String NAME_FIELD = "name";
+    private static final String NAME_FIELD = "name";
```
dbwiddis, line 35: Why are these all being changed to `private`?
 - we should be able to use them in testing
 - we should be able to use them in Workflow Steps to create them

src/main/java/org/opensearch/flowframework/workflow/ModelGroupStep.java
amitgalitz: I don't think register model group has a status though. its a synchronous api

src/main/java/org/opensearch/flowframework/workflow/ModelGroupStep.java
```diff
@@ -0,0 +1,141 @@
...
+            registerModelGroupFuture.completeExceptionally(
+                new FlowFrameworkException("Model group name is not provided", RestStatus.BAD_REQUEST)
+            );
+        } else {
+            MLRegisterModelGroupInputBuilder builder = MLRegisterModelGroupInput.builder();
+            builder.name(modelGroupName);
+            if (description != null) {
+                builder.description(description);
+            }
+            if (backendRoles != null && backendRoles.size() > 0) {
```
dbwiddis, line 120: You're defining this as an empty list to start with, so it should never be null. And `isEmpty()` is a few nanoseconds faster than getting the size.
```suggestion
            if (!backendRoles.isEmpty()) {
```

src/main/java/org/opensearch/flowframework/workflow/ModelGroupStep.java
```diff
@@ -0,0 +1,141 @@
...
+                logger.error("Failed to register model group");
+                registerModelGroupFuture.completeExceptionally(new FlowFrameworkException(e.getMessage(), ExceptionsHelper.status(e)));
+            }
+        };
+
+        String modelGroupName = null;
+        String description = null;
+        List<String> backendRoles = new ArrayList<>();
+        AccessMode modelAccessMode = null;
+        Boolean isAddAllBackendRoles = false;
```
dbwiddis, line 85: Why use a boxed `Boolean` here that you pre-initialize?  I'd expect to see it null (unset) so you can tell whether it was set false intentionally.

src/main/java/org/opensearch/flowframework/workflow/ModelGroupStep.java
```diff
@@ -0,0 +1,141 @@
...
+            if (description != null) {
+                builder.description(description);
+            }
+            if (backendRoles != null && backendRoles.size() > 0) {
+                builder.backendRoles(backendRoles);
+            }
+            if (modelAccessMode != null) {
+                builder.modelAccessMode(modelAccessMode);
+            }
+            if (isAddAllBackendRoles != null) {
```
dbwiddis, line 126: This will never be null with your initialization, but it probably should be initialized null.

src/test/java/org/opensearch/flowframework/workflow/CreateConnectorStepTests.java
```diff
@@ -49,6 +47,10 @@ public void setUp() throws Exception {
 
         MockitoAnnotations.openMocks(this);
 
+        ConnectorAction.ActionType actionType = ConnectorAction.ActionType.PREDICT;
+        String method = "post";
+        String url = "https://test.com";
```
dbwiddis, line 52: (Style/random) This is a real URL.

You may want to use `example.com` or `foo.test`. See https://www.rfc-editor.org/rfc/rfc2606.html :)

src/main/java/org/opensearch/flowframework/workflow/ModelGroupStep.java
```diff
@@ -0,0 +1,141 @@
...
+            for (Entry<String, Object> entry : content.entrySet()) {
+                switch (entry.getKey()) {
+                    case NAME_FIELD:
+                        modelGroupName = (String) content.get(NAME_FIELD);
+                        break;
+                    case DESCRIPTION:
+                        description = (String) content.get(DESCRIPTION);
+                        break;
+                    case BACKEND_ROLES_FIELD:
+                        backendRoles = (List<String>) content.get(BACKEND_ROLES_FIELD);
```
dbwiddis, line 99: This results in an unchecked cast warning.  

Can we refactor into one of:
- a temporary variable that we can suppress the warning on, or
- a helper method that always returns the correct type and does any required type checking/exception-throwing internally?

src/test/java/org/opensearch/flowframework/workflow/ModelGroupStepTests.java
```diff
@@ -0,0 +1,119 @@
...
+
+    }
+
+    public void testRegisterModelGroup() throws ExecutionException, InterruptedException, IOException {
+        String modelGroupId = "model_group_id";
+        String status = MLTaskState.CREATED.name();
+
+        ModelGroupStep modelGroupStep = new ModelGroupStep(machineLearningNodeClient);
+
+        ArgumentCaptor<ActionListener<MLRegisterModelGroupResponse>> actionListenerCaptor = ArgumentCaptor.forClass(ActionListener.class);
```
dbwiddis, line 67:
```suggestion
        @SuppressWarnings("unchecked")
        ArgumentCaptor<ActionListener<MLRegisterModelGroupResponse>> actionListenerCaptor = ArgumentCaptor.forClass(ActionListener.class);
```

src/test/java/org/opensearch/flowframework/workflow/ModelGroupStepTests.java
```diff
@@ -0,0 +1,119 @@
...
+        assertTrue(future.isDone());
+        assertEquals(modelGroupId, future.get().getContent().get("model_group_id"));
+        assertEquals(status, future.get().getContent().get("model_group_status"));
+
+    }
+
+    public void testRegisterModelGroupFailure() throws ExecutionException, InterruptedException, IOException {
+        ModelGroupStep modelGroupStep = new ModelGroupStep(machineLearningNodeClient);
+
+        ArgumentCaptor<ActionListener<MLRegisterModelGroupResponse>> actionListenerCaptor = ArgumentCaptor.forClass(ActionListener.class);
```
dbwiddis, line 89:
```suggestion
        @SuppressWarnings("unchecked")
        ArgumentCaptor<ActionListener<MLRegisterModelGroupResponse>> actionListenerCaptor = ArgumentCaptor.forClass(ActionListener.class);
```

src/main/java/org/opensearch/flowframework/workflow/ModelGroupStep.java
owaiskazi19: Status is there. That's why able to fetch it from `mlRegisterModelGroupResponse.getStatus()`.

dbwiddis (commented):
Some questions, nits, and useless but interesting style point.

src/main/java/org/opensearch/flowframework/model/Template.java
owaiskazi19: For workflow step we are fetching it from [here](https://github.com/opensearch-project/opensearch-ai-flow-framework/blob/main/src/main/java/org/opensearch/flowframework/common/CommonValue.java). These are specific to Template.

dbwiddis: > For workflow step we are fetching it from [here](https://github.com/opensearch-project/opensearch-ai-flow-framework/blob/main/src/main/java/org/opensearch/flowframework/common/CommonValue.java).

Then delete it here and fetch it from there. :)

src/main/java/org/opensearch/flowframework/workflow/ModelGroupStep.java
owaiskazi19: Added a help method. Can you review if that's how you thought? Thanks

src/test/java/org/opensearch/flowframework/workflow/CreateConnectorStepTests.java
owaiskazi19: Ah! Great one.

src/main/java/org/opensearch/flowframework/model/Template.java
```diff
@@ -25,27 +25,19 @@
...
+import static org.opensearch.flowframework.common.CommonValue.USE_CASE_FIELD;
+import static org.opensearch.flowframework.common.CommonValue.VERSION_FIELD;
+import static org.opensearch.flowframework.common.CommonValue.WORKFLOWS_FIELD;
 
 /**
  * The Template is the central data structure which configures workflows. This object is used to parse JSON communicated via REST API.
  */
 public class Template implements ToXContentObject {
 
-    /** The template field name for template name */
```
amitgalitz, line 34: nit: since I am making some changes to these fields in my PR as well, we might have conflict depending what gets merged first here

src/main/java/org/opensearch/flowframework/workflow/ModelGroupStep.java
```diff
@@ -0,0 +1,149 @@
...
+            registerModelGroupFuture.completeExceptionally(
+                new FlowFrameworkException("Model group name is not provided", RestStatus.BAD_REQUEST)
+            );
+        } else {
+            MLRegisterModelGroupInputBuilder builder = MLRegisterModelGroupInput.builder();
+            builder.name(modelGroupName);
+            if (description != null) {
+                builder.description(description);
+            }
+            if (!backendRoles.isEmpty()) {
```
amitgalitz, line 123: backendRoles.isEmpty might return a NPE when backendRoles is null i think

owaiskazi19: Made the change after @dbwiddis comment [here](https://github.com/opensearch-project/opensearch-ai-flow-framework/pull/118#discussion_r1375091164). @dbwiddis should I change it back?

src/main/java/org/opensearch/flowframework/model/Template.java
owaiskazi19: We can try to get this one in before yours.

src/main/java/org/opensearch/flowframework/workflow/ModelGroupStep.java
dbwiddis: backendRoles is initialized as an empty list, it won't be null.  This is fine.

dbwiddis (approved):

dbwiddis:
LGTM to merge. @amitgalitz can you resolve any conversations you still have open?

amitgalitz (approved):

Issue #117, opened 2023-10-26
[DOCS] Use Case Template Format


joshpalis:
This issue will be used to track any modifications to the Use Case Template format and should be considered the source of truth

### Summary of recent use case template changes
* removed fields
    * `operations` - not currently used, may be added back with the Orchestrate API
    * `user_inputs` 
    * `user_outputs` - moved to the State index
    * `resources_created` - moved to the State index

### Issues with current template format
* Dependent workflow step inputs need place holders (eg : `$(workflow_step_1.connector_id)`) whose values need to be substituted after parsing at runtime during workflow execution 
* validation of the workflow graph becomes more difficult

### Proposed Changes
* separates workflow node inputs field into two separate fields : `previous_node_inputs` , `user_inputs`
    * `previous_node_inputs` are inputs that are received from another workflow step
        * Format :
        * `{ "<workflow step ID>": "<field_name>"}`
    * `user_inputs` are inputs that are provided directly from the user

Example Template Components : 

* 3 workflow steps : `create_connector`, `register_model`, `deploy_model`
* 2 edges :
    *  `create_connector` → `register_model`
    *  `register_model` → `deploy_model`

### Current Template Format (note the placeholders)
```
{
  "name": "deploy-register-model",
  "description": "test case",
  "use_case": "TEST_USE_CASE",
  "version": {
    "template": "1.0.0",
    "compatibility": [
      "2.12.0",
      "3.0.0"
    ]
  },
  "workflows": {
    "provision": {
      "nodes": [
        {
          "id": "workflow_step_1",
          "type": "create_connector",
          "inputs": {
            "name": "my-connector",
            "description": "Remote connector to GPT-3.5-turbo",
            "version": "1",
            "protocol": "http",
            "credential": "...."
          }
        },
        {
          "id": "workflow_step_1",
          "type": "register_model",
          "inputs": {
            "name": "openAI-gpt-3.5-turbo",
            "function_name": "remote",
            "description": "test model",
            "connector_id": "${workflow_step_1.connector_id}"
          }
        },
        {
          "id": "workflow_step_2",
          "type": "deploy_model",
          "inputs": {
            "model_id": "${workflow_step_1.model_id}"
          }
        }
      ],
      "edges": [
        {
          "source": "workflow_step_1",
          "dest": "workflow_step_2"
        },
        {
          "source": "workflow_step_2",
          "dest": "workflow_step_3"
        }
      ]
    }
  }
}
```

### Proposed Template Format 
[...]

PR #119, opened 2023-10-27
Modifies use case template format and adds graph validation when provisioning

joshpalis:
### Description
Hoping to merge in https://github.com/opensearch-project/opensearch-ai-flow-framework/pull/118 first prior to this PR. 

This PR achieves multiple things
- Modifies the use case template to separate workflow inputs into `previous_node_inputs` and `user_inputs` as described by #117 
- Introduces a `workflow-step.json` to define the expected inputs and outputs for each workflow step
- Adds graph validation based on the `workflow-step.json`. Basically for each `ProcessNode` in a topologically sorted graph, we retrieve all the expected outputs of its predecessor nodes, combine this with the given user inputs (`WorkflowData` content) for the current node, and compare this list with the expected inputs of the current node type. If validation fails, we return to the user a list of missing inputs
- Adds tests to ensure that each registered `WorkflowStep` in the `WorkflowStepFactory` has a corresponding `WorkflowStepValidator` in the `workflow-steps.json`. This is a guard-rail to ensure that each new step that is registered has their inputs/outputs defined for use in validation

Example create workflow request to register and deploy a model  (with missing `create_connector` step that provides the `connector_id` needed for the `register_model_step`)
```
curl -i -XPOST "localhost:9200/_plugins/_flow_framework/workflow" -H "Content-Type:application/json" --data '{"name":"deploy-register-model","description":"test case","use_case":"TEST_USE_CASE","version":{"template":"1.0.0","compatibility":["2.12.0","3.0.0"]},"workflows":{"provision":{"nodes":[{"id":"workflow_step_1","type":"register_model","user_inputs":{"name":"openAI-gpt-3.5-turbo","function_name":"remote","description":"test model"}},{"id":"workflow_step_2","type":"deploy_model","user_inputs":{}}],"edges":[{"source":"workflow_step_1","dest":"workflow_step_2"}]}}}'
HTTP/1.1 200 OK
content-length: 38

{"workflow_id":"ewChbosBllTp7O9ki-A5"}%     
```

Executing async provisioning for workflow `ewChbosBllTp7O9ki-A5`
```
curl -i -XPOST "localhost:9200/_plugins/_flow_framework/workflow/ewChbosBllTp7O9ki-A5/_provision" 
HTTP/1.1 200 OK
content-length: 38

{"workflow_id":"ewChbosBllTp7O9ki-A5"}%     
```

OpenSearch logs show that graph validation failed and the `connector_id` field was not provided by `user_inputs` or `predecessor_node_inputs`
```
[2023-10-27T00:56:02,613][INFO ][o.o.f.w.CreateIndexStep  ] [dev-dsk-jpalis-2c-27c8aa11.us-west-2.amazon.com] create index:.plugins-ai-global-context
[2023-10-27T00:56:20,580][ERROR][o.o.f.t.ProvisionWorkflowTransportAction] [dev-dsk-jpalis-2c-27c8aa11.us-west-2.amazon.com] Provisioning failed for workflow ewChbosBllTp7O9ki-A5 : org.opensearch.flowframework.exception.FlowFrameworkException: Invalid graph, missing the following required inputs : [connector_id]
```

### Issues Resolved
Part of #88 
https://github.com/opensearch-project/opensearch-ai-flow-framework/issues/18
https://github.com/opensearch-project/opensearch-ai-flow-framework/issues/67
https://github.com/opensearch-project/opensearch-ai-flow-framework/issues/117


src/main/java/org/opensearch/flowframework/workflow/WorkflowProcessSorter.java
```diff
@@ -74,17 +75,68 @@ public List<ProcessNode> sortProcessNodes(Workflow workflow) {
...
+            }
+
+            // Retrieve all the user input data from this node
+            List<String> currentNodeUserInputs = new ArrayList<String>(processNode.input().getContent().keySet());
+
+            // Combine both predecessor outputs and current node user inputs
+            List<String> allInputs = new ArrayList<>();
+            allInputs.addAll(predecessorOutputs);
+            allInputs.addAll(currentNodeUserInputs);
+
```
owaiskazi19, line 126: 123-125 can be written as:

```java
List<String> allInputs =  Stream.concat(predecessorOutputs.stream(), currentNodeUserInputs.stream())

File: src/main/java/org/opensearch/flowframework/workflow/WorkflowProcessSorter.java [hunk 2]
```diff
@@ -74,17 +75,68 @@ public List<ProcessNode> sortProcessNodes(Workflow workflow) {
...
+            // Retrieve list of required inputs from the current process node and compare
+            List<String> expectedInputs = validator.getWorkflowStepValidators().get(processNode.workflowStep().getName()).getInputs();
+            if (!allInputs.containsAll(expectedInputs)) {
+                expectedInputs.removeAll(allInputs);
+                throw new IllegalArgumentException("Invalid graph, missing the following required inputs : " + expectedInputs.toString());
```
owaiskazi19, line 131: We need to throw this error in the execute API with a 4xx.

src/test/java/org/opensearch/flowframework/workflow/WorkflowProcessSorterTests.java
```diff
@@ -199,4 +202,31 @@ public void testExceptions() throws IOException {
...
+        Workflow workflow = new Workflow(Map.of(), List.of(registerModel, deployModel), List.of(edge));
+
+        List<ProcessNode> sortedProcessNodes = workflowProcessSorter.sortProcessNodes(workflow);
+        IllegalArgumentException ex = expectThrows(
+            IllegalArgumentException.class,
+            () -> workflowProcessSorter.validateGraph(sortedProcessNodes)
+        );
+        assertEquals("Invalid graph, missing the following required inputs : [connector_id]", ex.getMessage());
+
+    }
```
owaiskazi19, line 231: This test the failure case, can we have another test for a successfull case?

owaiskazi19 (commented):
Looks good overall with minor comments. Thanks for adding the validation. This is a great addition.
Let's try to break the PR into multiple PRs next time for easy review.

src/main/java/org/opensearch/flowframework/workflow/WorkflowProcessSorter.java
joshpalis: This looks cleaner, thanks for the suggestion

src/main/java/org/opensearch/flowframework/workflow/WorkflowProcessSorter.java
joshpalis: Sure, that makes sense. Currently, graph validation happens **after** we return a response back to the user's `_provision` call, but it makes sense to have validation before we attempt to execute. Ill make this modification so that we can return 4xx on failure to validate

src/test/java/org/opensearch/flowframework/workflow/WorkflowProcessSorterTests.java
joshpalis: sure

src/main/java/org/opensearch/flowframework/workflow/WorkflowProcessSorter.java
amitgalitz: Do you mean that it will be part of the create API then?

joshpalis: There's an issue https://github.com/opensearch-project/opensearch-ai-flow-framework/issues/99 to add a `dry-run` param to the create API which would do validation on a template there. 

For the provision API, we've also doing validation prior to executing. The modification that I'll make right now is to do validation before we execute the workflow on a separate thread. That way we can return a 4xx response back to the user.

src/main/java/org/opensearch/flowframework/model/WorkflowStepValidator.java
```diff
@@ -0,0 +1,91 @@
...
+        }
+        return new WorkflowStepValidator(parsedInputs, parsedOutputs);
+    }
+
+    /**
+     * Get the required inputs
+     * @return the inputs
+     */
+    public List<String> getInputs() {
+        return inputs;
```
dbwiddis, line 81:
```suggestion
        return List.copyOf(inputs);
```

src/main/java/org/opensearch/flowframework/model/WorkflowStepValidator.java
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    /**
+     * Get the required outputs
+     * @return the outputs
+     */
+    public List<String> getOutputs() {
+        return outputs;
```
dbwiddis, line 89:
```suggestion
        return List.copyOf(outputs);
```

src/main/java/org/opensearch/flowframework/model/WorkflowValidator.java
```diff
@@ -0,0 +1,87 @@
...
+    public static WorkflowValidator parse(String file) throws IOException {
+
+        URL url = WorkflowValidator.class.getClassLoader().getResource(file);
+        String json = Resources.toString(url, Charsets.UTF_8);
+        XContentParser parser = JsonXContent.jsonXContent.createParser(
+            NamedXContentRegistry.EMPTY,
+            LoggingDeprecationHandler.INSTANCE,
+            json
+        );
+        ensureExpectedToken(XContentParser.Token.START_OBJECT, parser.nextToken(), parser);
```
dbwiddis, line 75: Use this instead:
https://github.com/opensearch-project/opensearch-ai-flow-framework/blob/4106cbae2904d210659d4a6646574a2e51f71984/src/main/java/org/opensearch/flowframework/util/ParseUtils.java#L45-L53

src/main/java/org/opensearch/flowframework/model/WorkflowValidator.java
```diff
@@ -0,0 +1,87 @@
...
+     * Parse a workflow step JSON file into a WorkflowValidator object
+     *
+     * @param file the file name of the workflow step json
+     * @return A {@link WorkflowValidator} represented by the JSON
+     * @throws IOException on failure to read and parse the json file
+     */
+    public static WorkflowValidator parse(String file) throws IOException {
+
+        URL url = WorkflowValidator.class.getClassLoader().getResource(file);
+        String json = Resources.toString(url, Charsets.UTF_8);
```
dbwiddis, line 69: This code is duplicated here.
https://github.com/opensearch-project/opensearch-ai-flow-framework/blob/4106cbae2904d210659d4a6646574a2e51f71984/src/main/java/org/opensearch/flowframework/indices/FlowFrameworkIndicesHandler.java#L250-L253

Both should be moved into a util method somewhere (maybe in `ParseUtils`?)

src/main/java/org/opensearch/flowframework/model/WorkflowValidator.java
```diff
@@ -0,0 +1,87 @@
...
+        ensureExpectedToken(XContentParser.Token.START_OBJECT, parser.nextToken(), parser);
+        return parse(parser);
+    }
+
+    /**
+     * Get the map of WorkflowStepValidators
+     * @return the map of WorkflowStepValidators
+     */
+    public Map<String, WorkflowStepValidator> getWorkflowStepValidators() {
+        return this.workflowStepValidators;
```
dbwiddis, line 84:
```suggestion
        return Map.copyOf(this.workflowStepValidators);
```

src/main/java/org/opensearch/flowframework/workflow/WorkflowStepFactory.java
```diff
@@ -45,25 +41,6 @@ private void populateMap(ClusterService clusterService, Client client, MachineLe
         stepMap.put(DeployModelStep.NAME, new DeployModelStep(mlClient));
         stepMap.put(CreateConnectorStep.NAME, new CreateConnectorStep(mlClient));
         stepMap.put(ModelGroupStep.NAME, new ModelGroupStep(mlClient));
-
-        // TODO: These are from the demo class as placeholders, remove when demos are deleted
```
dbwiddis, line 49: Thanks for removing this. :)

src/main/java/org/opensearch/flowframework/workflow/CreateIndexStep.java
```diff
@@ -74,7 +74,7 @@ public void onFailure(Exception e) {
 
         for (WorkflowData workflowData : data) {
             Map<String, Object> content = workflowData.getContent();
-            index = (String) content.get("index-name");
+            index = (String) content.get("index_name");
```
dbwiddis, line 77: Do we have this field name in a constant somewhere? Let's use it (to avoid typos like the one you're fixing!)

Not part of your PR but the `settings != null` conditional below is never going to shortcut this loop as it's not ever fetched.  Is that missing in this iteration?

src/main/java/org/opensearch/flowframework/workflow/CreateIngestPipelineStep.java
```diff
@@ -127,7 +127,7 @@ public CompletableFuture<WorkflowData> execute(List<WorkflowData> data) {
                 logger.info("Created ingest pipeline : " + putPipelineRequest.getId());
 
                 // PutPipelineRequest returns only an AcknowledgeResponse, returning pipelineId instead
-                createIngestPipelineFuture.complete(new WorkflowData(Map.of("pipelineId", putPipelineRequest.getId())));
+                createIngestPipelineFuture.complete(new WorkflowData(Map.of("pipeline_id", putPipelineRequest.getId())));
```
dbwiddis, line 130: Let's try to make nearly all of these strings constants to make everyone's life easier.

dbwiddis (commented):
Looks great, a few suggestions.

src/main/java/org/opensearch/flowframework/model/WorkflowValidator.java
joshpalis: Ah good call out

src/main/java/org/opensearch/flowframework/workflow/CreateIndexStep.java
joshpalis: Ill change these to constants and move them over to our Common class. As for the index settings, we still need to add support for that, but its not a priority at the moment

src/main/java/org/opensearch/flowframework/workflow/WorkflowProcessSorter.java
owaiskazi19: @joshpalis I thought a little about this. What do you think of having the validation on create API? Basically #99. Also, that means we will be doing validation on create API only when `dry-run` param is present but for provision API we will be doing it always.

joshpalis: Yes, your understanding is correct, we're aligned.

src/main/java/org/opensearch/flowframework/workflow/WorkflowStepFactory.java
```diff
@@ -57,4 +57,12 @@ public WorkflowStep createStep(String type) {
         }
         throw new FlowFrameworkException("Workflow step type [" + type + "] is not implemented.", RestStatus.NOT_IMPLEMENTED);
     }
+
+    /**
+     * Gets the step map
+     * @return the step map
+     */
+    public Map<String, WorkflowStep> getStepMap() {
+        return this.stepMap;
```
dbwiddis, line 66: Sorry I missed this in earlier review. If we want to make this public we need a `copyOf()` to prevent the caller from modifying the map.

dbwiddis (approved):
Sorry, one last change please (but I clicked approve so once that's fixed I don't need to re-review.)

owaiskazi19 (approved):

PR #139, opened 2023-11-01
Search APIs to fetch all the workflows present

owaiskazi19:
### Description
This PR adds 2 APIs:
1. GET _plugins/_flow_framework/workflow/_search
2. POST _plugins/_flow_framework/workflow/_search

The request and response are present in the issue.

### Issues Resolved
Closes https://github.com/opensearch-project/opensearch-ai-flow-framework/issues/124


src/main/java/org/opensearch/flowframework/rest/AbstractSearchWorkflowAction.java
```diff
@@ -0,0 +1,98 @@
...
+    @Override
+    public List<Route> routes() {
+        List<Route> routes = new ArrayList<>();
+        for (String path : urlPaths) {
+            routes.add(new Route(RestRequest.Method.POST, path));
+            routes.add(new Route(RestRequest.Method.GET, path));
+        }
+        return routes;
+    }
+}
```
owaiskazi19, line 98: The tests for this file will covered through integ tests

src/main/java/org/opensearch/flowframework/util/RestHandlerUtils.java
```diff
@@ -0,0 +1,44 @@
...
+        // 1. Move UI_METADATA to GC Index
+        // 2. check if the request came from dashboard and exclude UI_METADATA
+        if (searchSourceBuilder.fetchSource() != null) {
+            String[] newArray = (String[]) ArrayUtils.addAll(searchSourceBuilder.fetchSource().excludes(), USER_EXCLUDE);
+            return new FetchSourceContext(true, searchSourceBuilder.fetchSource().includes(), newArray);
+        } else {
+            return null;
+        }
+    }
+}
```
owaiskazi19, line 44: The tests for this file will covered through integ tests

src/main/java/org/opensearch/flowframework/rest/AbstractSearchWorkflowAction.java
```diff
@@ -0,0 +1,98 @@
...
+        return channel -> client.execute(actionType, searchRequest, search(channel));
+    }
+
+    /**
+     * Builds the action response for the Search Request
+     *
+     * @param channel the REST channel
+     * @return the action response
+     */
+    protected RestResponseListener<SearchResponse> search(RestChannel channel) {
```
joshpalis, line 77: Let's handle the `onFailure` case here and wrap any exception with a `FlowFrameworkException`. I suggest rebasing with PR https://github.com/opensearch-project/opensearch-ai-flow-framework/pull/137 after its merged in. It modifies the `FlowFrameworkException` to implement `ToXContentObject` so that exceptions can be outputted in the correct format as part of a rest response.

joshpalis (commented):
Initial pass, looks good so far. A small comment, and can you also include logs in the PR description after hitting this API?

owaiskazi19:
> Initial pass, looks good so far. A small comment, and can you also include logs in the PR description after hitting this API?

Posted the response in the issue. Do you want me to add in the PR as well?

src/main/java/org/opensearch/flowframework/rest/AbstractSearchWorkflowAction.java
owaiskazi19: Here the error is read from `SearchResponse` itself and not from the `exception`
```
 return new BytesRestResponse(RestStatus.REQUEST_TIMEOUT, response.toString());
 ```

**Review by joshpalis — APPROVED** [review]

**Comment by joshpalis** [comment]
@owaiskazi19  Perhaps lets merge in https://github.com/opensearch-project/opensearch-ai-flow-framework/pull/142/ first prior to this PR. After rebasing, you can also set this new API behind the feature flag

File: src/main/java/org/opensearch/flowframework/rest/AbstractSearchWorkflowAction.java [hunk 4]
```diff
@@ -0,0 +1,115 @@
...
+
+    @Override
+    protected RestChannelConsumer prepareRequest(RestRequest request, NodeClient client) throws IOException {
+        if (!flowFrameworkFeatureEnabledSetting.isFlowFrameworkEnabled()) {
+            FlowFrameworkException ffe = new FlowFrameworkException(
+                "This API is disabled. To enable it, update the setting [" + FLOW_FRAMEWORK_ENABLED.getKey() + "] to true.",
+                RestStatus.FORBIDDEN
+            );
+            return channel -> channel.sendResponse(
+                new BytesRestResponse(ffe.getRestStatus(), ffe.toXContent(channel.newErrorBuilder(), ToXContent.EMPTY_PARAMS))
```
amitgalitz, line 77: just wondering what is the toXContent adding here, I just see: `return builder.startObject().field("error", this.getMessage()).endObject();` does it just add error before message?

joshpalis: Yes, it just converts the exception to json xcontent so that we can write it back to the rest channel

src/main/java/org/opensearch/flowframework/rest/AbstractSearchWorkflowAction.java
```diff
@@ -0,0 +1,115 @@
...
+            );
+        }
+        SearchSourceBuilder searchSourceBuilder = new SearchSourceBuilder();
+        searchSourceBuilder.parseXContent(request.contentOrSourceParamParser());
+        searchSourceBuilder.fetchSource(getSourceContext(request, searchSourceBuilder));
+        searchSourceBuilder.seqNoAndPrimaryTerm(true).version(true);
```
amitgalitz, line 83: I don't think we currently update seqNo and Primary term but its maybe something we should add. 
They help with concurency control when updating the index

src/main/java/org/opensearch/flowframework/rest/AbstractSearchWorkflowAction.java
amitgalitz: Discussed more offline with Josh and its the current best way we have for converting the exception to JSON when we want to take control of what the status will be. Can't think of a better way though, it just seemed like adding more work to call on ToXContent each time since we didn't have that in AD. Understand why its done now a little better

src/main/java/org/opensearch/flowframework/transport/SearchWorkflowTransportAction.java
```diff
@@ -0,0 +1,48 @@
...
+     */
+    @Inject
+    public SearchWorkflowTransportAction(TransportService transportService, ActionFilters actionFilters, Client client) {
+        super(SearchWorkflowAction.NAME, transportService, actionFilters, SearchRequest::new);
+        this.client = client;
+    }
+
+    @Override
+    protected void doExecute(Task task, SearchRequest request, ActionListener<SearchResponse> actionListener) {
+        try (ThreadContext.StoredContext context = client.threadPool().getThreadContext().stashContext()) {
```
amitgalitz, line 42: Can you add some TODO or maybe we should reference an issue when we need to add control access here to make sure user only gets workflows that we have access too.

amitgalitz (approved):

src/main/java/org/opensearch/flowframework/rest/AbstractSearchWorkflowAction.java
owaiskazi19: I am guessing this is something that will be handled with AccessController.

Issue #136, opened 2023-10-31
[FEATURE] Resources created in state index and de-provisioning mapping


amitgalitz:
### Is your feature request related to a problem?
Right now we have a resources created field in the state index but there are a few ways on how we want to store this information.

### What solution would you like?
We can either map the ID of what was created to the type/name of the workflow step that created it or to the object name itself. 

Option 1 [Preferred]:

```
"resources_created": [
        {
            "type": "create_index_step",
            "id": "FobagIsBjjuFSw4Sfl3c"
        },
        {
            "type": "register_model",
            "id": "FobagIsBjjuFSw4Sfl3c"
        }
    ]
    
 ```
 Option 2:
 ```
"resources_created": [
        {
            "type": "index",
            "id": "FobagIsBjjuFSw4Sfl3c"
        },
        {
            "type": "model",
            "id": "FobagIsBjjuFSw4Sfl3c"
        }
    ]
    
 ```

I think it makes sense to map the created objected to the step that created it, in order to avoid adding another category of things like `model/index/agent`. 

However two things we need to consider are:
1. The order of the created objects in terms of de-provisioning order. Especially if some things depend on each other. For example do we need to delete the agent first before deleting the model that uses it. We can also potentially get this order from the use case template's nodes and edges themselves and delete in reverse order. 

2. How will we map each workflow step that created the resources to the workflow step that deletes the resources. We should probably have a way to match the create request to the equivalent delete request when we add workflow nodes/steps.

There could be a global delete map that looks something like this, this would be however another thing to maintain when adding new steps:
create_api:delete_api
```
 "deletion_map": {
            "create_index_step": "delete_index_step",
            "register_model": "delete_model",
        },
```

## Current Solution:
```
"resources_created": [
        {
            "workflow_step_name": "create_connector",
            "resource_id": "fZTmz4sBVvQ80zEwkC2z"
        }
    ]
    
 ```

### Additional Context

Right now we have some implementation of updating the resources created field in CreateConnectorStep. We should decide if we want to add more information here by fetching the resource name from some enum. 

Additionally we can abstract the method in CreateConnectorStep to be used by all steps easily as mentioned in this comment: https://github.com/opensearch-project/opensearch-ai-flow-framework/pull/148#discussion_r1393484371

joshpalis:
[...]

PR #148, opened 2023-11-06
Adding a get status API

amitgalitz:
### Description
Currently implemented so when not given _all=true the response is:
```
{
    {
    "workflow_id": "fJTmz4sBVvQ80zEwbS1x",
    "state": "COMPLETED",
    "resources_created": [
        {
            "resource_id": "fZTmz4sBVvQ80zEwkC2z"
        }
    ]
}
}
```
when: `/_plugins/_flow_framework/workflow/yfRsposBMbAFAhZEJ1qE/_status?all=true`
then response includes all of state index like this:
```
{
    "workflow_id": "fJTmz4sBVvQ80zEwbS1x",
    "state": "COMPLETED",
    "provisioning_progress": "DONE",
    "provision_start_time": 1700000075690,
    "provision_end_time": 1700000076398,
    "resources_created": [
        {
            "resource_id": "fZTmz4sBVvQ80zEwkC2z"
        }
    ]
}
```

Currently GC content (template) is not embedded within response, added some of the prerequisites for that but not fully implemented, can be done when GET workflow API is complete.

Additional Issues highlighting further work:https://github.com/opensearch-project/opensearch-ai-flow-framework/issues/136 -> This PR adds the resources created addition logic to create_connector, following PR will add the work for each step.

https://github.com/opensearch-project/opensearch-ai-flow-framework/issues/171 -> adding the full template in the status response as well to help with faster frontend fetching

### Issues Resolved
Closes https://github.com/opensearch-project/opensearch-ai-flow-framework/issues/22


src/main/java/org/opensearch/flowframework/rest/RestGetWorkflowAction.java
```diff
@@ -0,0 +1,107 @@
...
+import java.util.Locale;
+
+import static org.opensearch.flowframework.common.CommonValue.WORKFLOW_ID;
+import static org.opensearch.flowframework.common.CommonValue.WORKFLOW_URI;
+import static org.opensearch.flowframework.common.FlowFrameworkFeatureEnabledSetting.FLOW_FRAMEWORK_ENABLED;
+
+/**
+ * Rest Action to facilitate requests to get a workflow status
+ */
+public class RestGetWorkflowAction extends BaseRestHandler {
```
joshpalis, line 38: Perhaps lets change this rest action name to `RestGetWorkflowStateAction`, since its returning an object of type `WorkflowState` instead of `Workflow`

src/main/java/org/opensearch/flowframework/transport/GetWorkflowAction.java
```diff
@@ -0,0 +1,27 @@
...
+package org.opensearch.flowframework.transport;
+
+import org.opensearch.action.ActionType;
+
+import static org.opensearch.flowframework.common.CommonValue.TRANSPORT_ACTION_NAME_PREFIX;
+
+/**
+ * External Action for public facing RestGetWorkflowAction
+ */
+public class GetWorkflowAction extends ActionType<GetWorkflowResponse> {
```
joshpalis, line 18: Lets change this name too

src/main/java/org/opensearch/flowframework/transport/GetWorkflowResponse.java
```diff
@@ -0,0 +1,65 @@
...
+import org.opensearch.core.xcontent.ToXContentObject;
+import org.opensearch.core.xcontent.XContentBuilder;
+import org.opensearch.flowframework.model.WorkflowState;
+
+import java.io.IOException;
+
+/**
+ * Transport Response from getting a workflow status
+ */
+public class GetWorkflowResponse extends ActionResponse implements ToXContentObject {
```
joshpalis, line 23: And lets change this class name to `GetWorkfowStateResponse`

src/main/java/org/opensearch/flowframework/transport/GetWorkflowTransportAction.java
```diff
@@ -0,0 +1,99 @@
...
+ */
+public class GetWorkflowTransportAction extends HandledTransportAction<WorkflowRequest, GetWorkflowResponse> {
+
+    private final Logger logger = LogManager.getLogger(GetWorkflowTransportAction.class);
+
+    private final Client client;
+    private final NamedXContentRegistry xContentRegistry;
+
+    /**
+     * Intantiates a new CreateWorkflowTransportAction
```
joshpalis, line 47: Lets change this javadoc to match the class name

src/main/java/org/opensearch/flowframework/transport/GetWorkflowTransportAction.java
```diff
@@ -0,0 +1,99 @@
...
+                    try (XContentParser parser = ParseUtils.createXContentParserFromRegistry(xContentRegistry, r.getSourceAsBytesRef())) {
+                        ensureExpectedToken(XContentParser.Token.START_OBJECT, parser.nextToken(), parser);
+                        WorkflowState workflowState = WorkflowState.parse(parser);
+                        listener.onResponse(new GetWorkflowResponse(workflowState, request.getAll()));
+                    } catch (Exception e) {
+                        logger.error("Failed to parse workflowState" + r.getId(), e);
+                        listener.onFailure(e);
+                    }
+                } else {
+                    listener.onFailure(new FlowFrameworkException("Fail to find workflow", RestStatus.BAD_REQUEST));
```
joshpalis, line 84: I think a `NOT_FOUND` status code would apply here instead of `BAD_REQUEST`

src/main/java/org/opensearch/flowframework/transport/GetWorkflowTransportAction.java
```diff
@@ -0,0 +1,99 @@
...
+                }
+            }, e -> {
+                if (e instanceof IndexNotFoundException) {
+                    listener.onFailure(new FlowFrameworkException("Fail to find workflow", RestStatus.BAD_REQUEST));
```
joshpalis, line 88: Additionally here, `NOT_FOUND` would suffice

src/main/java/org/opensearch/flowframework/transport/ProvisionWorkflowTransportAction.java
```diff
@@ -201,7 +233,6 @@ private void executeWorkflow(List<ProcessNode> workflowSequence, ActionListener<
             // Attempt to join each workflow step future, may throw a CompletionException if any step completes exceptionally
             workflowFutureList.forEach(CompletableFuture::join);
 
-            // TODO : Create State Index request with provisioning state, start time, end time, etc, pending implementation. String for now
             workflowListener.onResponse("READY");
```
joshpalis, line 236: The `workflowListener` is somewhat of a placeholder that accepts a string on response. I see that the string isnt used when updating the  state index, perhaps lets just forgo the action listener all-together

src/main/java/org/opensearch/flowframework/transport/WorkflowRequest.java
```diff
@@ -43,19 +43,36 @@ public class WorkflowRequest extends ActionRequest {
...
+
     /**
      * Instantiates a new WorkflowRequest
      * @param workflowId the documentId of the workflow
      * @param template the use case template which describes the workflow
      * @param dryRun flag to indicate if validation is necessary
+     * @param all whether the get request is looking for all fields in status
      */
-    public WorkflowRequest(@Nullable String workflowId, @Nullable Template template, boolean dryRun) {
+    public WorkflowRequest(@Nullable String workflowId, @Nullable Template template, boolean dryRun, boolean all) {
```
joshpalis, line 61: This `WorkflowRequest` class was meant primarily for creating/provisioning. I think it may be beneficial to create a separate request class for use in the status rest action rather than adding more fields to this. We won't be using this `all` field for creating/provisioning a workflow

joshpalis (commented):
Initial pass, looks good so far, a couple of comments mostly on naming.

src/main/java/org/opensearch/flowframework/rest/RestGetWorkflowAction.java
```diff
@@ -0,0 +1,107 @@
...
+            }
+
+            boolean all = request.paramAsBoolean("_all", false);
+            WorkflowRequest workflowRequest = new WorkflowRequest(workflowId, null, false, all);
+            return channel -> client.execute(GetWorkflowAction.INSTANCE, workflowRequest, ActionListener.wrap(response -> {
+                XContentBuilder builder = response.toXContent(channel.newBuilder(), ToXContent.EMPTY_PARAMS);
+                channel.sendResponse(new BytesRestResponse(RestStatus.OK, builder));
+            }, exception -> {
+                try {
+                    FlowFrameworkException ex = (FlowFrameworkException) exception;
```
owaiskazi19, line 85: We should create a new object for `FlowFrameworkException` than casting it.

src/main/java/org/opensearch/flowframework/transport/CreateWorkflowTransportAction.java
```diff
@@ -160,9 +163,9 @@ protected void doExecute(Task task, WorkflowRequest request, ActionListener<Work
         }
     }
 
-    private void validateWorkflows(Template template) throws Exception {
+    private void validateWorkflows(Template template, String testWorkflowID) throws Exception {
```
owaiskazi19, line 166:
```suggestion
    private void validateWorkflows(Template template, String workflowId) throws Exception {
```

src/main/java/org/opensearch/flowframework/transport/ProvisionWorkflowTransportAction.java
```diff
@@ -156,13 +159,42 @@ protected void doExecute(Task task, WorkflowRequest request, ActionListener<Work
...
         // TODO : Update Action listener type to State index Request
         ActionListener<String> provisionWorkflowListener = ActionListener.wrap(response -> {
-            logger.info("Provisioning completed successuflly for workflow {}", workflowId);
+            logger.info("Provisioning completed successfully for workflow {}", workflowId);
+            flowFrameworkIndicesHandler.updateFlowFrameworkSystemIndexDoc(
+                WORKFLOW_STATE_INDEX,
+                workflowId,
+                ImmutableMap.of(
+                    STATE_FIELD,
+                    State.COMPLETED,
```
owaiskazi19, line 168: Are we changing this state later, if failed?

src/main/java/org/opensearch/flowframework/workflow/CreateConnectorStep.java
```diff
@@ -69,8 +80,38 @@ public CompletableFuture<WorkflowData> execute(List<WorkflowData> data) throws I
...
+                        },
+                            exception -> {
+                                logger.error("Failed to update workflow state with newly created resource: {}", exception.getMessage());
+                            }
+                        )
+                    );
+                } catch (IOException e) {
+                    logger.error("Failed to parse new created resource", e);
+                }
+
```
owaiskazi19, line 114: These should be under `future.complete` below.

src/main/java/org/opensearch/flowframework/workflow/CreateConnectorStep.java
```diff
@@ -69,8 +80,38 @@ public CompletableFuture<WorkflowData> execute(List<WorkflowData> data) throws I
...
+                    flowFrameworkIndicesHandler.updateFlowFrameworkSystemIndexDocWithScript(
+                        WORKFLOW_STATE_INDEX,
+                        workflowId,
+                        script,
+                        ActionListener.wrap(updateResponse -> {
+                            logger.info("updated resources craeted of {}", workflowId);
+                        },
+                            exception -> {
+                                logger.error("Failed to update workflow state with newly created resource: {}", exception.getMessage());
```
owaiskazi19, line 107: Should be complete the future exceptionally here?

src/main/java/org/opensearch/flowframework/workflow/CreateConnectorStep.java
```diff
@@ -69,8 +80,38 @@ public CompletableFuture<WorkflowData> execute(List<WorkflowData> data) throws I
...
+                    XContentBuilder builder = XContentBuilder.builder(XContentType.JSON.xContent());
+                    newResource.toXContent(builder, ToXContentObject.EMPTY_PARAMS);
+
+                    // The script to append a new object to the resources_created array
+                    Script script = new Script(
+                        ScriptType.INLINE,
+                        "painless",
+                        "ctx._source.resources_created.add(params.newResource)",
+                        Collections.singletonMap("newResource", newResource)
+                    );
```
owaiskazi19, line 97: @joshpalis can you verify the painless script here?

owaiskazi19 (commented):
Initial pass. Need tests

src/main/java/org/opensearch/flowframework/rest/RestGetWorkflowAction.java
amitgalitz: I was trying to follow the same pattern as AD here where we have a GET action for everything that has the pattern of `_plugins/ad/detectors/<id>` and then there are different parameters like adding `_profile`. So when we do get workflow after, its the same place, we just can choose whether to add the _status after the id. I can change it too getworkflowstatus for now though or we can keep having the separate actions for get status vs workflow

src/main/java/org/opensearch/flowframework/workflow/CreateConnectorStep.java
amitgalitz: What do you mean by these? do you mean updating the state index? Should it not be a part of the on response here like I have it

src/main/java/org/opensearch/flowframework/workflow/CreateConnectorStep.java
amitgalitz: I can try changing to that, don't have as much experience with completableFuture

src/main/java/org/opensearch/flowframework/rest/RestGetWorkflowAction.java
owaiskazi19: It's fine to keep it as `RestGetWorkflowAction` and let's add `_status` param in the request then.

src/main/java/org/opensearch/flowframework/workflow/CreateConnectorStep.java
owaiskazi19: I meant the whole block should be under `createConnectorFuture.complete` [here](https://github.com/opensearch-project/opensearch-ai-flow-framework/pull/148/files#diff-d523a059b65e2f0888f523cd3329bb573eed90f3826558202b084a1efe2e5856R115)

ohltyler:
Typically params don't have the `_` prefix - suggest to change `_all` -> `all`

src/main/java/org/opensearch/flowframework/transport/GetWorkflowResponse.java
amitgalitz: look at above [comments](https://github.com/opensearch-project/opensearch-ai-flow-framework/pull/148#discussion_r1391750164) regarding reasoning for not having GetWorkflowStateResponse as separate.

src/main/java/org/opensearch/flowframework/transport/ProvisionWorkflowTransportAction.java
amitgalitz: What do you suggest is then done on exception in executeWorkflow, those are still in use in catch block below.

src/main/java/org/opensearch/flowframework/model/ResourcesCreated.java
```diff
@@ -0,0 +1,96 @@
...
+                case RESOURCE_ID_FIELD:
+                    resourceId = parser.text();
+                    break;
+                default:
+                    throw new IOException("Unable to parse field [" + fieldName + "] in a resources_created object.");
+            }
+        }
+        return new ResourcesCreated(resourceName, resourceId);
+    }
+
```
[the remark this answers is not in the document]
amitgalitz: We have a toXContent here, do you mean a ` public static Builder builder()` like in `WorkflowState`. I am not sure its needed here because we will always have the resource workflow step name and id  when building the object.

src/main/java/org/opensearch/flowframework/transport/CreateWorkflowTransportAction.java
amitgalitz: I wanted to call this test cause this is a random workflowID just for validation purpose before we actually index the workflow. I can change to `tempWorkflowId` or `validationWorkflowId`

src/main/java/org/opensearch/flowframework/transport/ProvisionWorkflowTransportAction.java
amitgalitz: yes below on line 190

ohltyler:
> Typically params don't have the `_` prefix - suggest to change `_all` -> `all`

I see this is covered - thanks!

src/main/java/org/opensearch/flowframework/transport/WorkflowRequest.java
amitgalitz: Created a GetWorkflowRequest class

src/main/java/org/opensearch/flowframework/rest/RestGetWorkflowAction.java
```diff
@@ -0,0 +1,107 @@
...
+                throw new FlowFrameworkException("Invalid request format", RestStatus.BAD_REQUEST);
+            }
+            // Validate params
+            String workflowId = request.param(WORKFLOW_ID);
+            if (workflowId == null) {
+                throw new FlowFrameworkException("workflow_id cannot be null", RestStatus.BAD_REQUEST);
+            }
+
+            boolean all = request.paramAsBoolean("_all", false);
+            WorkflowRequest workflowRequest = new WorkflowRequest(workflowId, null, false, all);
```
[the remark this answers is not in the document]
amitgalitz: addressed conflicts

src/main/java/org/opensearch/flowframework/workflow/WorkflowData.java
```diff
@@ -44,6 +45,19 @@ public WorkflowData(Map<String, Object> content) {
     public WorkflowData(Map<String, Object> content, Map<String, String> params) {
         this.content = Map.copyOf(content);
         this.params = Map.copyOf(params);
+        this.workflowId = "";
```
owaiskazi19, line 48: You can directly call the main constructor like:
`this(content, params, "");`

and remove L46-48

src/main/java/org/opensearch/flowframework/workflow/WorkflowData.java
```diff
@@ -23,9 +23,10 @@ public class WorkflowData {
 
     private final Map<String, Object> content;
     private final Map<String, String> params;
+    private String workflowId;
 
     private WorkflowData() {
-        this(Collections.emptyMap(), Collections.emptyMap());
+        this(Collections.emptyMap(), Collections.emptyMap(), "");
```
owaiskazi19, line 29: Why an empty string for workflowId?

src/main/java/org/opensearch/flowframework/transport/GetWorkflowTransportAction.java
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    @Override
+    protected void doExecute(Task task, GetWorkflowRequest request, ActionListener<GetWorkflowResponse> listener) {
+        String workflowId = request.getWorkflowId();
+        User user = ParseUtils.getUserContext(client);
+        GetRequest getRequest = new GetRequest(WORKFLOW_STATE_INDEX).id(workflowId);
+        try (ThreadContext.StoredContext context = client.threadPool().getThreadContext().stashContext()) {
+            client.get(getRequest, ActionListener.runBefore(ActionListener.wrap(r -> {
+                logger.debug("Completed Get Workflow Status Request, id:{}", workflowId);
```
owaiskazi19, line 73: If we want to have this debug statement, then it should be under the try block below

src/main/java/org/opensearch/flowframework/transport/CreateWorkflowTransportAction.java
dbwiddis: I agree we shouldn't use `test` as it has a very specific meaning. I'm missing why we need this, though?  Can we just have a nullable argument?

src/main/java/org/opensearch/flowframework/transport/GetWorkflowTransportAction.java
```diff
@@ -0,0 +1,100 @@
...
+                    }
+                } else {
+                    listener.onFailure(new FlowFrameworkException("Fail to find workflow", RestStatus.NOT_FOUND));
+                }
+            }, e -> {
+                if (e instanceof IndexNotFoundException) {
+                    listener.onFailure(new FlowFrameworkException("Fail to find workflow", RestStatus.NOT_FOUND));
+                } else {
+                    logger.error("Failed to get workflow status of  " + workflowId, e);
+                    listener.onFailure(e);
```
owaiskazi19, line 92: We should create a new `FlowFrameworkException` object and pass that to:

```suggestion
                    listener.onFailure(new FlowFrameworkException(e.getMessage(), ExceptionsHelper.status(e)));
```

Here and for the rest of the failure in the PR.

src/main/java/org/opensearch/flowframework/workflow/CreateConnectorStep.java
```diff
@@ -69,11 +81,41 @@ public CompletableFuture<WorkflowData> execute(List<WorkflowData> data) throws I
...
+
+                    flowFrameworkIndicesHandler.updateFlowFrameworkSystemIndexDocWithScript(
+                        WORKFLOW_STATE_INDEX,
+                        workflowId,
+                        script,
+                        ActionListener.wrap(updateResponse -> {
+                            logger.info("updated resources craeted of {}", workflowId);
```
owaiskazi19, line 108:
```suggestion
                            logger.info("updated resources created of {}", workflowId);
```

src/main/java/org/opensearch/flowframework/workflow/CreateConnectorStep.java
```diff
@@ -69,11 +81,41 @@ public CompletableFuture<WorkflowData> execute(List<WorkflowData> data) throws I
...
+                        }, exception -> {
+                            createConnectorFuture.completeExceptionally(
+                                new FlowFrameworkException(exception.getMessage(), ExceptionsHelper.status(exception))
+                            );
+                            logger.error("Failed to update workflow state with newly created resource: {}", exception);
+                        })
+                    );
+                } catch (IOException e) {
+                    logger.error("Failed to parse new created resource", e);
+                }
```
owaiskazi19, line 118: Let's pull this code block a new private method in ParseUtil class and call that method inside             `createConnectorFuture.complete(`.

We can use the same method to update the resources_created for all the workflow_steps which is missing in this PR.

src/main/java/org/opensearch/flowframework/transport/GetWorkflowTransportAction.java
```diff
@@ -0,0 +1,100 @@
...
+        Client client,
+        NamedXContentRegistry xContentRegistry
+    ) {
+        super(GetWorkflowAction.NAME, transportService, actionFilters, GetWorkflowRequest::new);
+        this.client = client;
+        this.xContentRegistry = xContentRegistry;
+    }
+
+    @Override
+    protected void doExecute(Task task, GetWorkflowRequest request, ActionListener<GetWorkflowResponse> listener) {
```
owaiskazi19, line 67: Need tests for this method.

src/main/java/org/opensearch/flowframework/rest/RestGetWorkflowAction.java
```diff
@@ -0,0 +1,108 @@
...
+        }
+    }
+
+    @Override
+    public List<Route> routes() {
+        return ImmutableList.of(
+            // Provision workflow from indexed use case template
+            new Route(RestRequest.Method.GET, String.format(Locale.ROOT, "%s/{%s}/%s", WORKFLOW_URI, WORKFLOW_ID, "_status"))
+        );
+    }
```
owaiskazi19, line 107: Need tests for this class

owaiskazi19 (commented):
Another pass

owaiskazi19:
@amitgalitz resource created should have the exact entry we have for WorkflowData

```java
createConnectorFuture.complete(
                    new WorkflowData(Map.ofEntries(Map.entry("connector_id", mlCreateConnectorResponse.getConnectorId())))
                );
```


```json

 "resources_created": [
        {
            "resource_id": "fZTmz4sBVvQ80zEwkC2z",
            "connector_id": "XYZ"
        }
    ]
````

File: src/main/java/org/opensearch/flowframework/rest/RestGetWorkflowAction.java (hunk 1 above)
Comment (joshpalis) [reply src/main/java/org/opensearch/flowframework/rest/RestGetWorkflowAction.java:39]: Agreed

File: src/main/java/org/opensearch/flowframework/workflow/WorkflowData.java (hunk 18 above)
Comment (amitgalitz) [reply src/main/java/org/opensearch/flowframework/workflow/WorkflowData.java:29]: I just had this before adding the workflowId and resources created to each step but I can that now at least for the workflowId part

File: src/main/java/org/opensearch/flowframework/transport/GetWorkflowTransportAction.java (hunk 19 above)
Comment (amitgalitz) [reply src/main/java/org/opensearch/flowframework/transport/GetWorkflowTransportAction.java:73]: removed actually

File: src/main/java/org/opensearch/flowframework/workflow/CreateConnectorStep.java (hunk 22 above)
Comment (amitgalitz) [reply src/main/java/org/opensearch/flowframework/workflow/CreateConnectorStep.java:118]: Io am okay trying to abstract this in future PR to resolve this [issue](https://github.com/opensearch-project/opensearch-ai-flow-framework/issues/136), not to block actual get status API content

**Comment by owaiskazi19** [comment]
> resource_id and connector_id are the same here

then let's call it connector_id

**Comment by amitgalitz** [comment]
> > resource_id and connector_id are the same here
> 
> then let's call it connector_id

I want it to be static field though so its cleaner on searching and mapping, the keys of resource_id and workflow_step_name wont change between resources. If you still think otherwise thats fair, more info here: https://github.com/opensearch-project/opensearch-ai-flow-framework/issues/136. This also might have some small changes as Dan implements the delete api.

File: src/main/java/org/opensearch/flowframework/model/WorkflowState.java [hunk 25]
```diff
@@ -196,8 +223,8 @@ public Builder userOutputs(Map<String, Object> userOutputs) {
          * @param resourcesCreated resourcesCreated
          * @return the Builder object
          */
-        public Builder resourcesCreated(Map<String, Object> resourcesCreated) {
-            this.userOutputs = resourcesCreated;
+        public Builder resourcesCreated(List<ResourcesCreated> resourcesCreated) {
+            this.resourcesCreated = resourcesCreated;
```
ohltyler, line 227: what if there's no created resources? Or will the caller pass an empty list if so?

src/main/java/org/opensearch/flowframework/rest/RestGetWorkflowAction.java
```diff
@@ -0,0 +1,108 @@
...
+            return channel -> client.execute(GetWorkflowAction.INSTANCE, getWorkflowRequest, ActionListener.wrap(response -> {
+                XContentBuilder builder = response.toXContent(channel.newBuilder(), ToXContent.EMPTY_PARAMS);
+                channel.sendResponse(new BytesRestResponse(RestStatus.OK, builder));
+            }, exception -> {
+                try {
+                    FlowFrameworkException ex = new FlowFrameworkException(exception.getMessage(), ExceptionsHelper.status(exception));
+                    XContentBuilder exceptionBuilder = ex.toXContent(channel.newErrorBuilder(), ToXContent.EMPTY_PARAMS);
+                    channel.sendResponse(new BytesRestResponse(ex.getRestStatus(), exceptionBuilder));
+                } catch (IOException e) {
+                    logger.error("Failed to send back provision workflow exception", e);
```
ohltyler, line 90: nit - failed to send back get workflow exception?

src/main/java/org/opensearch/flowframework/rest/RestGetWorkflowAction.java
```diff
@@ -0,0 +1,108 @@
...
+            return channel -> channel.sendResponse(
+                new BytesRestResponse(ex.getRestStatus(), ex.toXContent(channel.newErrorBuilder(), ToXContent.EMPTY_PARAMS))
+            );
+        }
+    }
+
+    @Override
+    public List<Route> routes() {
+        return ImmutableList.of(
+            // Provision workflow from indexed use case template
```
ohltyler, line 104: nit - remove provision wording here

src/main/java/org/opensearch/flowframework/transport/GetWorkflowRequest.java
```diff
@@ -0,0 +1,84 @@
...
+/**
+ * Transport Request to get a workflow or workflow status
+ */
+public class GetWorkflowRequest extends ActionRequest {
+
+    /**
+     * The documentId of the workflow entry within the Global Context index
+     */
+    @Nullable
+    private String workflowId;
```
ohltyler, line 28: why can workflowId be nullable?

src/main/resources/mappings/workflow-state.json
```diff
@@ -31,7 +31,15 @@
...
-      "type": "object"
+      "type": "nested",
+      "properties": {
+        "workflow_step_name": {
+          "type": "keyword"
+        },
+        "resource_id": {
+          "type": "keyword"
+        }
+      }
```
ohltyler, line 42: can we persist resource types as well? Frontend mocks currently have this included, e.g., "index", "model/connector"

src/main/java/org/opensearch/flowframework/transport/CreateWorkflowTransportAction.java
amitgalitz: change to just using nulllable, good call

src/main/java/org/opensearch/flowframework/workflow/WorkflowData.java
amitgalitz: changed this ^ added workflow ID to all steps

src/main/java/org/opensearch/flowframework/rest/RestGetWorkflowAction.java
amitgalitz: added

src/main/resources/mappings/workflow-state.json
amitgalitz: We could do this, but we would need some better mapping then what we have now: discussed some of this here: https://github.com/opensearch-project/opensearch-ai-flow-framework/issues/136

right now we just have the step name which means we need some mapping between step name and resources itself or to that info somewhere earlier.

src/main/java/org/opensearch/flowframework/transport/GetWorkflowRequest.java
amitgalitz: This is just for the createWorkflow dry run case. the dry run is executed as a validation before we index and before we have a workflow ID, previously I passed a randomly generated ID, I have that part commented out now, I was responding to comment by Dan to make it nullable if thats okay I will remove comment

src/main/java/org/opensearch/flowframework/model/WorkflowState.java
amitgalitz: We create the state here, user wont ever change the resources created, I initiate it with an empty list though so it takes care of that case.

src/main/java/org/opensearch/flowframework/transport/GetWorkflowTransportAction.java
amitgalitz: changed

src/main/resources/mappings/workflow-state.json
ohltyler: I see. I think it will be important to portray to the user a readable list of created resources to provide good transparency as to what this plugin is doing behind the scenes. The user should easily be able to see all associated connectors, knn indices, etc. that they've created so they can keep track of things. Just a ref to the workflow step name doesn't seem sufficient.

I'll add a comment in #136 since this is still a WIP. No blocker for this PR

src/main/java/org/opensearch/flowframework/transport/GetWorkflowTransportAction.java
amitgalitz: added along with tests on other parts of the transport action request and response, not everything explicitly mocked can add, also more will be added with integ tests

src/main/java/org/opensearch/flowframework/transport/ProvisionWorkflowTransportAction.java
amitgalitz: Removing this actually breaks the code right now, meaning we don't change state to complete or failed

joshpalis: Well since we're not using the onResponse value of `READY`, we can just invoke `updateFlowFrameworkSystemIndexDoc` after `workflowFutureList.forEach(CompletableFuture::join);` and also within the catch block. 

Why use an action listener if the given value is not used onResponse?

src/main/java/org/opensearch/flowframework/model/ResourcesCreated.java
```diff
@@ -0,0 +1,122 @@
...
+import static org.opensearch.flowframework.common.CommonValue.RESOURCE_ID_FIELD;
+import static org.opensearch.flowframework.common.CommonValue.WORKFLOW_STEP_NAME;
+
+/**
+ * This represents an object in the WorkflowState {@link WorkflowState}.
+ */
+public class ResourcesCreated implements ToXContentObject, Writeable {
+
+    private String workflowStepName;
+    private String resourceId;
```
dbwiddis, line 30: These should be made `final`. 

The record-style getters we are using imply immutability, and we only ever set these (once) in the constructors.

dbwiddis, line 30: How are we handling the case where multiple resources may be created by a workflow step?  

I'm not sure how a `String` gives me enough information to identify the resource created. Was it an index? A connector? A model?   Or is this JSON content with more detail (in which case naming it 'id' isn't right.

src/main/java/org/opensearch/flowframework/model/WorkflowState.java
```diff
@@ -69,7 +78,7 @@ public WorkflowState(
         Instant provisionEndTime,
         User user,
         Map<String, Object> userOutputs,
-        Map<String, Object> resourcesCreated
+        List<ResourcesCreated> resourcesCreated
```
dbwiddis, line 81: So I think I answered my previous question on more than one resource; it's handled here.  But in that case:
 - we should probably name the other class `ResourceCreated`
 - we still should have some sort of type identifier on what sort of resource it is (index, connector, etc.).  An enum is probably fine for that.

src/main/java/org/opensearch/flowframework/transport/CreateWorkflowTransportAction.java
```diff
@@ -97,6 +98,8 @@ protected void doExecute(Task task, WorkflowRequest request, ActionListener<Work
 
         if (request.isDryRun()) {
             try {
+                // generating random workflowId only for validation purpose
+                String uniqueID = UUID.randomUUID().toString();
```
dbwiddis, line 102: Where is this UUID used?

There are multiple variants of random UUIDs, is there a particular pattern required here?

dbwiddis (commented):
Generally looks good, but need to understand more about how the id String tells me what sort of resource I will need to remove later.   

Are we assuming a one-to-one correspondence between workflow step and resource?  I'm not sure how certain that assumption is.

src/main/java/org/opensearch/flowframework/transport/ProvisionWorkflowTransportAction.java
amitgalitz: removed it

src/main/java/org/opensearch/flowframework/transport/CreateWorkflowTransportAction.java
amitgalitz: removed, I previously used this only for the `dryRun` purpose that Josh added for validation, but instead I just switched workflowId to be @Nullable so it still works.

src/main/java/org/opensearch/flowframework/model/WorkflowState.java
amitgalitz: yes will change to `ResourceCreated`. Regarding adding some variable for each step, Tyler also mentioned that with additional section for associated resource (resource we modify but don't create (knn index can be a common one with adding a pipeline to it), so users can keep track of that). 

I think it would make sense to add that in additional PR, I mentioned it in the description with the issue I have for it. I can add a TODO as well

src/main/java/org/opensearch/flowframework/model/ResourcesCreated.java
amitgalitz: Like you mentioned below, we have a list of resources created, we can add multiple of them to workflowData and index if a step requires it. Regarding info to identify the resource created, lets move that to a different PR as we are discussing it on a separate issue and i want to merge this API.

src/main/java/org/opensearch/flowframework/rest/RestGetWorkflowAction.java
amitgalitz: added

dbwiddis (approved):
LGTM with pending TODO handled in another PR

ohltyler (approved):
Read latest updates and comments, no concerns. Giving a soft approval :)

src/main/java/org/opensearch/flowframework/workflow/CreateConnectorStep.java
amitgalitz: Will have more tests when adding created resources is fully complete

PR #155, opened 2023-11-07
Separates RegisterModelStep into RegisterLocalModelStep and RegisterRemoteModelStep. Adds GetMLTaskStep. Handles optional params

joshpalis:
### Description
Creates a new workflow step to register a local model based on [API documentation](https://opensearch.org/docs/latest/ml-commons-plugin/api/#request-fields)

Creates a separate workflow step to register a remote model based on [API Documentation](https://opensearch.org/docs/latest/ml-commons-plugin/extensibility/index/#step-3-register-a-remote-model).

Creates a `GetMLTaskStep` to retrieve the model ID given the task ID outputted by the `RegisterLocalModelStep` 
(Necessary to first merge in https://github.com/opensearch-project/ml-commons/pull/1616 prior to this PR)

Note : I have not added the internal connector configuration capability to the `RegisterRemoteModelStep` since we already have a `CreateConnectorStep`, but have marked it as a TODO later down the line

### Issues Resolved
https://github.com/opensearch-project/opensearch-ai-flow-framework/issues/149
Part of #88 


joshpalis:
Still need to handle create connector optional params, will include in the next commit

owaiskazi19:
> Still need to handle create connector optional params, will include in the next commit

@joshpalis raise a separate PR for that.

src/main/java/org/opensearch/flowframework/workflow/RegisterLocalModelStep.java
```diff
@@ -0,0 +1,198 @@
...
+                        break;
+
+                }
+            }
+        }
+
+        if (Stream.of(
+            modelName,
+            modelVersion,
+            description,
```
owaiskazi19, line 151: From [here](https://github.com/opensearch-project/ml-commons/blob/main/common/src/main/java/org/opensearch/ml/common/transport/register/MLRegisterModelInput.java#L94-L108), looks like only modelName, functionaName, modelFormat, modelConfig and url are mandatory fields not rest.

src/main/java/org/opensearch/flowframework/workflow/RegisterRemoteModelStep.java
```diff
@@ -127,28 +115,32 @@ public void onFailure(Exception e) {
             }
         }
 
-        if (Stream.of(functionName, modelName, description, connectorId).allMatch(x -> x != null)) {
+        if (Stream.of(modelName, functionName, description, connectorId).allMatch(x -> x != null)) {
```
owaiskazi19, line 118: Description is not a mandatory field.

owaiskazi19 (commented):
Are we planning to have a separate PR for optional params of Register Steps?

src/main/java/org/opensearch/flowframework/workflow/RegisterLocalModelStep.java
joshpalis: model config is comprised of the `modelType` `embeddingDImension`, `frameworkType`, with an optional `all_config` field. We're not directly reading the model config from the `WorkflowData`, instead we check for the aforementioned fields and build the config in this step. 

This is based on the documentation ([link](https://opensearch.org/docs/latest/ml-commons-plugin/api/#request-fields)).

Additionally, there is no `function_name` for a local model registration, that is for the `RegisterRemoteModelStep`

joshpalis:
@owaiskazi19 optional params have been handled in this PR

src/main/java/org/opensearch/flowframework/workflow/RegisterLocalModelStep.java
owaiskazi19: We should create ModelConfig object before the `if` condition then and handle the optional params of modelCondig before creating that object.

src/main/java/org/opensearch/flowframework/workflow/RegisterLocalModelStep.java
```diff
@@ -0,0 +1,190 @@
...
+                .normalizeResult(false);
+
+            if (allConfig != null) {
+                modelConfigBuilder.allConfig(allConfig);
+            }
+
+            MLModelConfig modelConfig = modelConfigBuilder.build();
+
+            MLRegisterModelInput mlInput = MLRegisterModelInput.builder()
+                .modelName(modelName)
```
owaiskazi19, line 167: We are not adding `description` field here. It can be optional but need to be included if present. For handling optional params refer [this](https://github.com/opensearch-project/opensearch-ai-flow-framework/blob/main/src/main/java/org/opensearch/flowframework/workflow/ModelGroupStep.java#L118-L132)

joshpalis: ah good catch, I have added this to both register remote model and register local model

src/main/java/org/opensearch/flowframework/workflow/RegisterLocalModelStep.java
```diff
@@ -0,0 +1,196 @@
...
+            .allMatch(x -> x != null)) {
+
+            // Create model configuration, assuming null pooling mode, null model max length, normalize results set to false
+            TextEmbeddingModelConfigBuilder modelConfigBuilder = TextEmbeddingModelConfig.builder()
+                .modelType(modelType)
+                .embeddingDimension(Integer.valueOf(embeddingDimension))
+                .frameworkType(frameworkType)
+                .poolingMode(null)
+                .modelMaxLength(null)
+                .normalizeResult(false);
```
owaiskazi19, line 159: We are setting these values statically. Any reason?

src/main/java/org/opensearch/flowframework/workflow/RegisterLocalModelStep.java
```diff
@@ -0,0 +1,196 @@
...
+        CompletableFuture<WorkflowData> registerLocalModelFuture = new CompletableFuture<>();
+
+        ActionListener<MLRegisterModelResponse> actionListener = new ActionListener<>() {
+            @Override
+            public void onResponse(MLRegisterModelResponse mlRegisterModelResponse) {
+                logger.info("Local Model registration successful");
+                registerLocalModelFuture.complete(
+                    new WorkflowData(
+                        Map.ofEntries(
+                            Map.entry(MODEL_ID, mlRegisterModelResponse.getModelId()),
```
owaiskazi19, line 78: Can you verify once the register API present on `feature/agent_dev` branch is returning model_id and not task_id. If it returns taskId then we might need a GetTaskStep to get the model id from task id.

src/main/resources/mappings/workflow-steps.json
```diff
@@ -39,18 +39,34 @@
             "connector_id"
         ]
     },
-    "register_model": {
+    "register_local_model": {
         "inputs":[
-            "function_name",
             "name",
-            "description",
```
owaiskazi19, line 46: We can include the description field here. Or are we just including the required fields here for validation?

joshpalis: these are only the required fields

src/main/java/org/opensearch/flowframework/workflow/RegisterLocalModelStep.java
```diff
@@ -0,0 +1,196 @@
...
+                        modelVersion = (String) content.get(VERSION_FIELD);
+                        break;
+                    case DESCRIPTION_FIELD:
+                        description = (String) content.get(DESCRIPTION_FIELD);
+                        break;
+                    case MODEL_FORMAT:
+                        modelFormat = MLModelFormat.from((String) content.get(MODEL_FORMAT));
+                        break;
+                    case MODEL_GROUP_ID:
+                        modelGroupId = (String) content.get(MODEL_GROUP_ID);
```
owaiskazi19, line 122: How are we handling
1. If modelId is present in request/template, use that.
2. If not, check if WorkflowData has one, because user can use model group step to create one.

src/main/java/org/opensearch/flowframework/workflow/RegisterLocalModelStep.java
joshpalis: These are the default values for remote model registration, but looking through the [TextEmbeddingModelConfig](https://github.com/opensearch-project/ml-commons/blob/main/common/src/main/java/org/opensearch/ml/common/model/TextEmbeddingModelConfig.java) class again, I can remove these from our end

src/main/java/org/opensearch/flowframework/workflow/RegisterLocalModelStep.java
```diff
@@ -0,0 +1,196 @@
...
+                        url = (String) content.get(URL);
+                        break;
+                    default:
+                        break;
+
+                }
+            }
+        }
+
+        if (Stream.of(modelName, modelVersion, modelFormat, modelGroupId, embeddingDimension, frameworkType, modelContentHashValue, url)
```
owaiskazi19, line 149: `embeddingDimension` is not a mandatory field based on [this](https://opensearch.org/docs/latest/ml-commons-plugin/api/#request-fields). Also, since mlConfig is not mandatory, we can't have `embeddingDimension` mandatory.

joshpalis: That documentation clearly states that the only optional field in model config is `all_config`, the remaining fields are required

owaiskazi19: Ah! You are right. I missed the descriptionof it.

owaiskazi19: Even the modelType is a mandatory field, right?

joshpalis: ah missed that too

owaiskazi19: Checking these params requirement can be a little confusing.

src/main/java/org/opensearch/flowframework/workflow/RegisterLocalModelStep.java
joshpalis: Good call out, the `MLRegisterModelResponse` has an optional `modelD` field, but the `TransportMLRegisterModelAction` just responds with the taskId [here](https://github.com/opensearch-project/ml-commons/blob/7e44eb282f0d3f6deeb07993d83a00bcfcd5edd5/plugin/src/main/java/org/opensearch/ml/action/register/TransportRegisterModelAction.java#L267C35-L267C35) after it dispatches the task to another node. I can add the ml task step in my next commit

src/main/java/org/opensearch/flowframework/workflow/RegisterLocalModelStep.java
joshpalis: Well the model group ID wouldn't be present in the request/template since it'll be generated by the model group step. If the user doesnt have a model group step that connects to the register local model step then it'll fail validation

src/main/java/org/opensearch/flowframework/workflow/RegisterLocalModelStep.java
owaiskazi19: A month old work I did before.
```java
/*
 * Copyright OpenSearch Contributors
 * SPDX-License-Identifier: Apache-2.0
 *
 * The OpenSearch Contributors require contributions made to
 * this file be licensed under the Apache-2.0 license or a
 * compatible open source license.
 */
package org.opensearch.flowframework.workflow.Model;

import org.apache.logging.log4j.LogManager;
import org.apache.logging.log4j.Logger;
import org.opensearch.client.Client;
import org.opensearch.client.node.NodeClient;
import org.opensearch.flowframework.client.MLClient;
import org.opensearch.flowframework.workflow.WorkflowData;
import org.opensearch.flowframework.workflow.WorkflowStep;
import org.opensearch.ml.client.MachineLearningNodeClient;
import org.opensearch.ml.common.MLTask;

import java.io.IOException;
import java.util.List;
import java.util.Map;
import java.util.Map.Entry;
import java.util.concurrent.CompletableFuture;

public class GetTaskStep implements WorkflowStep {

    private static final Logger logger = LogManager.getLogger(GetTaskStep.class);

    private Client client;

    private final String NAME = "get_task";
    private static final String TASK_ID = "task_id";

    public GetTaskStep(Client client) {
        this.client = client;
    }

    @Override
    public CompletableFuture<WorkflowData> execute(List<WorkflowData> data) {
        CompletableFuture<WorkflowData> getTaskFuture = new CompletableFuture<>();

        String taskId = null;

            Map<String, Object> content = workflowData.getContent();

            for (Entry<String, Object> entry : content.entrySet()) {
                if (entry.getKey() == TASK_ID) {
                    taskId = (String) content.get(TASK_ID);
                    break;
                }
            }

        }

        if (taskId != null) {
            MachineLearningNodeClient machineLearningNodeClient = MLClient.createMLClient((NodeClient) client);

            MLTask mlTaskResponse = machineLearningNodeClient.getTask(taskId).actionGet();

            getTaskFuture.complete(new WorkflowData(Map.of("modelId", mlTaskResponse.getModelId())));
        } else {
            logger.error("Failed to fetch model");
            getTaskFuture.completeExceptionally(new IOException("Failed to fetch model"));
        }

        return getTaskFuture;
    }

    @Override
    public String getName() {
        return NAME;
    }
}
```

src/main/java/org/opensearch/flowframework/workflow/RegisterLocalModelStep.java
owaiskazi19: How are we handling if user provides a model_group id or connector id instead of using the template?

joshpalis: user inputs / outputs from previous workflow steps are all in `WorkflowData` either way, so checking for this field as we iterate through the workflow data will cover both cases. 

In terms of validation, we check for required inputs from both provided user inputs and connected node outputs, so if a user provides a model_group_id/connector_id instead of including the corresponding steps then it will pass

owaiskazi19: Makes sense.

src/main/java/org/opensearch/flowframework/workflow/RegisterRemoteModelStep.java
```diff
@@ -31,89 +30,78 @@
...
             public void onResponse(MLRegisterModelResponse mlRegisterModelResponse) {
-                logger.info("Model registration successful");
-                registerModelFuture.complete(
+                logger.info("Remote Model registration successful");
+                registerRemoteModelFuture.complete(
                     new WorkflowData(
                         Map.ofEntries(
-                            Map.entry("model_id", mlRegisterModelResponse.getModelId()),
-                            Map.entry("register_model_status", mlRegisterModelResponse.getStatus())
+                            Map.entry(MODEL_ID, mlRegisterModelResponse.getModelId()),
```
owaiskazi19, line 69: Just to confirm. Remote model returns modelId instead of taskId, right?

joshpalis: yes, its not an async call

owaiskazi19 (approved):
LGTM! Thanks for patiently addressing all the comment.

dbwiddis (approved):
LGTM, and commented on the retry TODO issue

PR #156, opened 2023-11-10
Moved UI_METADATA from state to GC index and excluded it from search …

owaiskazi19:
### Description
1. Moved UI_METADATA from state to GC index
2. Excluded UI_METADATA from search response for now
3. Created Builder for Template Model

### Issues Resolved
Closes https://github.com/opensearch-project/opensearch-ai-flow-framework/issues/140


src/main/java/org/opensearch/flowframework/model/Template.java
```diff
@@ -74,9 +78,129 @@ public Template(
...
+            this.user = user;
+            return this;
+        }
+
+        /**
+         * Allows building a template
+         * @return Template Object containing all needed fields
+         */
+        public Template build() {
+            Template template = new Template();
```
joshpalis, line 190: When parsing a template object, we require a name, otherwise it fails to be created. Should we similarly have the same guardrails here when creating a Template via builder?

joshpalis (approved):
Approved with a comment

dbwiddis (approved):

src/main/java/org/opensearch/flowframework/model/Template.java
owaiskazi19: This is just an empty private constructor. I think it's fine to have it like this.

PR #178, opened 2023-11-20
Check if Index exists before max workflow check

### LINKED ISSUE #174: [BUG] Checking max workflow check before checking if index exists
### What is the bug?
We currently can't create a workflow if there is no global context index available, this is due to the fact that we are checking for the total workflows in the index in order to enable our throttling but we don't check if the index exists first

### Solution

We should check if the index exists before checking the number of workflows we currently have

amitgalitz:
### Description
Checking if GC index exists before checking how many existing current workflows we have,  and initializing if it doesn't

All I did was wrap it with following code, github change looks weird 
```

        flowFrameworkIndicesHandler.initGlobalContextIndexIfAbsent(ActionListener.wrap(indexCreated -> {
            if (!indexCreated) {
                listener.onFailure(new FlowFrameworkException("No response to create global_context index", INTERNAL_SERVER_ERROR));
                return;
            }
            
            
 ....
 
 e -> {
            logger.error("Failed to create global_context index", e);
            listener.onFailure(new FlowFrameworkException("Failed to create global_context index", INTERNAL_SERVER_ERROR));
        }));
            
 ```

### Issues Resolved
resolves #174

### DIFF
--- a/src/main/java/org/opensearch/flowframework/transport/CreateWorkflowTransportAction.java
+++ b/src/main/java/org/opensearch/flowframework/transport/CreateWorkflowTransportAction.java
@@ -39,6 +39,7 @@ import org.opensearch.transport.TransportService;
 
 import java.util.List;
 
+import static org.opensearch.core.rest.RestStatus.INTERNAL_SERVER_ERROR;
 import static org.opensearch.flowframework.common.CommonValue.PROVISIONING_PROGRESS_FIELD;
 import static org.opensearch.flowframework.common.CommonValue.STATE_FIELD;
 import static org.opensearch.flowframework.util.ParseUtils.getUserContext;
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/main/java/org/opensearch/flowframework/transport/CreateWorkflowTransportAction.java [hunk 1]
```diff
@@ -109,83 +110,100 @@ protected void doExecute(Task task, WorkflowRequest request, ActionListener<Work
...
+                checkMaxWorkflows(request.getRequestTimeout(), request.getMaxWorkflows(), ActionListener.wrap(max -> {
+                    if (!max) {
+                        String errorMessage = "Maximum workflows limit reached " + request.getMaxWorkflows();
+                        logger.error(errorMessage);
+                        FlowFrameworkException ffe = new FlowFrameworkException(errorMessage, RestStatus.BAD_REQUEST);
+                        listener.onFailure(ffe);
+                        return;
+                    } else {
+                        // Create new global context and state index entries
+                        flowFrameworkIndicesHandler.putTemplateToGlobalContext(
```
owaiskazi19, line 129: This method creates a GC index in it. `initGlobalContextIndexIfAbsent` method.

amitgalitz: yeah I originally thought I can just move it earlier and forgot to remove that init from putTemplateToGlobalContext but just took Josh's suggestion and added the check for if index exists earlier since no index means max workflows hasn't been reached

joshpalis (approved):

src/main/java/org/opensearch/flowframework/transport/CreateWorkflowTransportAction.java
```diff
@@ -195,6 +195,10 @@ protected void doExecute(Task task, WorkflowRequest request, ActionListener<Work
      *  @param internalListener listener for search request
      */
     protected void checkMaxWorkflows(TimeValue requestTimeOut, Integer maxWorkflow, ActionListener<Boolean> internalListener) {
+        if (!flowFrameworkIndicesHandler.doesIndexExist(CommonValue.GLOBAL_CONTEXT_INDEX)) {
+            internalListener.onResponse(true);
+            return;
+        }
```
owaiskazi19, line 201: Can we add a small test to cover this?

src/main/java/org/opensearch/flowframework/transport/CreateWorkflowTransportAction.java
```diff
@@ -195,6 +195,10 @@ protected void doExecute(Task task, WorkflowRequest request, ActionListener<Work
      *  @param internalListener listener for search request
      */
     protected void checkMaxWorkflows(TimeValue requestTimeOut, Integer maxWorkflow, ActionListener<Boolean> internalListener) {
+        if (!flowFrameworkIndicesHandler.doesIndexExist(CommonValue.GLOBAL_CONTEXT_INDEX)) {
+            internalListener.onResponse(true);
+            return;
```
owaiskazi19, line 200: Let's have the rest of the code under `else` rather returning from here. We should persist the listener throughout the method.

owaiskazi19 (approved):

PR #191, opened 2023-11-22
Validate that all node IDs are unique

### LINKED ISSUE #189: [BUG] Validate that all step IDs are unique
### What is the bug?
Currently we are allowed to create a template where we have duplicate step IDs throughout the steps. It would be beneficial to ensure these step IDs are unique so they can be used to identify different steps, especially in the resources created section.

This can be crucial for users wanting to know which step they listed corresponds to the created resource id.

### Solution:

We can add validation to make sure all step_id's are unique, this can be done by adding the step_id to a set and make sure we don't have duplicates.

dbwiddis:
### Description

Adds a validation check that all workflow nodes have a unique ID.

### Issues Resolved

Fixes #189

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -146,7 +146,8 @@ dependencies {
     configurations.all {
         resolutionStrategy {
             force("com.google.guava:guava:32.1.3-jre") // CVE for 31.1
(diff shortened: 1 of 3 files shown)


joshpalis (approved):
LGTM thanks Dan!

amitgalitz (approved):
Thanks for the quick fix!

PR #231, opened 2023-11-30
[feature/agent_framework] Changing resources created format 

### LINKED ISSUE #136: [FEATURE] Resources created in state index and de-provisioning mapping

### LINKED ISSUE #173: [FEATURE] Create an enum for Resources

To support the deprovisioning workflow we need a consistent set of Strings associated with resource types (e.g. `connector_id`).

Currently many of these are constant strings in the `common` package which is a good first step.  Some of them are hardcoded in the workflow steps which is not a good practice.

Create an enum associated with Resources created where the enum name can be used (lowercased) for the JSON string key.

amitgalitz:
### Description

Added updates to resources_created section for register_model_group, register model, create index, create ingest pipeline and create connector steps.

Still didn't add this for register agent since there is more changes that will be merged for that soon. 

I also added `retryOnConflict` here based on same number as ml-commons has, since adding this I haven't seen the version conflict issues but I will continue to see if there are better concurrency controls we can have, added a TODO for this.

Changed the format of creating a resource to: 
```
 {
                "_index": ".plugins-workflow-state",
                "_id": "RvEFF4wBKbeLfRazqAdh",
                "_score": 1.0,
                "_source": {
                    "workflow_id": "RvEFF4wBKbeLfRazqAdh",
                    "state": "PROVISIONING",
                    "provisioning_progress": "IN_PROGRESS",
                    "provision_start_time": 1701193321252,
                    "resources_created": [
                        {
                            "workflow_step_id": "workflow_step_1-amit-test",
                            "connector_id": "TfEGF4wBKbeLfRazCwcp"
                        },
                        {
                            "workflow_step_id": "workflow_step_2",
                            "model_id": "T_EGF4wBKbeLfRazCwdx"
                        }
                    ],
                    "provision_end_time": 1701193305145
                }
            }
```

### Issues Resolved
 resolves #173 
 resolves #136 


joshpalis (commented):
Initial pass, looks great, but just had a few thoughts

src/main/java/org/opensearch/flowframework/indices/FlowFrameworkIndicesHandler.java
```diff
@@ -478,4 +485,41 @@ public void updateFlowFrameworkSystemIndexDocWithScript(
...
+     * @param resourceId the id of the newly created resource
+     * @param completableFuture the CompletableFuture used for this step
+     * @throws IOException if parsing fails on new resource
+     */
+    public void updateResourceInStateIndex(
+        String workflowId,
+        String nodeId,
+        String workflowStepName,
+        String resourceId,
+        CompletableFuture<WorkflowData> completableFuture
```
joshpalis, line 503: Instead of a `CompletableFuture` here, maybe we can use an action listener instead. Then within each `WorkflowStep` we can complete `onResponse` or `CompleteExceptionally` on failure.

joshpalis: So in each workflow step, the updates would look something like : 
```java
flowFrameworkIndicesHandler.updateResourceInStateIndex(
                        currentNodeInputs.getWorkflowId(),
                        currentNodeId,
                        getName(),
                        mlCreateConnectorResponse.getConnectorId(),
                        ActionListener.wrap(
                            onResponse -> {
                                future.complete(WorkflowData);
                            },
                            onFailure -> {
                                future.completeExceptionally(onFailure.getMessage());
                            }
                        )
                    );
```

src/main/java/org/opensearch/flowframework/common/WorkflowResources.java
```diff
@@ -0,0 +1,87 @@
...
+        return resourceCreated;
+    }
+
+    /**
+     * gets the resources created type based on the workflowStep
+     * @param workflowStep workflow step name
+     * @return the resource that will be created
+     * @throws IOException if workflow step doesn't exist in enum
+     */
+    public static String getResourceByWorkflowStep(String workflowStep) throws IOException {
```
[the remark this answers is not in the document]
amitgalitz: Right now I haven't seen a step that returns multiple resources. It makes it a little more complicated if we have to do a mapping of step 1 creates two fields "connector_id" and "model_id" for example and then we map each ID to the correct one when updating. If you think its likely we will have a step like this I can change the logic

src/main/java/org/opensearch/flowframework/common/WorkflowResources.java
```diff
@@ -0,0 +1,87 @@
...
+        logger.error("Unable to find resource type for step: " + workflowStep);
+        throw new IOException("Unable to find resource type for step: " + workflowStep);
+    }
+
+    /**
+     * Returns all the possible resource created types in enum
+     * @return a set of all the resource created types
+     */
+    public static Set<String> getAllResourcesCreated() {
+        return Stream.of(values()).map(WorkflowResources::getResourceCreated).collect(Collectors.toSet());
```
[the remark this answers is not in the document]
amitgalitz: changed

src/main/java/org/opensearch/flowframework/workflow/CreateConnectorStep.java
```diff
@@ -87,44 +81,28 @@ public CompletableFuture<WorkflowData> execute(
...
-                    new WorkflowData(
-                        Map.ofEntries(Map.entry("connector_id", mlCreateConnectorResponse.getConnectorId())),
-                        workflowId,
-                        currentNodeInputs.getNodeId()
-                    )
-                );
+
                 try {
+                    String resourceName = WorkflowResources.getResourceByWorkflowStep(getName());
+                    createConnectorFuture.complete(
```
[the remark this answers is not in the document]
amitgalitz: made this change

joshpalis (approved):
Great work @amitgalitz ! Thanks for addressing the changes

src/main/java/org/opensearch/flowframework/common/WorkflowResources.java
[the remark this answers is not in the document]
amitgalitz: I could add this with my PR for converting the workflow-step.json to an enum if that is okay.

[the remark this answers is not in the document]
amitgalitz: tried this out and there are some things to still decide on and I can expand in an issue. Basically making sure we map the right resource key to the right resource value could be a little tricky. For example if we have a list of resourcesCreatedTypes like (“connector_id”, “model_id”, “detector_id”). We need to create a map with each one of those as the key and to know which resource to get but that is arbitrary on how the resource looks like in the response.

```
 List<String> resourceNames = WorkflowResources.getResourcesByWorkflowStep(getName());
 Map<String, Object> map = resourceNames.stream()
            .collect(Collectors.toMap(
                    key -> key,                       // Key Mapper: Use the list item as the key.
                    value -> {//getting correct corresponding value might not be straightforward cause we would}
                        // a mapping of each possible resource to the correct method for example"
                        // "connector_id" -> mlCreateConnectorResponse.getConnectorId()
                        // "detector" -> however that client defines it as and etc. 
            ));

```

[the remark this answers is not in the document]
owaiskazi19: I agree with @dbwiddis here. We should return a list rather a single resource. Follow up PR should be fine.

src/main/java/org/opensearch/flowframework/common/WorkflowResources.java
```diff
@@ -0,0 +1,90 @@
...
+    public static String getResourceByWorkflowStep(String workflowStep) throws IOException {
+        if (workflowStep != null && !workflowStep.isEmpty()) {
+            for (WorkflowResources mapping : values()) {
+                if (mapping.getWorkflowStep().equals(workflowStep)) {
+                    return mapping.getResourceCreated();
+                }
+            }
+        }
+        logger.error("Unable to find resource type for step: " + workflowStep);
+        throw new IOException("Unable to find resource type for step: " + workflowStep);
```
owaiskazi19, line 80: FlowFrameworkException here?

src/main/java/org/opensearch/flowframework/indices/FlowFrameworkIndicesHandler.java
```diff
@@ -468,7 +472,8 @@ public void updateFlowFrameworkSystemIndexDocWithScript(
                 // TODO: Also add ability to change other fields at the same time when adding detailed provision progress
                 updateRequest.script(script);
                 updateRequest.setRefreshPolicy(WriteRequest.RefreshPolicy.IMMEDIATE);
-                // TODO: decide what condition can be considered as an update conflict and add retry strategy
+                updateRequest.retryOnConflict(5);
```
owaiskazi19, line 475: How are we deciding the numbers here `5` here and `3` on L441?

src/main/java/org/opensearch/flowframework/model/ResourceCreated.java
```diff
@@ -102,22 +123,50 @@ public static ResourceCreated parse(XContentParser parser) throws IOException {
...
                     break;
-                case RESOURCE_ID_FIELD:
-                    resourceId = parser.text();
+                case WORKFLOW_STEP_ID:
+                    workflowStepId = parser.text();
                     break;
                 default:
-                    throw new IOException("Unable to parse field [" + fieldName + "] in a resources_created object.");
+                    if (!isValidFieldName(fieldName)) {
+                        throw new IOException("Unable to parse field [" + fieldName + "] in a resources_created object.");
```
owaiskazi19, line 131: FlowFrameworkException?

owaiskazi19 (commented):
Overall looks good with a general comment of changing the exception to `FlowFrameworkException`.

src/main/java/org/opensearch/flowframework/indices/FlowFrameworkIndicesHandler.java
amitgalitz: 3 was used across the board in ml-commons. Had 5 here cause the script update was the one that happens more often, can change it back to 3

src/main/java/org/opensearch/flowframework/common/WorkflowResources.java
amitgalitz: changed to FlowFrameworkException

src/main/java/org/opensearch/flowframework/model/ResourceCreated.java
amitgalitz: changed to FlowFrameworkException

src/main/java/org/opensearch/flowframework/indices/FlowFrameworkIndicesHandler.java
amitgalitz: check that it works well with 3 also, moved both 3 to keep consistent for now

src/main/java/org/opensearch/flowframework/common/WorkflowResources.java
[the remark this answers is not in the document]
owaiskazi19: @amitgalitz why not create a global map here like we had `List<WorkflowData>` in the execute method, name it as `Map<String, ResourcesCreated>` and store the resources of a step in it as a new entry to the list.

This way with status API, we need to just iterate on the map which would have key as workflow_step_name and values as the resources created.

dbwiddis (approved):

PR #234, opened 2023-12-01
Add Util method to fetch inputs from parameters, content, and previous step output

### LINKED ISSUE #210: [FEATURE] WorkflowStep Execution Improvements
Currently, `WorkflowStep`s are each given a `List<WorkflowData>` which is populated by its own user inputs, and outputs from predecessor nodes. From this data, a `WorkflowStep` then retrieves values from this list, based on a unique field name for the expected inputs. 

There is a case in which there could be multiple entries of data with the same field name, which runs the risk of having a `WorkflowStep` consume the incorrect data.

It is necessary to modify the `WorkflowStep::execute` method to provide additional data, forgoing a `List<WorkflowData>` and instead providing a `Map<String workflowStepID, WorkflowData output>` which would allow us to provide multiple entries of the same type of data, uniquely identified by the workflow step ID that produced the output. 

Additionally, given a `Map<String workflowStepID, String fieldName>` a `WorkflowStep` can then refer to this mapping to ascertain which keys to look for in the output map and the specific field to search for within that entries `WorkflowData`. The full definition is as follows : 

```java
public CompletableFuture<WorkflowData> execute(
    String currentNodeId,
    WorkflowData currentNodeInputs,
    Map<String, WorkflowData> outputs,
    Map<String, String> previousNodeInputs
)
```

dbwiddis:
### Description

Adds a Util method to fetch required and optional inputs.  Priority is given to:
 - Output of previous input steps explicitly stated in the template
 - The current node's input parameters (from template)
 - The current node's input content (from template)
 - Any matching key in any previous step output

If the fetched value is of the form `${{ foo.bar }}` then it fetches the value of key `bar` from step `foo`.

In addition to the parsing method, I've implemented it in some of the existing steps as an example.  Some implementations break existing step tests (because missing input exceptions are different) and can be done later; doing them while awaiting review, but merging this earlier will permit it to be used for current step development.

### Issues Resolved

Fixes #210 

### DIFF
--- a/src/main/java/org/opensearch/flowframework/util/ParseUtils.java
+++ b/src/main/java/org/opensearch/flowframework/util/ParseUtils.java
@@ -8,21 +8,6 @@
  */
 package org.opensearch.flowframework.util;
 
-import static org.opensearch.core.xcontent.XContentParserUtils.ensureExpectedToken;
-import static org.opensearch.flowframework.common.CommonValue.MODEL_ID;
-import static org.opensearch.flowframework.common.CommonValue.PARAMETERS_FIELD;
-
-import java.io.IOException;
-import java.time.Instant;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.Map;
-import java.util.Map.Entry;
-import java.util.Optional;
-import java.util.Set;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
-
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.opensearch.client.Client;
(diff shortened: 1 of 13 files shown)


src/main/java/org/opensearch/flowframework/util/ParseUtils.java
```diff
@@ -228,4 +238,95 @@ public static Map<String, String> getStringToStringMap(Object map, String fieldN
...
+                .findAny();
+            if (previousNodeForKey.isPresent()) {
+                WorkflowData previousNodeOutput = outputs.get(previousNodeForKey.get());
+                if (previousNodeOutput != null) {
+                    value = previousNodeOutput.getContent().get(key);
+                }
+            }
+            // Priority 2: inputs specified in template
+            // ... fetch from currentNodeInputs (params take precedence)
+            if (value == null) {
```
joshpalis, line 285: Looking at all the workflow steps, it seems we've always retrieved values from `content` rather than `params`. Can you explain the rationale as to why the parameters take priority here if we've never used params to get values from before?

joshpalis (approved):
This is a great addition Dan, everything looks so much cleaner now. Approved, but just had one question for clarification

owaiskazi19 (approved):

src/main/java/org/opensearch/flowframework/util/ParseUtils.java
dbwiddis: Params are passed by the user on the command line and should generally override other settings. 

Someone could create a template with a default and then change something on the rest path ad hoc. 

Not sure it will ever be used but left it for completeness!

Issue #238, opened 2023-12-01
[FEATURE] Add a retry functinality for DeployModelstep and return a model_id


owaiskazi19:
### Is your feature request related to a problem?
Currently, we are returning modelId only from RegisterLocalModelStep
https://github.com/opensearch-project/flow-framework/blob/aba7dea51f3334db506a2e39d8b511cf352bf81c/src/main/java/org/opensearch/flowframework/workflow/RegisterLocalModelStep.java#L249
and RegisterRemoteModelStep
https://github.com/opensearch-project/flow-framework/blob/aba7dea51f3334db506a2e39d8b511cf352bf81c/src/main/java/org/opensearch/flowframework/workflow/RegisterRemoteModelStep.java#L75

But when fetching a modelID for other steps like RegisterAgent or Tools or any other workflow steps which requires a modelId, we should wait for the complete model to be deployed and use the modelID returned.

### What solution would you like?
This boils down to, add a retry capability for DeployModelStep and wait for the model deployment status to be COMPLETED and return a modelID in the WorkflowData.

### What alternatives have you considered?
Add a `?deploy=true` flag for RegisterModel steps.

### Do you have any additional context?
_Add any other context or screenshots about the feature request here._

owaiskazi19:
Steps to test : 

1. Enable flow framework APIs
```
curl -i -XPUT "localhost:9200/_cluster/settings" -H "Content-Type:application/json" --data '{"transient":{"plugins.flow_framework.enabled":true}}'

HTTP/1.1 200 OK
X-OpenSearch-Version: OpenSearch/3.0.0-SNAPSHOT (opensearch)
content-type: application/json; charset=UTF-8
content-length: 99

{"acknowledged":true,"persistent":{},"transient":{"plugins":{"flow_framework":{"enabled":"true"}}}}
```

2. Enable model registration via url and model deployment on non-ml nodes
```
curl -i -XPUT "localhost:9200/_cluster/settings" -H "Content-Type:application/json" --data '{"persistent":{"plugins.ml_commons.allow_registering_model_via_url":true,"plugins.ml_commons.only_run_on_ml_node":false}}'
HTTP/1.1 200 OK
X-OpenSearch-Version: OpenSearch/3.0.0-SNAPSHOT (opensearch)
content-type: application/json; charset=UTF-8
content-length: 149

[...]

PR #245, opened 2023-12-04
[Feature/agent_framework] Added Retry functionality for Deploy Model

owaiskazi19:
### Description
1. Added Retry functionality for Deploy Model
2. Pulled the retryTask method in abstract class
3. Increased node timeout as it takes around 15s to deploy a model and fetch the status
4. Added tests

Steps to test: https://github.com/opensearch-project/flow-framework/issues/238#issuecomment-1837673453

### Issues Resolved
Closes https://github.com/opensearch-project/flow-framework/issues/238

### DIFF
--- a/src/main/java/org/opensearch/flowframework/model/WorkflowNode.java
+++ b/src/main/java/org/opensearch/flowframework/model/WorkflowNode.java
@@ -47,7 +47,7 @@ public class WorkflowNode implements ToXContentObject {
     /** The field defining the timeout value for this node */
     public static final String NODE_TIMEOUT_FIELD = "node_timeout";
     /** The default timeout value if the template doesn't override it */
-    public static final String NODE_TIMEOUT_DEFAULT_VALUE = "10s";
+    public static final String NODE_TIMEOUT_DEFAULT_VALUE = "15s";
 
     private final String id; // unique id
     private final String type; // maps to a WorkflowStep
--- a/src/main/java/org/opensearch/flowframework/workflow/AbstractRetryableWorkflowStep.java
+++ b/src/main/java/org/opensearch/flowframework/workflow/AbstractRetryableWorkflowStep.java
@@ -8,27 +8,131 @@
  */
 package org.opensearch.flowframework.workflow;
 
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.opensearch.ExceptionsHelper;
 import org.opensearch.cluster.service.ClusterService;
 import org.opensearch.common.settings.Settings;
+import org.opensearch.common.util.concurrent.FutureUtils;
+import org.opensearch.core.action.ActionListener;
+import org.opensearch.core.rest.RestStatus;
+import org.opensearch.flowframework.common.WorkflowResources;
+import org.opensearch.flowframework.exception.FlowFrameworkException;
+import org.opensearch.flowframework.indices.FlowFrameworkIndicesHandler;
+import org.opensearch.ml.client.MachineLearningNodeClient;
+import org.opensearch.ml.common.MLTaskState;
 
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+import java.util.stream.Stream;
+
+import static org.opensearch.flowframework.common.CommonValue.REGISTER_MODEL_STATUS;
 import static org.opensearch.flowframework.common.FlowFrameworkSettings.MAX_GET_TASK_REQUEST_RETRY;
 
 /**
  * Abstract retryable workflow step
  */
 public abstract class AbstractRetryableWorkflowStep implements WorkflowStep {
-
+    private static final Logger logger = LogManager.getLogger(AbstractRetryableWorkflowStep.class);
     /** The maximum number of transport request retries */
     protected volatile Integer maxRetry;
+    private final MachineLearningNodeClient mlClient;
+    private final FlowFrameworkIndicesHandler flowFrameworkIndicesHandler;
 
     /**
      * Instantiates a new Retryable workflow step
      * @param settings Environment settings
      * @param clusterService the cluster service
      */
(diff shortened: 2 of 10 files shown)


src/main/java/org/opensearch/flowframework/workflow/AbstractRetryableWorkflowStep.java
```diff
@@ -8,27 +8,133 @@
...
+            MLTaskState currentState = response.getState();
+            if (currentState != MLTaskState.COMPLETED) {
+                if (Stream.of(MLTaskState.FAILED, MLTaskState.COMPLETED_WITH_ERROR).anyMatch(x -> x == currentState)) {
+                    // Model registration failed or completed with errors
+                    String errorMessage = message + " failed with error : " + response.getError();
+                    logger.error(errorMessage);
+                    future.completeExceptionally(new FlowFrameworkException(errorMessage, RestStatus.BAD_REQUEST));
+                } else {
+                    // Task still in progress, attempt retry
+                    throw new IllegalStateException(message + " is not yet completed");
```
dbwiddis, line 87: Avoid using exceptions for normal process flow.  They're a serious performance hit. https://dzone.com/articles/exceptions-as-controlflow-in-java

Why not just do the retry recursion here instead of in a catch block?

I recognize that regular recursion impacts the stack but if you add a `() ->` in the return you can basically achieve tail recursion and avoid it.  See https://github.com/jonckvanderkogel/java-tail-recursion/blob/master/src/main/java/tailrecursion/Factorial.java

But better than that, why not just do this all in a loop?   while... not COMPLETE and tries < retries...

src/test/java/org/opensearch/flowframework/workflow/RegisterLocalModelStepTests.java
```diff
@@ -156,7 +156,7 @@ public void testRegisterLocalModelSuccess() throws Exception {
         verify(machineLearningNodeClient, times(1)).getTask(any(), any());
 
         assertTrue(future.isDone());
-        assertTrue(!future.isCompletedExceptionally());
+        assertFalse(future.isCompletedExceptionally());
```
dbwiddis, line 159: Thank you.

dbwiddis (commented):
FIrst pass

src/main/java/org/opensearch/flowframework/workflow/AbstractRetryableWorkflowStep.java
```diff
@@ -8,27 +8,133 @@
...
+        String taskId,
+        int retries,
+        String message
+    ) {
+        mlClient.getTask(taskId, ActionListener.wrap(response -> {
+            MLTaskState currentState = response.getState();
+            if (currentState != MLTaskState.COMPLETED) {
+                if (Stream.of(MLTaskState.FAILED, MLTaskState.COMPLETED_WITH_ERROR).anyMatch(x -> x == currentState)) {
+                    // Model registration failed or completed with errors
+                    String errorMessage = message + " failed with error : " + response.getError();
```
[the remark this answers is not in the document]
owaiskazi19: How about `workflowStep`?

src/main/java/org/opensearch/flowframework/model/WorkflowNode.java
```diff
@@ -47,7 +47,7 @@ public class WorkflowNode implements ToXContentObject {
     /** The field defining the timeout value for this node */
     public static final String NODE_TIMEOUT_FIELD = "node_timeout";
     /** The default timeout value if the template doesn't override it */
-    public static final String NODE_TIMEOUT_DEFAULT_VALUE = "10s";
+    public static final String NODE_TIMEOUT_DEFAULT_VALUE = "15s";
```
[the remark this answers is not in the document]
amitgalitz: Should we have the same timeout for each step or any differences when steps are supposed to take a lot longer (for example local model is an async API that was made async specifically cause it could take 10x longer then remote model)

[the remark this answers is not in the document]
owaiskazi19: It can be different for workflow steps depending on their response time.  Created https://github.com/opensearch-project/flow-framework/issues/249

src/main/java/org/opensearch/flowframework/workflow/AbstractRetryableWorkflowStep.java
owaiskazi19: @joshpalis has this method. I pulled it in the AbstractClass. @joshpalis WDYT?

src/main/java/org/opensearch/flowframework/workflow/AbstractRetryableWorkflowStep.java
```diff
@@ -8,27 +8,133 @@
...
+                    logger.error(errorMessage);
+                    future.completeExceptionally(new FlowFrameworkException(errorMessage, RestStatus.BAD_REQUEST));
+                } else {
+                    // Task still in progress, attempt retry
+                    throw new IllegalStateException(workflowStep + " is not yet completed");
+                }
+            } else {
+                try {
+                    logger.info(workflowStep + " successful for {} and modelId {}", workflowId, response.getModelId());
+                    String resourceName = WorkflowResources.getResourceByWorkflowStep(getName());
```
amitgalitz, line 93: would the getName() here return the step it was called from or a `get_task` step

owaiskazi19: Yes. From the enum we have in `WorkflowResources`.

amitgalitz (approved):

src/main/java/org/opensearch/flowframework/workflow/AbstractRetryableWorkflowStep.java
joshpalis: No strong opinion, we can modify this retry function to just do everything in a loop

joshpalis (approved):

Issue #89, opened 2023-10-12
[META] Implement basic manual Undo/Rollback capability


dbwiddis:
https://github.com/opensearch-project/opensearch-ai-flow-framework/issues/32 describes an overall vision for a full rollback or overwrite capability, and near the bottom says:
> ### What will it take to execute?
> 
> In the near term we should keep track of resources created in order to enable manual removal, regardless of any more automatic behavior.

Which means steps 1 and 3 of:

> We need a capability to:
> 
> 1. Track what resources have been created in a workflow execution
> 2. Automate the removal of these resources (if possible)
> 3. Provide users with a specific list of resources created to enable manual removal in case automated removal fails, or they want more control over the process.

This issue is intended to do exactly that near term requirement.  Review comments on #32 for some more context.

 - [x] Store all resources created in a state index.  https://github.com/opensearch-project/opensearch-ai-flow-framework/issues/74
 - [x] Update Template execution APIs to ensure that this list of resources created is returned to the user as part of the successful response.
 - [x] Update Template execution APIs to ensure that this list of resources created is returned to the user as part of a failed response.
 - [x] Provide a REST API that allows a user to list the resources they created.  This will probably use the same code path which was used in the response generation in the previous steps.
 - [ ] Provide a REST API that allows a user to delete the resources they created.  Look into tagging the resources in such a way that they can be easily filtered/identified.  Consider integrating this as an optional parameter to the delete template API.

PR #246, opened 2023-12-04
[Feature/agent_framework] Add Delete Agent Step

dbwiddis:
### Description

Adds a workflow step to delete an agent (opposite of register).

### Issues Resolved

Part of #89

### DIFF
--- a/src/main/java/org/opensearch/flowframework/workflow/DeleteAgentStep.java
+++ b/src/main/java/org/opensearch/flowframework/workflow/DeleteAgentStep.java
@@ -1,0 +1,100 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+package org.opensearch.flowframework.workflow;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.opensearch.ExceptionsHelper;
+import org.opensearch.action.delete.DeleteResponse;
+import org.opensearch.core.action.ActionListener;
+import org.opensearch.flowframework.exception.FlowFrameworkException;
+import org.opensearch.flowframework.util.ParseUtils;
+import org.opensearch.ml.client.MachineLearningNodeClient;
+
+import java.io.IOException;
+import java.util.Collections;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.CompletableFuture;
+
+import static org.opensearch.flowframework.common.CommonValue.AGENT_ID;
+
+/**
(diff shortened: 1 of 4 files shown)


src/main/java/org/opensearch/flowframework/workflow/DeleteAgentStep.java
```diff
@@ -0,0 +1,100 @@
...
+            Map<String, Object> inputs = ParseUtils.getInputsFromPreviousSteps(
+                requiredKeys,
+                optionalKeys,
+                currentNodeInputs,
+                outputs,
+                previousNodeInputs
+            );
+            String agentId = (String) inputs.get(AGENT_ID);
+
+            mlClient.deleteConnector(agentId, actionListener); // TODO s/Connector/Agent/
```
owaiskazi19, line 89: Don't we have a delete agent method for mlClient?

owaiskazi19: Just saw the PR description. Ignore the comment.

owaiskazi19:
You might need to add the step to the [enum](https://github.com/opensearch-project/flow-framework/blob/feature/agent_framework/src/main/java/org/opensearch/flowframework/common/WorkflowResources.java#L23) as well.

dbwiddis:
> You might need to add the step to the [enum](https://github.com/opensearch-project/flow-framework/blob/feature/agent_framework/src/main/java/org/opensearch/flowframework/common/WorkflowResources.java#L23) as well.

The register agent resource is already there, this would delete the same resource (agent id)

owaiskazi19 (approved):

src/main/java/org/opensearch/flowframework/workflow/DeleteAgentStep.java
```diff
@@ -0,0 +1,100 @@
...
+    ) throws IOException {
+        CompletableFuture<WorkflowData> deleteAgentFuture = new CompletableFuture<>();
+
+        ActionListener<DeleteResponse> actionListener = new ActionListener<>() {
+
+            @Override
+            public void onResponse(DeleteResponse deleteResponse) {
+                deleteAgentFuture.complete(
+                    new WorkflowData(
+                        Map.ofEntries(Map.entry("agent_id", deleteResponse.getId())),
```
amitgalitz, line 62: should we use `WorkflowResources.getResourceByWorkflowStep("register_agent")` or still add it to the enum just for the fetching constant process. Just in terms of less places we type even something like `"agent_id"`

dbwiddis: Sure, but we should do the same for delete connector, delete model, and unload model.

This class is a literal copy/paste of the other two delete classes, whatever we change here should match there.  Can we go back in another PR to refactor all of these?

amitgalitz (approved):

Issue #61, opened 2023-09-28
[FEATURE] Better Thread Pool integration


dbwiddis:
### Is your feature request related to a problem?

#47 implemented basic use of an ExecutorService to give better control of the async process execution.
#66 improved on #47 by using the actual thread pool but is still just using the generic name.

We can implement more custom thread pool behavior/prioritization similar to how other plugins have done so, moving beyond the generic thread pool.

### What solution would you like?

Probably carefully choose the correct [existing thread pool](https://github.com/opensearch-project/OpenSearch/blob/e1565821a28109c5c7f773e77764fc5104de7f88/server/src/main/java/org/opensearch/threadpool/ThreadPool.java#L94) for the threads you're creating.

Possibly a custom thread pool integration properly prioritizing this plugin's threads. Custom thread pools override [`getExecutorBuilders()`](https://github.com/opensearch-project/OpenSearch/blob/e1565821a28109c5c7f773e77764fc5104de7f88/server/src/main/java/org/opensearch/plugins/Plugin.java#L235), see examples [here](https://github.com/opensearch-project/OpenSearch/blob/e1565821a28109c5c7f773e77764fc5104de7f88/plugins/repository-s3/src/main/java/org/opensearch/repositories/s3/S3RepositoryPlugin.java#L94-L105) and [here](https://github.com/opensearch-project/OpenSearch/blob/e1565821a28109c5c7f773e77764fc5104de7f88/plugins/repository-azure/src/main/java/org/opensearch/repositories/azure/AzureRepositoryPlugin.java#L110C1-L116)

### Do you have any additional context?

This is an admittedly vague feature because it's an optional prioritization, entered primarily to note that the initial implementation took the lowest/default priority in the interest of unblocking rapid development. This issue is intended to prompt a more careful implementation.

austintlee:
https://github.com/opensearch-project/OpenSearch/issues/10248 - none of this stuff needs to run in transport threads?

dbwiddis:
> [opensearch-project/OpenSearch#10248](https://github.com/opensearch-project/OpenSearch/issues/10248) - none of this stuff needs to run in transport threads?

We're calling the workflow steps via the MLClient, so the eventual calls will run on their appropriate threads.  All of our calls have timeout (unless the user explicitly overrides.)

[...]

PR #265, opened 2023-12-09
Change thread queue to 100 and fix headers parsing bug

amitgalitz:
### Description
I changed the provision thread queue to 100. This is because currently the size queue of 10 is too low. Basically right now if all the threads are busy and If an 11th request comes, the request will be just lost completely if we have no extra queue or re-route logic of our own. Currently we have no extra re-route logic but I think expanding the queue to a size of 100 to allow more provisioning should be sufficient for now. We should probably come back to this after doing more intense load testing.

I also additionally added a StringToObjectMap parser and builder so we can handle cases where we encounter an Object[] that these objects can be either be Map<String, String> or Map<String, Map<String, String>> like in the case of create connector actions

```
    "actions": [
        {
            "action_type": "predict",
            "method": "POST",
            "url": "https://${parameters.endpoint}/v1/chat/completions",
            "headers": {
                "Authorization": "Bearer ${credential.openAI_key}"
            },
            "request_body": "{ \"model\": \"${parameters.model}\", \"messages\": ${parameters.messages} }"
        }
    ]
 ```
           
I tested manually through API calls, however I didn't add a UT for the StringToObject because I didn't want to use reflection to check the object type here, since its a generic object.             
           
           
### Issues Resolved

### DIFF
--- a/src/main/java/org/opensearch/flowframework/FlowFrameworkPlugin.java
+++ b/src/main/java/org/opensearch/flowframework/FlowFrameworkPlugin.java
@@ -158,7 +158,7 @@ public class FlowFrameworkPlugin extends Plugin implements ActionPlugin {
                 settings,
                 PROVISION_THREAD_POOL,
                 OpenSearchExecutors.allocatedProcessors(settings),
-                10,
+                100,
                 FLOW_FRAMEWORK_THREAD_POOL_PREFIX + PROVISION_THREAD_POOL
             )
         );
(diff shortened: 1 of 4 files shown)


dbwiddis:
>  the request will be just lost completely if we have no extra queue or re-route logic of our own. 

What does this mean? Does it fail silently?  I thought an ExecutorService just blocked until a thread was available, in which case these should eventually time out, right?  Or am I missing something?

dbwiddis (approved):

joshpalis:
Attaching this related issue : https://github.com/opensearch-project/flow-framework/issues/61

joshpalis (approved):

PR #268, opened 2023-12-09
Fix max workflows setting test

dbwiddis:
### Description

Fixes a typo where the setting wasn't actually being checked and was using the default.

Also fixed some javadoc warnings


amitgalitz (approved):

amitgalitz:
btw if we merge this to main, it will mean it will be slightly out of order when we merge things back from feature branch. We might have already broken the order in other places though

dbwiddis:
> it will mean it will be slightly out of order when we merge things back from feature branch. We might have already broken the order in other places though

Ordering doesn't matter much, it's whether there are conflicting changes to the same files.  One file hasn't been touched on the feature branch, one has changes only from backporting something from main, and the last has only overlapping identical changes in imports (which won't conflict) and changed lines of code that won't conflict.

PR #275, opened 2023-12-12
Update resources_created with deploy model:

### LINKED ISSUE #272: [BUG] Deploying a model doesn't update the workflow state
### What is the bug?

The workflow status API does not indicate whether a model has been deployed, preventing it from being undeployed during deprovisioning.

1. Use the provision API to provision a workflow including the "deploy_model" step.
2. Use the status API to get deployed resources.
3. Observe no entry for the deployed model.

An entry for the "deploy_model" resource in the status API result.

This is a blocker for #271

amitgalitz:
### Description

Added update to include deploy model in resources created list. This is needed for de-provisioning to work correctly. Future enhancement could be made if confusion is stated by users on both deploy and register model being under resources created. 

Example response:

```
{
    "workflow_id": "G513W4wBRItvelfmy_w4",
    "state": "COMPLETED",
    "provisioning_progress": "DONE",
    "provision_start_time": 1702341633430,
    "provision_end_time": 1702341639045,
    "resources_created": [
        {
            "workflow_step_name": "register_agent",
            "workflow_step_id": "sub_agent",
            "agent_id": "HJ13W4wBRItvelfm5fyw"
        },
        {
            "workflow_step_id": "openAI_connector",
            "connector_id": "HZ13W4wBRItvelfm5vzf"
        },
        {
            "workflow_step_id": "gpt-3.5-model",
            "model_id": "H513W4wBRItvelfm5_xS"
        },
        {
            "workflow_step_name": "deploy_model",
            "workflow_step_id": "deployed-gpt-3.5-model",
            "model_id": "H513W4wBRItvelfm5_xS"
        },
        {
            "workflow_step_name": "register_agent",
            "workflow_step_id": "root_agent",
            "agent_id": "IZ13W4wBRItvelfm-_xt"
        }
    ]
}
```

### Issues Resolved
resolves #272 

### DIFF
--- a/src/main/java/org/opensearch/flowframework/workflow/AbstractRetryableWorkflowStep.java
+++ b/src/main/java/org/opensearch/flowframework/workflow/AbstractRetryableWorkflowStep.java
@@ -20,7 +20,6 @@ import org.opensearch.flowframework.common.WorkflowResources;
 import org.opensearch.flowframework.exception.FlowFrameworkException;
 import org.opensearch.flowframework.indices.FlowFrameworkIndicesHandler;
 import org.opensearch.ml.client.MachineLearningNodeClient;
-import org.opensearch.ml.common.MLTask;
 import org.opensearch.ml.common.MLTaskState;
 
 import java.util.Map;
@@ -60,24 +59,6 @@ public abstract class AbstractRetryableWorkflowStep implements WorkflowStep {
     }
 
     /**
-     * Completes the future for either deploy or register local model step
-     * @param resourceName resource name for the given step
-     * @param nodeId node ID of the given step
-     * @param workflowId workflow ID of the given workflow
-     * @param response Response from ml commons get Task API
-     * @param future CompletableFuture of the given step
-     */
-    public void completeFuture(String resourceName, String nodeId, String workflowId, MLTask response, CompletableFuture future) {
-        future.complete(
-            new WorkflowData(
-                Map.ofEntries(Map.entry(resourceName, response.getModelId()), Map.entry(REGISTER_MODEL_STATUS, response.getState().name())),
-                workflowId,
-                nodeId
-            )
-        );
-    }
-
-    /**
      * Retryable get ml task
      * @param workflowId the workflow id
      * @param nodeId the workflow node id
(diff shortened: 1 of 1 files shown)


src/main/java/org/opensearch/flowframework/workflow/AbstractRetryableWorkflowStep.java
```diff
@@ -110,25 +91,36 @@ void retryableGetMlTask(
                 try {
                     logger.info(workflowStep + " successful for {} and modelId {}", workflowId, response.getModelId());
                     String resourceName = WorkflowResources.getResourceByWorkflowStep(getName());
+                    String id;
                     if (getName().equals(WorkflowResources.DEPLOY_MODEL.getWorkflowStep())) {
-                        completeFuture(resourceName, nodeId, workflowId, response, future);
+                        id = response.getModelId();
```
dbwiddis, line 96: By the time we get here, MLTaskState is COMPLETED in both register and deploy model.  So both have a model ID.  Why are we treating the task Id differently for them?  Shouldn't it be `modelId` for both of them?

Interestingly the existing code uses `response.getTaskId()` for the resource ID, but the output is clearly the model Id.

Is taskid = modelid for an initial model registration?

src/main/java/org/opensearch/flowframework/workflow/AbstractRetryableWorkflowStep.java
```diff
@@ -110,25 +91,36 @@ void retryableGetMlTask(
...
+                        nodeId,
+                        getName(),
+                        id,
+                        ActionListener.wrap(updateResponse -> {
+                            logger.info("successfully updated resources created in state index: {}", updateResponse.getIndex());
+                            future.complete(
+                                new WorkflowData(
+                                    Map.ofEntries(
+                                        Map.entry(resourceName, response.getModelId()),
+                                        Map.entry(REGISTER_MODEL_STATUS, response.getState().name())
```
dbwiddis, line 111: Not sure we need "register model status" for a deployed model, but doesn't hurt to have extra field in the output.

dbwiddis (approved):
Approved to get this merged and unblock me. however I have two questions for educational/future-PR reference.

src/main/java/org/opensearch/flowframework/workflow/AbstractRetryableWorkflowStep.java
joshpalis: > Is taskid = modelid for an initial model registration?

It is not the same, a completed task will give us a model ID, the task ID is merely the document ID of the ML Task index.

src/main/java/org/opensearch/flowframework/workflow/AbstractRetryableWorkflowStep.java
```diff
@@ -110,25 +91,36 @@ void retryableGetMlTask(
                 try {
                     logger.info(workflowStep + " successful for {} and modelId {}", workflowId, response.getModelId());
                     String resourceName = WorkflowResources.getResourceByWorkflowStep(getName());
+                    String id;
```
joshpalis, line 94: This logic should ideally be in the Workflow Step classes themselves, rather than this abstract class, since we're dealing with outputs specific to those Workflow Step classes. If I recall correctly, this abstract class is just here to provide the setting which limits the maximum amount of times we can retry a request

joshpalis (approved):
Approved to unblock @dbwiddis

src/main/java/org/opensearch/flowframework/workflow/AbstractRetryableWorkflowStep.java
amitgalitz: here we are updating just the workflowData and not the index, so just thought on the same line that it couldn't hurt to add this.

src/main/java/org/opensearch/flowframework/workflow/AbstractRetryableWorkflowStep.java
amitgalitz: I'll remove this and only use modelId in follow up PR, do you mean then its okay to keep the status update call here in that case or move it regardless?

src/main/java/org/opensearch/flowframework/workflow/AbstractRetryableWorkflowStep.java
amitgalitz: will change to both use modelId in follow up PR

joshpalis: Talked with @amitgalitz offline, we've combined the `RegisterLocalModelStep` and `GetMLTaskStep` to make the call effectively synchronous within our framework. The task ID is returned by the local model registration API call, and we use this to ascertain the status of the model deployment, which gives us the model ID once it is in a completed state. The output of this RegisterLocalModelStep should just be the `model_id`, the task id is just used internally within the step

src/main/java/org/opensearch/flowframework/workflow/AbstractRetryableWorkflowStep.java
dbwiddis: I agree with @joshpalis that ideally an abstract class shouldn't have any specifics to a class.  Good for later refactoring.

Issue #259, opened 2023-12-07
[FEATURE] Combine create and provision APIs


dbwiddis:
> If we add validation by default as a prerequisite to saving a WIP template, that could be a frustrating experience for a builder. In that regard, I think it should be a separate capability and having an optional dry-run param for the `RestCreateWorkflowAction` to validate prior to saving makes sense.
> 
> I think validation goes more with the "execution" portion, but we should be able to do the validation before saving.  Which means we sort of need a "save-and-execute" single API call.
> 
> Which brings back up the possibility of combining APIs.
> 
> We have
> ```
> POST _plugins/_flow_framework/workflow
> {
>     // use case template
> }
> ```
> followed by
> ```
> POST _plugins/_flow_framework/workflow/<workflow_id>/_provision
> {}
> ```
> This works well for a "save WIP and finally execute when done" but in the case we have a completed/debugged template that we always want to execute, we could do that all-in-one with:
> ```
> POST _plugins/_flow_framework/workflow/_provision
> {
>     // use case template
> }
> ```
> Now, throw a validation param in there and we have a one-liner just like we want:
> ```
> POST _plugins/_flow_framework/workflow/_provision?dryrun=true
> {
>     // use case template
> }
> ```
> (or similar)

_Originally posted by @dbwiddis in https://github.com/opensearch-project/flow-framework/issues/88#issuecomment-1765403845_

dbwiddis:
It's also possible to add `provision=true` as a param to the existing create API rather than having a new API.

joshpalis:
> It's also possible to add provision=true as a param to the existing create API rather than having a new API.

+1 to this approach :
```
// Request
POST _plugins/_flow_framework/workflow?provision=true 
{
    // use case template
}

// Response
{"workflow_id":"<workflowId>"}
```

From what I can gather, the sequence would be as follows : 
1. User invokes the above API with `provision=true`
2. This hits our `RestCreateWorkflowAction`, which passes the request to the `CreateWorkflowTransportAction`
3. The `CreateWorkflowTransportAction` indexes the template into the Global Context index and responds to the action listener with a `workflow_id`
4. Once we get the response back in the `ResCreateWorkflowAction`, we check if `provision` param is true and then invoke `ProvisionWokflowTransportAction`, which will handle validation/executing the workflow
5. The `ProvisionWorkflowTrasportAction` will respond with a `workflow_id` or an error if validation/execution fails 

CC : @jackiehanyang

PR #282, opened 2023-12-12
Combine create api with provision api by adding a provision param

jackiehanyang:
### Description
Combine create api with provision api by adding a provision=true parameter to create api

### Issues Resolved
https://github.com/opensearch-project/flow-framework/issues/259

### DIFF
--- a/src/main/java/org/opensearch/flowframework/rest/RestCreateWorkflowAction.java
+++ b/src/main/java/org/opensearch/flowframework/rest/RestCreateWorkflowAction.java
@@ -31,6 +31,7 @@ import java.util.List;
 import java.util.Locale;
 
 import static org.opensearch.flowframework.common.CommonValue.DRY_RUN;
+import static org.opensearch.flowframework.common.CommonValue.PROVISION_WORKFLOW;
 import static org.opensearch.flowframework.common.CommonValue.WORKFLOW_ID;
 import static org.opensearch.flowframework.common.CommonValue.WORKFLOW_URI;
 import static org.opensearch.flowframework.common.FlowFrameworkSettings.FLOW_FRAMEWORK_ENABLED;
@@ -90,9 +91,9 @@ public class RestCreateWorkflowAction extends AbstractWorkflowAction {
 
             String workflowId = request.param(WORKFLOW_ID);
             Template template = Template.parse(request.content().utf8ToString());
-            boolean dryRun = request.paramAsBoolean(DRY_RUN, false);
+            boolean provision = request.paramAsBoolean(PROVISION_WORKFLOW, false);
 
-            WorkflowRequest workflowRequest = new WorkflowRequest(workflowId, template, dryRun, requestTimeout, maxWorkflows);
+            WorkflowRequest workflowRequest = new WorkflowRequest(workflowId, template, provision, requestTimeout, maxWorkflows);
 
             return channel -> client.execute(CreateWorkflowAction.INSTANCE, workflowRequest, ActionListener.wrap(response -> {
                 XContentBuilder builder = response.toXContent(channel.newBuilder(), ToXContent.EMPTY_PARAMS);
--- a/src/main/java/org/opensearch/flowframework/rest/RestCreateWorkflowAction.java
+++ b/src/main/java/org/opensearch/flowframework/rest/RestCreateWorkflowAction.java
@@ -30,6 +30,7 @@ import java.io.IOException;
 import java.util.List;
 import java.util.Locale;
 
+import static org.opensearch.flowframework.common.CommonValue.DRY_RUN;
 import static org.opensearch.flowframework.common.CommonValue.PROVISION_WORKFLOW;
 import static org.opensearch.flowframework.common.CommonValue.WORKFLOW_ID;
 import static org.opensearch.flowframework.common.CommonValue.WORKFLOW_URI;
@@ -90,9 +91,10 @@ public class RestCreateWorkflowAction extends AbstractWorkflowAction {
 
             String workflowId = request.param(WORKFLOW_ID);
             Template template = Template.parse(request.content().utf8ToString());
+            boolean dryRun = request.paramAsBoolean(DRY_RUN, false);
             boolean provision = request.paramAsBoolean(PROVISION_WORKFLOW, false);
 
-            WorkflowRequest workflowRequest = new WorkflowRequest(workflowId, template, provision, requestTimeout, maxWorkflows);
+            WorkflowRequest workflowRequest = new WorkflowRequest(workflowId, template, dryRun, provision, requestTimeout, maxWorkflows);
 
             return channel -> client.execute(CreateWorkflowAction.INSTANCE, workflowRequest, ActionListener.wrap(response -> {
                 XContentBuilder builder = response.toXContent(channel.newBuilder(), ToXContent.EMPTY_PARAMS);
--- a/src/main/java/org/opensearch/flowframework/rest/RestCreateWorkflowAction.java
+++ b/src/main/java/org/opensearch/flowframework/rest/RestCreateWorkflowAction.java
@@ -30,7 +30,6 @@ import java.io.IOException;
 import java.util.List;
 import java.util.Locale;
 
-import static org.opensearch.flowframework.common.CommonValue.DRY_RUN;
 import static org.opensearch.flowframework.common.CommonValue.PROVISION_WORKFLOW;
 import static org.opensearch.flowframework.common.CommonValue.WORKFLOW_ID;
 import static org.opensearch.flowframework.common.CommonValue.WORKFLOW_URI;
--- a/src/main/java/org/opensearch/flowframework/transport/CreateWorkflowTransportAction.java
+++ b/src/main/java/org/opensearch/flowframework/transport/CreateWorkflowTransportAction.java
@@ -28,12 +28,16 @@ import org.opensearch.flowframework.indices.FlowFrameworkIndicesHandler;
 import org.opensearch.flowframework.model.ProvisioningProgress;
 import org.opensearch.flowframework.model.State;
 import org.opensearch.flowframework.model.Template;
+import org.opensearch.flowframework.model.Workflow;
+import org.opensearch.flowframework.workflow.ProcessNode;
 import org.opensearch.flowframework.workflow.WorkflowProcessSorter;
 import org.opensearch.index.query.QueryBuilder;
 import org.opensearch.index.query.QueryBuilders;
 import org.opensearch.search.builder.SearchSourceBuilder;
 import org.opensearch.tasks.Task;
 import org.opensearch.transport.TransportService;
+
+import java.util.List;
 
 import static org.opensearch.flowframework.common.CommonValue.PROVISIONING_PROGRESS_FIELD;
 import static org.opensearch.flowframework.common.CommonValue.STATE_FIELD;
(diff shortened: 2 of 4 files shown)


amitgalitz:
Do we not have a dry_run optional param at all anymore?

jackiehanyang:
> Do we not have a dry_run optional param at all anymore?

What's the benefit of keeping the dry_run parameter? provision covers all the check dry_run does. Also as this project is not launched yet, this is not a breaking change.

joshpalis:
> What's the benefit of keeping the dry_run parameter? provision covers all the check dry_run does. Also as this project is not launched yet, this is not a breaking change.

Dry run param in the create workflow API should remain unchanged. This is to allow for validation prior to saving into the global context index. Combining the two APIs together should just add a new path rather than supersede existing ones.

Though provision does validate as well, since we're combining the two it would make sense to validate on create (before we index) rather than after. I would suggest we invoke the `CreateWorkflowTransportAction` with a dry-run flag set to true and then provision via `ProvisionWorkflowTransportAction`

jackiehanyang:
> > What's the benefit of keeping the dry_run parameter? provision covers all the check dry_run does. Also as this project is not launched yet, this is not a breaking change.
> 
> Dry run param in the create workflow API should remain unchanged. This is to allow for validation prior to saving into the global context index. Combining the two APIs together should just add a new path rather than supersede existing ones.
> 
> Though provision does validate as well, since we're combining the two it would make sense to validate on create (before we index) rather than after. I would suggest we invoke the `CreateWorkflowTransportAction` with a dry-run flag set to true and then provision via `ProvisionWorkflowTransportAction`

Are you suggesting when `provision=true` we treat it like `dryrun=true && provision=true`?

joshpalis:
> Are you suggesting when provision=true we treat it like dryrun=true && provision=true?
Yes

src/main/java/org/opensearch/flowframework/rest/RestCreateWorkflowAction.java
```diff
@@ -90,9 +90,9 @@ protected RestChannelConsumer prepareRequest(RestRequest request, NodeClient cli
 
             String workflowId = request.param(WORKFLOW_ID);
             Template template = Template.parse(request.content().utf8ToString());
-            boolean dryRun = request.paramAsBoolean(DRY_RUN, false);
+            boolean provision = request.paramAsBoolean(PROVISION_WORKFLOW, false);
```
joshpalis, line 93: Can we keep both `dryRun` and `provision`?

if `dryrun=true`, then we validate and then index into the global context,

if `provision=true`, then we validate and then index into the global context, and then provision

dbwiddis: I agree we should keep both but we should not allow both provision and dryRun to be true at the same time.

src/main/java/org/opensearch/flowframework/transport/WorkflowRequest.java
```diff
@@ -105,7 +110,7 @@ public WorkflowRequest(StreamInput in) throws IOException {
         this.workflowId = in.readOptionalString();
         String templateJson = in.readOptionalString();
         this.template = templateJson == null ? null : Template.parse(templateJson);
-        this.dryRun = in.readBoolean();
+        this.provision = in.readBoolean();
```
dbwiddis, line 113: This needs a corresponding change in `writeTo()`

src/test/java/org/opensearch/flowframework/transport/CreateWorkflowTransportActionTests.java
```diff
@@ -132,7 +144,16 @@ public void setUp() throws Exception {
...
     }
 
-    public void testFailedDryRunValidation() {
+    public void testDryRunValidation_Success() {
+        Template validTemplate = generateValidTemplate();
+
+        @SuppressWarnings("unchecked")
+        ActionListener<WorkflowResponse> listener = mock(ActionListener.class);
+        WorkflowRequest createNewWorkflow = new WorkflowRequest(null, validTemplate, true, null, null);
+        createWorkflowTransportAction.doExecute(mock(Task.class), createNewWorkflow, listener);
```
dbwiddis, line 153: You should be able to capture the listener argument and verify onResponse was called

dbwiddis:
> > Are you suggesting when provision=true we treat it like dryrun=true && provision=true?
> 
> Yes

I may be misunderstanding dryrun or maybe it's poorly named.

Dry run usually means do validation checks but don't actually index anything.  So my thoughts would be:
 - dryRun false, provision false = validates and saves to index, does not provision
 - dryRun false, provision true = validates and saves to index, then provisions
 - dryRun true, provision false = validates but doesn't save anything
 - dryRun true, provision true = illegal

Does it mean something else than that?

joshpalis:
> Dry run usually means do validation checks but don't actually index anything. 

@dbwiddis currently when we dry-run, if the template passes validation, then we index, if not then we return an error

jackiehanyang:
> > > Are you suggesting when provision=true we treat it like dryrun=true && provision=true?
> > 
> > 
> > Yes
> 
> I may be misunderstanding dryrun or maybe it's poorly named.
> 
> Dry run usually means do validation checks but don't actually index anything. So my thoughts would be:
> 
> * dryRun false, provision false = validates and saves to index, does not provision <-- normal 2-step case front-end will use
> * dryRun false, provision true = validates and saves to index, then provisions <-- normal use case for a debugged workflow for one-click
> * dryRun true, provision false = validates but doesn't save anything <-- kinda like precommit checks :)
> * dryRun true, provision true = illegal
> 
> Does it mean something else than that?

My understanding of all the conditions is:
 - `dryRun=false && provision=false`, no formate validation, no provision. Just save the template to index
 - `dryRun=false && provision=true`, save the template to index, then provisioning the template. Provision could failed due to template formatting error like missing certain input files, or could due to template is a cyclic graph
 - `dryRun=true && provision = false`, do formate validation first, only save the template to index if it passes the check. 
 - `dryRun=true && provision = true`, do formate validation first, save the template to index if it passes the check, then provisioning the template. Although provision step also checks for formatting issues, which is already covered by dryRun check.
 
I think no matter how customer mix&match these two parameters, they should all be legal. The `dryRun=true && provision = true` case helps customer to fail early is there's any formatting issue

src/test/java/org/opensearch/flowframework/transport/CreateWorkflowTransportActionTests.java
jackiehanyang: This test case primarily tests the logic in the validateWorkflows method, which occurs before any ActionListener is involved.

joshpalis:
> My understanding of all the conditions is:
> 
> dryRun=false && provision=false, no formate validation, no provision. Just save the template to index
> dryRun=false && provision=true, save the template to index, then provisioning the template. Provision could failed due to template formatting error like missing certain input files, or could due to template is a cyclic graph. (it just performs all the check that come with the Provision API)
> dryRun=true && provision = false, do formate validation first, only save the template to index if it passes the check.
> dryRun=true && provision = true, do formate validation first, save the template to index if it passes the check, then provisioning the template. Although provision step also checks for formatting issues, which is already covered by dryRun check.
> I think no matter how customer mix&match these two parameters, they should all be legal. The dryRun=true && provision = true case helps customer to fail early is there's any formatting issue
> 
> PROPOSAL: We retain both the dryRun and provision parameters, allowing customers to mix and match them as they wish.

Agreed with this proposal

dbwiddis:
> > My understanding of all the conditions is:
> > dryRun=false && provision=false, no formate validation, no provision. Just save the template to index
> > dryRun=false && provision=true, save the template to index, then provisioning the template. Provision could failed due to template formatting error like missing certain input files, or could due to template is a cyclic graph. (it just performs all the check that come with the Provision API)
> > dryRun=true && provision = false, do formate validation first, only save the template to index if it passes the check.
> > dryRun=true && provision = true, do formate validation first, save the template to index if it passes the check, then provisioning the template. Although provision step also checks for formatting issues, which is already covered by dryRun check.
> > I think no matter how customer mix&match these two parameters, they should all be legal. The dryRun=true && provision = true case helps customer to fail early is there's any formatting issue
> > PROPOSAL: We retain both the dryRun and provision parameters, allowing customers to mix and match them as they wish.
> 
> Agreed with this proposal

I'm ok with this except that it doesn't match my understanding of the term "dryRun".  For example in AD the "create detector" with dryRun does not actually create a detector (actually it's just a variable but it corresponds to validating).

So if we're going this way let's use "validate" rather than dry run.

https://en.wikipedia.org/wiki/Dry_run_(testing)

dbwiddis:
Other opensearch use of dryrun: https://opensearch.org/docs/latest/observing-your-data/alerting/api/
> You can add the optional ?dryrun=true parameter to the URL to show the results of a run without actions sending any message.

Use of validate: https://opensearch.org/docs/latest/observing-your-data/ad/api/
> Returns whether the detector configuration has any issues that might prevent OpenSearch from creating the detector.
>
> You can use the validate detector API operation to identify issues in your detector configuration before creating the detector.

Use of validate: https://opensearch.org/docs/latest/im-plugin/ism/error-prevention/api/
> Pass the validate_action=true path parameter in the Explain API URI to see the validation status and message.

jackiehanyang:
> > > My understanding of all the conditions is:
> > > dryRun=false && provision=false, no formate validation, no provision. Just save the template to index
> > > dryRun=false && provision=true, save the template to index, then provisioning the template. Provision could failed due to template formatting error like missing certain input files, or could due to template is a cyclic graph. (it just performs all the check that come with the Provision API)
> > > dryRun=true && provision = false, do formate validation first, only save the template to index if it passes the check.
> > > dryRun=true && provision = true, do formate validation first, save the template to index if it passes the check, then provisioning the template. Although provision step also checks for formatting issues, which is already covered by dryRun check.
> > > I think no matter how customer mix&match these two parameters, they should all be legal. The dryRun=true && provision = true case helps customer to fail early is there's any formatting issue
> > > PROPOSAL: We retain both the dryRun and provision parameters, allowing customers to mix and match them as they wish.
> > 
> > 
> > Agreed with this proposal
> 
> I'm ok with this except that it doesn't match my understanding of the term "dryRun". For example in AD the "create detector" with dryRun does not actually create a detector (actually it's just a variable but it corresponds to validating).
> 
> So if we're going this way let's use "validate" rather than dry run.
> 
> https://en.wikipedia.org/wiki/Dry_run_(testing)

When dryRun is set to true, we proceed with indexing only if it passes the validation check - this is the current design of the create workflow API. If we think it's necessary to revise the design of the create workflow API, I can open a separate issue for this matter.

dbwiddis:
> When dryRun is set to true, we proceed with indexing only if it passes the validation check 

I'm fine with this, I'm just quibbling over using the term "dryRun" (used only once in the opensearch API) vs. "validate" used many times and much more clear on what's happening.

jackiehanyang:
> > When dryRun is set to true, we proceed with indexing only if it passes the validation check
> 
> I'm fine with this, I'm just quibbling over using the term "dryRun" (used only once in the opensearch API) vs. "validate" used many times and much more clear on what's happening.

Let's the continue the discussion in this issue - https://github.com/opensearch-project/flow-framework/issues/291

dbwiddis (approved):

joshpalis (approved):

src/main/java/org/opensearch/flowframework/rest/RestCreateWorkflowAction.java
amitgalitz: I agree we can have check here to throw an exception if they are both equaling true. 

The naming here is a little weird because when I see a `dryrun` or `validate` I don't expect to be creating something. And when I create something I would assume some inherit validation is done. However in our case users might want to create even with errors and come back to edit the template later (its more of a save action).

I think for now, like Dan said we should not allow both provision and dryrun to be true together but we should revisit this in the issue Jackie made, I will put further thoughts there

src/main/java/org/opensearch/flowframework/transport/CreateWorkflowTransportAction.java
```diff
@@ -135,7 +135,28 @@ protected void doExecute(Task task, WorkflowRequest request, ActionListener<Work
                                         user,
                                         ActionListener.wrap(stateResponse -> {
                                             logger.info("create state workflow doc");
-                                            listener.onResponse(new WorkflowResponse(globalContextResponse.getId()));
+                                            if (request.isProvision()) {
+                                                logger.info("provision parameter");
```
amitgalitz, line 139: nit: can you change to "provisioning workflow after creation" or "provisioning param is set to true", not 100% how clear `provision parameter` is.

amitgalitz (approved):
LGTM, one very minor comment only

PR #283, opened 2023-12-13
Permit ordering of tools in register agent step

### LINKED ISSUE #281: [FEATURE] Order of tools in Register Agent Step should be deterministic

For some agents (Flow Agent), the order of the tools matters.  Currently when creating the list of tools we are using non-deterministic ordering from a HashMap entry set to create the `previousNodes` list.
```java
List<String> previousNodes = previousNodeInputs.entrySet()
    .stream()
    .filter(e -> TOOLS_FIELD.equals(e.getValue()))
    .map(Map.Entry::getKey)
```
We need a way to (optionally) enforce an ordering of these tools.

The AgentStep already has an optional "tools" field in the code but it is never read, instead just iterating over the previous node inputs and looking for "tool" in the value.

If that field is populated with an array of tool ID's (the workflow step ID they were created with), the list should be sorted in that order.

Always sort in order of workflow step ID.  However, we may want different agents with different orderings created from the same definition.

dbwiddis:
### Description

Allows users to include an optional "tools" field in the register agent node.  If present, any previous nodes listed in that array will be used in order prior to any other tools listed in previous inputs.  Duplicates will be ignored ("tools" field takes priority).

### Issues Resolved

Fixes #281 

### DIFF
--- a/src/main/java/org/opensearch/flowframework/model/WorkflowNode.java
+++ b/src/main/java/org/opensearch/flowframework/model/WorkflowNode.java
@@ -24,6 +24,7 @@ import java.util.Map.Entry;
 import java.util.Objects;
 
 import static org.opensearch.core.xcontent.XContentParserUtils.ensureExpectedToken;
+import static org.opensearch.flowframework.common.CommonValue.TOOLS_FIELD;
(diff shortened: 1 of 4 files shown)


ohltyler (approved):
LGTM

joshpalis (approved):

Issue #126, opened 2023-10-30
[FEATURE] Check the plugins installed before running provisioning


owaiskazi19:
### Is your feature request related to a problem?
Coming from 2nd task of https://github.com/opensearch-project/opensearch-ai-flow-framework/issues/108. 

Provision API runs a workflow Step which is nothing but a building block for a specific plugin. To execute the APIs of that plugin, it needs to be installed in the OS cluster or else flow framework plugin won't be able to do provisioning. We can't add all the plugins in `extendedPlugins`, that will create dependency on all the plugins even if their APIs are not required in any workflow step of the template.

### What solution would you like?
There should be a mechanism to check the list of plugins installed and template should have a field called as `plugins` consisting of list of plugins required for a specific workflow. The first template validation should be to check if the plugins mentioned in the template are installed in the OS cluster.

### What alternatives have you considered?
_A clear and concise description of any alternative solutions or features you've considered._

### Do you have any additional context?
_Add any other context or screenshots about the feature request here._

dbwiddis:
My preferred implementation:  
1. add a field "required_plugins" (String array) to the workflow step JSON [here](https://github.com/opensearch-project/opensearch-ai-flow-framework/blob/main/src/main/resources/mappings/workflow-steps.json).
2. when validating a plugin, iterate the steps (which we are already doing for validating input/output) and collect the strings from this entry, adding them to a `Set` as we go along
3. take this set and compare it to the list of installed plugins.  See [sample code here](https://github.com/opensearch-project/opensearch-ai-flow-framework/blob/ac76a44927a6e475694f563b2a033aaa498d13e8/src/test/java/org/opensearch/flowframework/FlowFrameworkPluginIT.java#L22-L29) for obtaining installed plugins.

joshpalis:
Closed by https://github.com/opensearch-project/flow-framework/pull/290

PR #290, opened 2023-12-14
[Feature/agent_framework] Adding installed plugins validation

joshpalis:
### Description
This PR adds a new field `required_plugins` to the `workflow_step` json which is used to determine if the workflow step's corresponding plugin is currently installed within the local node. Additionally, this PR adds validation when `creating` (with validation) and `provisioning` to check if the given graph has all the required plugins prior to executing. 

The following example uses a create connector node with an intentionally modified required plugin named `test-plugin` to demonstrate how the validation works : 

Creating the workflow : 
```
curl -i -XPOST "localhost:9200/_plugins/_flow_framework/workflow" -H "Content-Type:application/json" --data '{"name":"create-connector-register-deploy-model","description":"test case","use_case":"TEST_CASE","version":{"template":"1.0.0","compatibility":["2.12.0","3.0.0"]},"workflows":{"provision":{"nodes":[{"id":"workflow_step_1","type":"create_connector","user_inputs":{"name":"OpenAI Chat Connector","description":"The connector to public OpenAI model service for GPT 3.5","version":"1","protocol":"http","parameters":{"endpoint":"api.openai.com","model":"gpt-3.5-turbo"},"credential":{"openAI_key":"12345"},"actions":[{"action_type":"predict","method":"POST","url":"https://${parameters.endpoint}/v1/chat/completions"}]}},{"id":"workflow_step_2","type":"register_remote_model","previous_node_inputs":{"workflow_step_1":"connector_id"},"user_inputs":{"name":"openAI-gpt-3.5-turbo","function_name":"remote","description":"test model"}},{"id":"workflow_step_3","type":"deploy_model","previous_node_inputs":{"workflow_step_2":"model_id"}}],"edges":[{"source":"workflow_step_1","dest":"workflow_step_2"},{"source":"workflow_step_2","dest":"workflow_step_3"}]}}}'

HTTP/1.1 201 Created
content-length: 38

{"workflow_id":"n-jqaYwBsLRi2-luOv83"}
```
Provisioning the workflow results in an error that states that the `create_connector` step requires the `test-plugin` to be installed
```
curl -i -XPOST "localhost:9200/_plugins/_flow_framework/workflow/n-jqaYwBsLRi2-luOv83/_provision"
HTTP/1.1 400 Bad Request
content-length: 108

{"error":"The workflowStep create_connector requires the following plugins to be installed : [test-plugin]"}
```

### Issues Resolved
https://github.com/opensearch-project/flow-framework/issues/126
Part of #88 

### DIFF
--- a/src/main/java/org/opensearch/flowframework/transport/ProvisionWorkflowTransportAction.java
+++ b/src/main/java/org/opensearch/flowframework/transport/ProvisionWorkflowTransportAction.java
@@ -123,6 +123,7 @@ public class ProvisionWorkflowTransportAction extends HandledTransportAction<Wor
                 Workflow provisionWorkflow = template.workflows().get(PROVISION_WORKFLOW);
                 List<ProcessNode> provisionProcessSequence = workflowProcessSorter.sortProcessNodes(provisionWorkflow, workflowId);
                 workflowProcessSorter.validateGraph(provisionProcessSequence);
+                workflowProcessSorter.validatePluginsInstalled(provisionProcessSequence);
 
                 flowFrameworkIndicesHandler.updateFlowFrameworkSystemIndexDoc(
                     workflowId,
--- a/src/main/java/org/opensearch/flowframework/transport/ProvisionWorkflowTransportAction.java
+++ b/src/main/java/org/opensearch/flowframework/transport/ProvisionWorkflowTransportAction.java
@@ -122,8 +122,7 @@ public class ProvisionWorkflowTransportAction extends HandledTransportAction<Wor
                 // Sort and validate graph
                 Workflow provisionWorkflow = template.workflows().get(PROVISION_WORKFLOW);
                 List<ProcessNode> provisionProcessSequence = workflowProcessSorter.sortProcessNodes(provisionWorkflow, workflowId);
-                workflowProcessSorter.validateGraph(provisionProcessSequence);
-                workflowProcessSorter.validatePluginsInstalled(provisionProcessSequence);
+                workflowProcessSorter.validate(provisionProcessSequence);
 
                 flowFrameworkIndicesHandler.updateFlowFrameworkSystemIndexDoc(
                     workflowId,
--- a/src/main/java/org/opensearch/flowframework/workflow/WorkflowProcessSorter.java
+++ b/src/main/java/org/opensearch/flowframework/workflow/WorkflowProcessSorter.java
@@ -10,15 +10,21 @@ package org.opensearch.flowframework.workflow;
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.opensearch.action.admin.cluster.node.info.NodeInfo;
+import org.opensearch.action.admin.cluster.node.info.NodesInfoRequest;
+import org.opensearch.action.admin.cluster.node.info.PluginsAndModules;
+import org.opensearch.client.Client;
 import org.opensearch.cluster.service.ClusterService;
 import org.opensearch.common.settings.Settings;
 import org.opensearch.common.unit.TimeValue;
+import org.opensearch.core.action.ActionListener;
 import org.opensearch.core.rest.RestStatus;
 import org.opensearch.flowframework.exception.FlowFrameworkException;
 import org.opensearch.flowframework.model.Workflow;
 import org.opensearch.flowframework.model.WorkflowEdge;
 import org.opensearch.flowframework.model.WorkflowNode;
 import org.opensearch.flowframework.model.WorkflowValidator;
+import org.opensearch.plugins.PluginInfo;
 import org.opensearch.threadpool.ThreadPool;
 
 import java.util.ArrayDeque;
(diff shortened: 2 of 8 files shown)


src/main/java/org/opensearch/flowframework/transport/ProvisionWorkflowTransportAction.java
```diff
@@ -123,6 +123,7 @@ protected void doExecute(Task task, WorkflowRequest request, ActionListener<Work
                 Workflow provisionWorkflow = template.workflows().get(PROVISION_WORKFLOW);
                 List<ProcessNode> provisionProcessSequence = workflowProcessSorter.sortProcessNodes(provisionWorkflow, workflowId);
                 workflowProcessSorter.validateGraph(provisionProcessSequence);
+                workflowProcessSorter.validatePluginsInstalled(provisionProcessSequence);
```
dbwiddis, line 126: There's some duplication in these two methods.  Suggest combining these into a single method call, then you only have to call `WorkflowValidator validator = WorkflowValidator.parse("mappings/workflow-steps.json")` once and can pass that object to the two individual methods.

src/main/java/org/opensearch/flowframework/workflow/WorkflowProcessSorter.java
```diff
@@ -122,6 +135,59 @@ public List<ProcessNode> sortProcessNodes(Workflow workflow, String workflowId)
...
+
+            installedPluginsFuture.complete(installedPlugins);
+
+        }, exception -> {
+            logger.error("Failed to retrieve installed plugins");
+            installedPluginsFuture.completeExceptionally(exception);
+        }));
+
+        // Block execution until installed plugin list is returned
+        List<String> installedPlugins = installedPluginsFuture.get();
```
dbwiddis, line 170: 1. This blocks.  Use `join()` instead to wait unblocking, and call `get()` after the `join()` passes.

2. If you've completed exceptionally (line 166), this (or `join()`) will throw an exception here when you try to `get()`.  Probably want to catch that and throw a FlowFrameworkException.

dbwiddis (commented):
Great job!  First pass, some suggestions.

src/main/java/org/opensearch/flowframework/transport/ProvisionWorkflowTransportAction.java
joshpalis: I intentionally left this separate from graph validation since we use graph validation when creating a workflow with the dry-run flag set to true. I can combine these, but then we would have installed plugin validation during create instead of only provision, which goes against the requirement in the issue. 

I can combine both if we're ok with having installed plugin validation for both APIs

dbwiddis: I think it's reasonable to validate both.

joshpalis: sure ill make this change

src/main/java/org/opensearch/flowframework/workflow/WorkflowProcessSorter.java
dbwiddis: Actually I'm wrong.  join blocks too.  There's `thenAccept()` to do nonblocking.... but for now let's leave this as-is.

dbwiddis (approved):
Approved, suggestions above remain suggestions for consideration but OK as is.

src/main/java/org/opensearch/flowframework/transport/ProvisionWorkflowTransportAction.java
joshpalis: I have confirmed that installed plugin validation works for creating a workflow with validation (using an incorrect required plugin name for the create connector step) : 
```
curl -i -XPOST "localhost:9200/_plugins/_flow_framework/workflow?dryrun=true" -H "Content-Type:application/json" --data '{"name":"create-connector-register-deploy-model","description":"test case","use_case":"TEST_CASE","version":{"template":"1.0.0","compatibility":["2.12.0","3.0.0"]},"workflows":{"provision":{"nodes":[{"id":"workflow_step_1","type":"create_connector","user_inputs":{"name":"OpenAI Chat Connector","description":"The connector to public OpenAI model service for GPT 3.5","version":"1","protocol":"http","parameters":{"endpoint":"api.openai.com","model":"gpt-3.5-turbo"},"credential":{"openAI_key":"12345"},"actions":[{"action_type":"predict","method":"POST","url":"https://${parameters.endpoint}/v1/chat/completions"}]}},{"id":"workflow_step_2","type":"register_remote_model","previous_node_inputs":{"workflow_step_1":"connector_id"},"user_inputs":{"name":"openAI-gpt-3.5-turbo","function_name":"remote","description":"test model"}},{"id":"workflow_step_3","type":"deploy_model","previous_node_inputs":{"workflow_step_2":"model_id"}}],"edges":[{"source":"workflow_step_1","dest":"workflow_step_2"},{"source":"workflow_step_2","dest":"workflow_step_3"}]}}}'

HTTP/1.1 400 Bad Request
content-length: 107

{"error":"The workflowStep create_connector requires the following plugins to be installed : [testsesete]"}
```

PR #298, opened 2023-12-16
Refactor RetryableWorkflowStep to avoid recursion and subclass specifics

### LINKED ISSUE #277: [CHORE] Refactor AbstractRetryableWorkflowStep

The `AbstractRetryableWorkflowStep` works for existing register and deploy model steps which return a task ID and require periodic polling until completion.  However, it needs a bit more work to be more generic and robust to other tasks and longer-running tasks.

1. Any reference to specific subclass implementations should be removed.  If needed, abstract method stubs to be implemented by subclasses can be used.
2. The existing implementation is recursive using exceptions as logic control.  Ideally this can be replaced with a loop.
3. A constant retry time is used, which may not be appropriate for very long running tasks and is not user-configurable.  This logic should be updated to allow a template to specify retry times.

dbwiddis:
### Description

Refactors the `RetryableWorkflowStep` class:
 - runs the `getTask()` code in a loop instead of using recursion
 - uses an abstract method to allow the subclass to provide the needed ID method

### Issues Resolved

Fixes #277 

### DIFF
--- a/src/main/java/org/opensearch/flowframework/workflow/AbstractRetryableWorkflowStep.java
+++ b/src/main/java/org/opensearch/flowframework/workflow/AbstractRetryableWorkflowStep.java
@@ -125,8 +125,9 @@ public abstract class AbstractRetryableWorkflowStep implements WorkflowStep {
                             future.completeExceptionally(new FlowFrameworkException(errorMessage, RestStatus.BAD_REQUEST));
                             break;
                         case CANCELLED:
-                            logger.error("{} task was cancelled.", workflowStep);
-                            FutureUtils.cancel(future);
+                            errorMessage = workflowStep + " task was cancelled.";
+                            logger.error(errorMessage);
+                            future.completeExceptionally(new FlowFrameworkException(errorMessage, RestStatus.REQUEST_TIMEOUT));
                             break;
                         default:
                             // Task started or running, do nothing
--- a/src/main/java/org/opensearch/flowframework/workflow/AbstractRetryableWorkflowStep.java
+++ b/src/main/java/org/opensearch/flowframework/workflow/AbstractRetryableWorkflowStep.java
@@ -125,7 +125,7 @@ public abstract class AbstractRetryableWorkflowStep implements WorkflowStep {
                             future.completeExceptionally(new FlowFrameworkException(errorMessage, RestStatus.BAD_REQUEST));
                             break;
                         case CANCELLED:
-                            logger.error(workflowStep + " task was cancelled.");
+                            logger.error("{} task was cancelled.", workflowStep);
                             FutureUtils.cancel(future);
                             break;
                         default:
@@ -141,6 +141,7 @@ public abstract class AbstractRetryableWorkflowStep implements WorkflowStep {
                     Thread.sleep(5000);
                 } catch (InterruptedException e) {
                     FutureUtils.cancel(future);
+                    Thread.currentThread().interrupt();
                 }
             }
             if (!future.isDone()) {
--- a/src/main/java/org/opensearch/flowframework/workflow/AbstractRetryableWorkflowStep.java
+++ b/src/main/java/org/opensearch/flowframework/workflow/AbstractRetryableWorkflowStep.java
@@ -19,15 +19,16 @@ import org.opensearch.core.rest.RestStatus;
 import org.opensearch.flowframework.exception.FlowFrameworkException;
 import org.opensearch.flowframework.indices.FlowFrameworkIndicesHandler;
 import org.opensearch.ml.client.MachineLearningNodeClient;
-import org.opensearch.ml.common.MLTaskState;
+import org.opensearch.ml.common.MLTask;
+import org.opensearch.threadpool.ThreadPool;
 
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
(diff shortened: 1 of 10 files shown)


joshpalis (approved):

src/main/java/org/opensearch/flowframework/workflow/AbstractRetryableWorkflowStep.java
```diff
@@ -65,82 +70,92 @@ protected AbstractRetryableWorkflowStep(
...
-                logger.error("Failed to retrieve" + workflowStep + ",maximum retries exceeded");
-                future.completeExceptionally(new FlowFrameworkException(exception.getMessage(), ExceptionsHelper.status(exception)));
             }
-        }));
+            if (!future.isDone()) {
+                String errorMessage = workflowStep + " did not complete after " + maxRetry + " retries";
+                logger.error(errorMessage);
+                future.completeExceptionally(new FlowFrameworkException(errorMessage, RestStatus.REQUEST_TIMEOUT));
+            }
+        }, threadPool.executor(PROVISION_THREAD_POOL));
```
amitgalitz, line 152: just for learning purpose, why do we need this exactly now, when we didn't have this before (the `threadPool.executor`

src/main/java/org/opensearch/flowframework/workflow/AbstractRetryableWorkflowStep.java
```diff
@@ -65,82 +70,92 @@ protected AbstractRetryableWorkflowStep(
...
+                            break;
+                        case FAILED:
+                        case COMPLETED_WITH_ERROR:
+                            String errorMessage = workflowStep + " failed with error : " + response.getError();
+                            logger.error(errorMessage);
+                            future.completeExceptionally(new FlowFrameworkException(errorMessage, RestStatus.BAD_REQUEST));
+                            break;
+                        case CANCELLED:
+                            logger.error("{} task was cancelled.", workflowStep);
+                            FutureUtils.cancel(future);
```
amitgalitz, line 129: unlike other cases we don't `completeExceptionally` here, does this mean we don't propagate an exception to update the state index with? Not sure if you tested this, just cause we are treating `cancelled` differently here

src/main/java/org/opensearch/flowframework/workflow/AbstractRetryableWorkflowStep.java
dbwiddis: Previously we were using an ActionListener (either `onResponse` or `onFailure`) to respond to the result, and relying on an exception to terminate that listener thread.   The new implementation uses `runAsync()` which needs an executor.  We could use the OpenSearch `generic()` one but since we've defined our own provision thread pool and this retry is part of provisioning it seemed the best one to use for this purpose.

src/main/java/org/opensearch/flowframework/workflow/AbstractRetryableWorkflowStep.java
dbwiddis: Cancelling a future will complete it, which either `join()` or `get()` is waiting on.   Since it is not completed normally, it will throw a `CancellationException` (unchecked) on either of those methods.  

The future completed here is the one returned from the associated WorkflowStep subclass to the ProcessNode where it is then returned to the `ProvisionWorkflowTransportAction` (and corresponding deprovision equivalent).  That calls `join` on all the step futures, so will throw an exception on this line.
```java
workflowFutureList.forEach(CompletableFuture::join);
```

This will be caught in the same catch block as being completed exceptionally, but with the different exception type.  

Since the `CancellationException` thrown from the cancelled future won't have a message, we may want to actually test the exception type in that catch block and give a different response.  WDYT?

dbwiddis: Handled the cancellation (and throwable with cause) error messages in https://github.com/opensearch-project/flow-framework/pull/298/commits/13426e4947e70f71ffb50ec1786d4657a6055e5c

amitgalitz: Yes this looks good to me so we can make sure that we update the error message with either the error or in case of cancellation some sort of indication that a cancellation happened, I wonder if we can improve though by propagating what step was cancelled, for now I am okay with it being generic like in your commit.

src/main/java/org/opensearch/flowframework/workflow/AbstractRetryableWorkflowStep.java
amitgalitz: that sounds good to me, I am not sure on the 100% best standard of using threads in this way from within the plugin, but relaying on our provision thread pool makes sense to me

amitgalitz (approved):

src/main/java/org/opensearch/flowframework/workflow/AbstractRetryableWorkflowStep.java
dbwiddis: Just completed exceptionally instead of cancelling in https://github.com/opensearch-project/flow-framework/pull/298/commits/a0fc9834f16685b064530d5159ee59df65a33efd

Issue #291, opened 2023-12-14
Replace the term dryrun with validate within create workflow api


jackiehanyang:
Currently, create workflow api has a dryrun parameter that allows customer to do a format validation check before indexing the template. When `dryrun=true`, we will create if the validation passes. Now based on the conversion in this [pr](https://github.com/opensearch-project/flow-framework/pull/282#issuecomment-1856199900), we have these two proposals:

- change the term dryrun with validate
- don't create when the parameter is set to true, only perform validation check

dbwiddis:
OK, after taking some time to think on this
 - "dry run" can sometimes imply validation but not always. It's appropriate for a "check to see if this works without actually doing it" for an API where "actually doing it" is the norm.  It's roughly synonymous, but there are a vew implied differences
   - "dry run" is added when we are by default always doing validation, as an option to *not* do the validation, so is appropriate if validation is the default outcome (this is our case)
   - "validate" is often added when we are by default not doing validation, so is appropriate if not validating is the default outcome, but it's also equally possible to default "on" and allow turning it off.

There's a more important connotation of "dry run" in that "you don't actually do the thing the API implies".  So I would expect:
 - for a "create" API, dry run means I don't create it, I just validate if creation would have been successful
 - for a "provision" api, dry run means I don't provision, I just validate if creation would have been successful
 - for a combo create+provision API, it is ambiguous

Then there's the user perspective:
 - Generally by default I don't want to save an invalid template
   - But sometimes during "WIP" editing situation I do want to save an invalid template.  I've been frustrated in the past with "precommit hooks" that don't even let me commit changed files in git if they don't pass the various validation checks, but I often want to just "save a spot I can come back to".

I think the situation should generally be:
 - always do validation and fail fast with no action if it fails
 - allow the user to override this default validation and do things like creating an (invalid) template.  
   - this seems to call out for a "validate=false" param, which is what prompted my original comment.  "dryrun=false" doesn't seem to make sense.

Related: Maybe we need to add a field to the saved template that says it's invalid, so we don't provision it.

dbwiddis:
> this seems to call out for a "validate=false" param

Leading to create API:
 - validate=true (default), provision=false (default):  validates and saves only if successful
 - validate=false (override), provision=false (default):  does no validation, saves
 - validate=true (default), provision=true (override):  validates and saves only if successful, then provisions
[...]

PR #308, opened 2023-12-20
Change dryrun param in create workflow to validation

jackiehanyang:
### Description
This PR changes the `dryrun` parameter in create workflow api to `validation`. And updating related integration test case.

### Issues Resolved
https://github.com/opensearch-project/flow-framework/issues/291

### DIFF
--- a/src/main/java/org/opensearch/flowframework/rest/RestCreateWorkflowAction.java
+++ b/src/main/java/org/opensearch/flowframework/rest/RestCreateWorkflowAction.java
@@ -32,8 +32,8 @@ import java.util.List;
 import java.util.Locale;
 
 import static org.opensearch.core.xcontent.XContentParserUtils.ensureExpectedToken;
-import static org.opensearch.flowframework.common.CommonValue.DRY_RUN;
 import static org.opensearch.flowframework.common.CommonValue.PROVISION_WORKFLOW;
+import static org.opensearch.flowframework.common.CommonValue.VALIDATION;
 import static org.opensearch.flowframework.common.CommonValue.WORKFLOW_ID;
 import static org.opensearch.flowframework.common.CommonValue.WORKFLOW_URI;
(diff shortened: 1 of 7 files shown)


src/main/java/org/opensearch/flowframework/rest/RestCreateWorkflowAction.java
```diff
@@ -94,10 +94,17 @@ protected RestChannelConsumer prepareRequest(RestRequest request, NodeClient cli
             XContentParser parser = request.contentParser();
             ensureExpectedToken(XContentParser.Token.START_OBJECT, parser.nextToken(), parser);
             Template template = Template.parse(parser);
-            boolean dryRun = request.paramAsBoolean(DRY_RUN, false);
+            boolean validation = request.paramAsBoolean(VALIDATION, true);
```
dbwiddis, line 97: So after thinking on this for a while, I think I like "validation" as a name but would like it to be more fine-grained.  We don't need to do all the choices for 2.12, but can we change this to a `String` parameter defaulting to "all"?

That will leave the door open to adding more options in the future like maybe validating everything except plugins "-plugins" or only version compatibility, etc.

dbwiddis (commented):
LGTM but a suggestion for consideration.

src/main/java/org/opensearch/flowframework/rest/RestCreateWorkflowAction.java
dbwiddis: For clarity the (future) capability I'm envisioning is similar to the OpenSearch [Nodes Stats API](https://opensearch.org/docs/latest/api-reference/nodes-apis/nodes-stats/) where you can specify a list of stats you want, or "all" to get all of 'em.

dbwiddis (approved):
LGTM!

Issue #249, opened 2023-12-04
[FEATURE] Add node timeout for the workflow steps


owaiskazi19:
### Is your feature request related to a problem?
Currently, we have a default timeout for all the workflow steps 
https://github.com/opensearch-project/flow-framework/blob/3f6559673f2bfdd22b045f622771525da448b7ee/src/main/java/org/opensearch/flowframework/model/WorkflowNode.java#L51

Coming from https://github.com/opensearch-project/flow-framework/pull/245#discussion_r1413290533 we should have a specific timeout for all the workflow steps in workflow-steps.json.

### What solution would you like?
Add specific timeout for the workflow steps and make sure to have `15s` for Deploy model step. Rest can be `10s`.

### What alternatives have you considered?
Make the default timeout to `15s` which is not a good option.

### Do you have any additional context?
_Add any other context or screenshots about the feature request here._

PR #317, opened 2023-12-23
Add timeout for each workflow steps

jackiehanyang:
### 12/26 revision:
- Change the system default `NODE_TIMEOUT_DEFAULT_VALUE` to `10s`
- Removed all `10s` timeout values in the workflow-step.json file, only leaving the timeout field for those workflow steps that require an overwrite timeout value other than `10s`.
For example, `register_local_model` has default timeout value as 60s, and `deploy_model` has default timeout value as 15s

- Do try catch at where we read `workflow-step.json` file instead of passing it up. 
- Add a new unit test case to cover all possible ways of how we parse `timeout` value

--------

### Description
This pr does two things:
- Having a specific timeout for all the workflow steps in workflow-steps.json
- rename indices:
 `.plugins-ai-global-context` -> `.plugins-flow-framework-templates`
 `.plugins-workflow-state` -> `.plugins-flow-framework-state`

### Issues Resolved
https://github.com/opensearch-project/flow-framework/issues/249

### DIFF
--- a/src/main/java/org/opensearch/flowframework/model/WorkflowNode.java
+++ b/src/main/java/org/opensearch/flowframework/model/WorkflowNode.java
@@ -8,6 +8,7 @@
  */
 package org.opensearch.flowframework.model;
 
+import org.opensearch.common.unit.TimeValue;
 import org.opensearch.core.xcontent.ToXContentObject;
 import org.opensearch.core.xcontent.XContentBuilder;
 import org.opensearch.core.xcontent.XContentParser;
@@ -23,6 +24,7 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Objects;
 
+import static java.util.concurrent.TimeUnit.SECONDS;
 import static org.opensearch.core.xcontent.XContentParserUtils.ensureExpectedToken;
 import static org.opensearch.flowframework.common.CommonValue.TOOLS_ORDER_FIELD;
 import static org.opensearch.flowframework.util.ParseUtils.buildStringToObjectMap;
@@ -50,7 +52,7 @@ public class WorkflowNode implements ToXContentObject {
     /** The field defining the timeout value for this node */
     public static final String NODE_TIMEOUT_FIELD = "node_timeout";
     /** The default timeout value if the template doesn't override it */
-    public static final String NODE_TIMEOUT_DEFAULT_VALUE = "10s";
+    public static final TimeValue NODE_TIMEOUT_DEFAULT_VALUE = new TimeValue(10, SECONDS);
 
     private final String id; // unique id
     private final String type; // maps to a WorkflowStep
--- a/src/main/java/org/opensearch/flowframework/model/WorkflowNode.java
+++ b/src/main/java/org/opensearch/flowframework/model/WorkflowNode.java
@@ -50,7 +50,7 @@ public class WorkflowNode implements ToXContentObject {
     /** The field defining the timeout value for this node */
     public static final String NODE_TIMEOUT_FIELD = "node_timeout";
     /** The default timeout value if the template doesn't override it */
-    public static final String NODE_TIMEOUT_DEFAULT_VALUE = "15s";
+    public static final String NODE_TIMEOUT_DEFAULT_VALUE = "10s";
 
     private final String id; // unique id
     private final String type; // maps to a WorkflowStep
--- a/src/main/java/org/opensearch/flowframework/model/WorkflowStepValidator.java
+++ b/src/main/java/org/opensearch/flowframework/model/WorkflowStepValidator.java
@@ -95,7 +95,7 @@ public class WorkflowStepValidator {
                     try {
                         timeout = TimeValue.parseTimeValue(parser.text(), TIMEOUT);
                     } catch (IllegalArgumentException e) {
-                        logger.error("Failed to parse TIMEOUT value for field [" + fieldName + "]", e);
+                        logger.error("Failed to parse TIMEOUT value for field [{}]", fieldName, e);
                         throw new FlowFrameworkException(
                             "Failed to parse workflow-step.json file for field [" + fieldName + "]",
                             RestStatus.INTERNAL_SERVER_ERROR
--- a/src/main/java/org/opensearch/flowframework/model/WorkflowStepValidator.java
+++ b/src/main/java/org/opensearch/flowframework/model/WorkflowStepValidator.java
@@ -40,6 +40,7 @@ public class WorkflowStepValidator {
      * @param inputs the workflow step inputs
      * @param outputs the workflow step outputs
      * @param requiredPlugins the required plugins for this workflow step
+     * @param timeout the timeout for this workflow step
      */
     public WorkflowStepValidator(List<String> inputs, List<String> outputs, List<String> requiredPlugins, String timeout) {
         this.inputs = inputs;
--- a/src/main/java/org/opensearch/flowframework/model/WorkflowStepValidator.java
+++ b/src/main/java/org/opensearch/flowframework/model/WorkflowStepValidator.java
@@ -27,21 +27,25 @@ public class WorkflowStepValidator {
     private static final String OUTPUTS_FIELD = "outputs";
     /** Required Plugins field name */
     private static final String REQUIRED_PLUGINS = "required_plugins";
+    /** Timeout field name */
+    private static final String TIMEOUT = "timeout";
 
     private List<String> inputs;
     private List<String> outputs;
     private List<String> requiredPlugins;
+    private String timeout;
 
     /**
-     * Intantiate the object representing a Workflow Step validator
+     * Instantiate the object representing a Workflow Step validator
      * @param inputs the workflow step inputs
      * @param outputs the workflow step outputs
      * @param requiredPlugins the required plugins for this workflow step
      */
-    public WorkflowStepValidator(List<String> inputs, List<String> outputs, List<String> requiredPlugins) {
+    public WorkflowStepValidator(List<String> inputs, List<String> outputs, List<String> requiredPlugins, String timeout) {
         this.inputs = inputs;
         this.outputs = outputs;
         this.requiredPlugins = requiredPlugins;
+        this.timeout = timeout;
     }
 
     /**
@@ -54,6 +58,7 @@ public class WorkflowStepValidator {
         List<String> parsedInputs = new ArrayList<>();
         List<String> parsedOutputs = new ArrayList<>();
         List<String> requiredPlugins = new ArrayList<>();
+        String timeout = null;
 
         ensureExpectedToken(XContentParser.Token.START_OBJECT, parser.currentToken(), parser);
         while (parser.nextToken() != XContentParser.Token.END_OBJECT) {
@@ -78,11 +83,14 @@ public class WorkflowStepValidator {
                         requiredPlugins.add(parser.text());
                     }
                     break;
+                case TIMEOUT:
+                    timeout = parser.text();
+                    break;
                 default:
                     throw new IOException("Unable to parse field [" + fieldName + "] in a WorkflowStepValidator object.");
             }
         }
-        return new WorkflowStepValidator(parsedInputs, parsedOutputs, requiredPlugins);
+        return new WorkflowStepValidator(parsedInputs, parsedOutputs, requiredPlugins, timeout);
     }
 
     /**
@@ -103,9 +111,17 @@ public class WorkflowStepValidator {
 
     /**
      * Get the required plugins
-     * @return the outputs
+     * @return the required plugins
      */
     public List<String> getRequiredPlugins() {
         return List.copyOf(requiredPlugins);
+    }
+
+    /**
+     * Get the timeout
+     * @return the timeout
+     */
+    public String getTimeout() {
+        return timeout;
     }
 }
--- a/src/main/java/org/opensearch/flowframework/model/WorkflowStepValidator.java
+++ b/src/main/java/org/opensearch/flowframework/model/WorkflowStepValidator.java
@@ -8,7 +8,12 @@
  */
 package org.opensearch.flowframework.model;
 
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.opensearch.common.unit.TimeValue;
+import org.opensearch.core.rest.RestStatus;
 import org.opensearch.core.xcontent.XContentParser;
+import org.opensearch.flowframework.exception.FlowFrameworkException;
 
 import java.io.IOException;
 import java.util.ArrayList;
@@ -20,6 +25,8 @@ import static org.opensearch.core.xcontent.XContentParserUtils.ensureExpectedTok
  * This represents an object of workflow steps json which maps each step to expected inputs and outputs
  */
 public class WorkflowStepValidator {
+
+    private static final Logger logger = LogManager.getLogger(WorkflowStepValidator.class);
 
     /** Inputs field name */
     private static final String INPUTS_FIELD = "inputs";
@@ -33,7 +40,7 @@ public class WorkflowStepValidator {
     private List<String> inputs;
     private List<String> outputs;
     private List<String> requiredPlugins;
-    private String timeout;
+    private TimeValue timeout;
 
     /**
      * Instantiate the object representing a Workflow Step validator
@@ -42,7 +49,7 @@ public class WorkflowStepValidator {
      * @param requiredPlugins the required plugins for this workflow step
      * @param timeout the timeout for this workflow step
      */
-    public WorkflowStepValidator(List<String> inputs, List<String> outputs, List<String> requiredPlugins, String timeout) {
+    public WorkflowStepValidator(List<String> inputs, List<String> outputs, List<String> requiredPlugins, TimeValue timeout) {
         this.inputs = inputs;
         this.outputs = outputs;
         this.requiredPlugins = requiredPlugins;
@@ -59,7 +66,7 @@ public class WorkflowStepValidator {
         List<String> parsedInputs = new ArrayList<>();
         List<String> parsedOutputs = new ArrayList<>();
         List<String> requiredPlugins = new ArrayList<>();
-        String timeout = null;
+        TimeValue timeout = null;
 
         ensureExpectedToken(XContentParser.Token.START_OBJECT, parser.currentToken(), parser);
         while (parser.nextToken() != XContentParser.Token.END_OBJECT) {
@@ -85,7 +92,15 @@ public class WorkflowStepValidator {
                     }
                     break;
                 case TIMEOUT:
-                    timeout = parser.text();
+                    try {
+                        timeout = TimeValue.parseTimeValue(parser.text(), TIMEOUT);
+                    } catch (IllegalArgumentException e) {
+                        logger.error("Failed to parse TIMEOUT value for field [" + fieldName + "]", e);
+                        throw new FlowFrameworkException(
+                            "Failed to parse workflow-step.json file for field [" + fieldName + "]",
+                            RestStatus.INTERNAL_SERVER_ERROR
+                        );
+                    }
                     break;
                 default:
                     throw new IOException("Unable to parse field [" + fieldName + "] in a WorkflowStepValidator object.");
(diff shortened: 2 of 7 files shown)


src/main/java/org/opensearch/flowframework/workflow/WorkflowProcessSorter.java
```diff
@@ -118,20 +120,28 @@ public List<ProcessNode> sortProcessNodes(Workflow workflow, String workflowId)
...
-                node.previousNodeInputs(),
-                data,
-                predecessorNodes,
-                threadPool,
-                nodeTimeout
-            );
-            idToNodeMap.put(processNode.id(), processNode);
-            nodes.add(processNode);
-        }
+            try {
```
dbwiddis, line 123: This is a second-level of passing up an IOException reading a json on the classpath. We should handle this exception where it's read rather than passing it up.

src/main/java/org/opensearch/flowframework/workflow/WorkflowProcessSorter.java
```diff
@@ -244,11 +254,17 @@ public void validateGraph(List<ProcessNode> processNodes, WorkflowValidator vali
                 );
             }
         }
+    }
 
+    private WorkflowValidator readWorkflowValidator() throws IOException {
+        return WorkflowValidator.parse("mappings/workflow-steps.json");
```
dbwiddis, line 260: Let's handle the exception here (and rethrow an unchecked FlowFrameworkException) here rather than throwing a checked IOException that needs handling far from where the issue happens.

Let's also make this package-private and write a test method to validate this parsing so a user. will never see this error. 🤞

src/main/java/org/opensearch/flowframework/workflow/WorkflowProcessSorter.java
```diff
@@ -244,11 +254,17 @@ public void validateGraph(List<ProcessNode> processNodes, WorkflowValidator vali
...
     }
 
-    private TimeValue parseTimeout(WorkflowNode node) {
-        String timeoutValue = (String) node.userInputs().getOrDefault(NODE_TIMEOUT_FIELD, NODE_TIMEOUT_DEFAULT_VALUE);
+    private TimeValue parseTimeout(WorkflowNode node) throws IOException {
+        WorkflowValidator validator = readWorkflowValidator();
+        String nodeTimeoutValue = Optional.ofNullable(validator.getWorkflowStepValidators().get(node.type()).getTimeout())
+            .orElse(NODE_TIMEOUT_DEFAULT_VALUE);
```
dbwiddis, line 266: I really like this bit of code for handling the possible null I mentioned needing handling in an earlier comment.

That said, this implies we can/should leave the timeout value blank in the json if we want the default, but you've added this same default (10s) to every node, so it's a bit inconsistent.

src/main/java/org/opensearch/flowframework/workflow/WorkflowProcessSorter.java
```diff
@@ -244,11 +254,17 @@ public void validateGraph(List<ProcessNode> processNodes, WorkflowValidator vali
...
+        String timeoutValue = (String) node.userInputs().getOrDefault(NODE_TIMEOUT_FIELD, nodeTimeoutValue);
```
dbwiddis, line 267: Part of me thinks we should avoid calculating the default if we don't need it.

The other part of me thinks that mot of the time the user won't provide an override and we will use the default, and working around it is premature optimization.  So I guess this is all fine as is.

dbwiddis (commented):
Nice code, and I learned something!   Two general comments:
 - Let's use `TimeValue` rather than `String` as early as we can in parsing
 - Let's handle exceptions as early as we can in parsing

jackiehanyang:
> Nice code, and I learned something! Two general comments:
> 
> * Let's use `TimeValue` rather than `String` as early as we can in parsing
> * Let's handle exceptions as early as we can in parsing

Thanks for the quick review! Yeah, I immediately regretted the decision not to handle exceptions in parsing right after I sent out this PR and closed my laptop. But since I had already closed my laptop, I wanted to wait until Tuesday to address this. I will send out a new revision soon to address it. :)

src/main/java/org/opensearch/flowframework/model/WorkflowStepValidator.java
```diff
@@ -78,11 +84,14 @@ public static WorkflowStepValidator parse(XContentParser parser) throws IOExcept
                         requiredPlugins.add(parser.text());
                     }
                     break;
+                case TIMEOUT:
+                    timeout = parser.text();
```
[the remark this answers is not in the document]
jackiehanyang: I prefer to leave `timeout` as a `String` value, and leave it as `timeout = parser.text()` for the following reasons:

1. `TimeValue.parseTimeValue(...)` throws NPE if `parser.text()` is null - https://github.com/opensearch-project/OpenSearch/blob/71d7e4daf7dc05c9598d447e4424072e76de79d1/libs/common/src/main/java/org/opensearch/common/unit/TimeValue.java#L374-L375. But a String type allows it to be null.
2. The reason we allow it to be null is that we want to handle parsing errors when we parse it, rather than in the actual parser. If we handle it in the actual parser, we would need to catch every switch case. However, if we handle it when we call the parser, we can catch all parsing errors at once. For example, we don't throw exception here, instead we do try and catch here - https://github.com/opensearch-project/flow-framework/blob/main/src/main/java/org/opensearch/flowframework/workflow/WorkflowProcessSorter.java#L143-L144
3. Since the TIMEOUT in the workflow-step.json file is part of our own configuration, it shouldn't be null. However, just in case it is, we assign a default value in the code to prevent failing the entire workflow. As it's not a user-provided field, it makes no sense for the workflow to fail just because it's null. Therefore, we catch it and assign a default value in every step where possible.

jackiehanyang:
> Nice code, and I learned something! Two general comments:
> 
> * Let's use `TimeValue` rather than `String` as early as we can in parsing
> * Let's handle exceptions as early as we can in parsing

I prefer to leave `timeout` as a `String` value, and leave it as `timeout = parser.text()` for the following reasons:

1. `TimeValue.parseTimeValue(...)` throws NPE if `parser.text()` is null - https://github.com/opensearch-project/OpenSearch/blob/71d7e4daf7dc05c9598d447e4424072e76de79d1/libs/common/src/main/java/org/opensearch/common/unit/TimeValue.java#L374-L375. But a String type allows it to be null.
2. The reason we allow it to be null is that we want to handle parsing errors when we parse it, rather than in the actual parser. If we handle it in the actual parser, we would need to catch every switch case. However, if we handle it when we call the parser, we can catch all parsing errors at once. For example, we don't throw exception here, instead we do try and catch here - https://github.com/opensearch-project/flow-framework/blob/main/src/main/java/org/opensearch/flowframework/workflow/WorkflowProcessSorter.java#L143-L144
3. Since the TIMEOUT in the workflow-step.json file is part of our own configuration, it shouldn't be null. However, just in case it is, we assign a default value in the code to prevent failing the entire workflow. As it's not a user-provided field, it makes no sense for the workflow to fail just because it's null. Therefore, we catch it and assign a default value in every step where possible.

src/main/java/org/opensearch/flowframework/workflow/WorkflowProcessSorter.java
jackiehanyang: As mentioned in this issue - https://github.com/opensearch-project/flow-framework/issues/249, I added 15s for Deploy model step, and 10s for the rest steps. For the new revision, according to your [comment above](https://github.com/opensearch-project/flow-framework/pull/317#discussion_r1435777193) I changed the timeout from 10s to 60s for `register_local_model` step :)

src/main/resources/mappings/workflow-steps.json
```diff
@@ -73,7 +77,8 @@
         ],
         "required_plugins":[
             "opensearch-ml"
-        ]
+        ],
+        "timeout": "10s"
```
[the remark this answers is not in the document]
jackiehanyang: Changed timeout value for `register_local_model` to `60s`

src/main/java/org/opensearch/flowframework/workflow/WorkflowProcessSorter.java
jackiehanyang: Address this comment by moving the try catch logic to where we read the json file in the latest revision

src/main/java/org/opensearch/flowframework/workflow/WorkflowProcessSorter.java
jackiehanyang: Added a new test case to cover all possible ways we parse this timeout value.

src/main/java/org/opensearch/flowframework/model/WorkflowStepValidator.java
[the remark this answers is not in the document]
dbwiddis: For point 1 (and 3), `parser.text()` should never be null.  By the time we get here we have explicitly encountered a field called `timeout`.  Per your 3rd point we control this JSON so we should never put null there.   If we want a null value we can just completely omit the `timeout` field name.

I'm not sure what you mean by point 2.   Are we collecting all of the errors and returning a complete list of parse failures?  If so there's some benefit to waiting.  If not (we stop on first failure) then we're just wasting time moving past the first failure.

src/main/java/org/opensearch/flowframework/workflow/WorkflowProcessSorter.java
dbwiddis: While writing docs for the settings, I see a timeout setting that we have that seems to only apply to the Create Workflow API.  This seems silly as we already have an OpenSearch timeout for that.   But we probably should have a setting for the default workflow step timeout (or use that exact setting)... unless we are using it and I'm confused!  :|

src/main/java/org/opensearch/flowframework/model/WorkflowStepValidator.java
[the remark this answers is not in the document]
jackiehanyang: I believe there's been some misunderstanding. Please check the new test cases I have added in `src/test/java/org/opensearch/flowframework/workflow/WorkflowProcessSorterTests.java`. Hope this test case explains what we want to do clearly.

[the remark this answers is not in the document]
dbwiddis: > I believe there's been some misunderstanding.

Probably me not being clear enough.   I'm 100% on board with your test cases and they should remain the same.  I'm okay with the value of the variable `timeout` remaining null and being handled as you currently do (although I am suggesting you make it a null `TimeValue` and not a null `String`).

The point I'm trying to make is associated with this code:
```java
   case TIMEOUT:
                    timeout = parser.text();
```
We only enter this block if we have encountered that `TIMEOUT` key in the JSON.  If it's omitted, we never even try to parse a value and the variable `timeout` remains null.

The only way for `parser.text()` to be null here is if we explicitly include it in our JSON like this:
```json
    {
        "required_plugins":[
            "opensearch-ml"
        ],
        "timeout": null
    },
```
(or `"timeout": false` or `"timeout": 42` or other numeric value or anything that's not a string).

That should not happen if we define our own mapping. 

Which basically means we're passing around a checked exception (and the requirement to handle it) that shouldn't ever actually occur.

It's not that big a deal, but I don't see why we can't immediately parse it to a `TimeValue` right here.  If there's an error, it's easy to debug.  If we want it null we never get to that line and we have a `null` TimeValue we can ignore instead of checking for a null string and then parsing it later and then figuring out what to do with it, with a stack trace that doesn't necessarily even point out where the error occurred.

src/main/java/org/opensearch/flowframework/model/WorkflowNode.java
```diff
@@ -50,7 +50,7 @@ public class WorkflowNode implements ToXContentObject {
     /** The field defining the timeout value for this node */
     public static final String NODE_TIMEOUT_FIELD = "node_timeout";
     /** The default timeout value if the template doesn't override it */
-    public static final String NODE_TIMEOUT_DEFAULT_VALUE = "15s";
+    public static final String NODE_TIMEOUT_DEFAULT_VALUE = "10s";
```
dbwiddis, line 53: Mentioned this as an aside [here](https://github.com/opensearch-project/flow-framework/pull/317/files#r1436647786), this is OK for this PR but we might want to consider using one of the settings for this instead of hardcoding it.

src/main/java/org/opensearch/flowframework/workflow/WorkflowProcessSorter.java
```diff
@@ -89,7 +91,7 @@ public WorkflowProcessSorter(
      * @param workflowId The workflowId associated with the step
      * @return A list of Process Nodes sorted topologically.  All predecessors of any node will occur prior to it in the list.
      */
-    public List<ProcessNode> sortProcessNodes(Workflow workflow, String workflowId) {
+    public List<ProcessNode> sortProcessNodes(Workflow workflow, String workflowId) throws IOException {
```
dbwiddis, line 94: I don't think we need to throw this exception here?  (If we did it needs a javadoc.)

The only possible exceptions are unchecked `IllegalArgumentException` or `FlowFrameworkExceptions` from your parsing of the time value here.

However, this brings up another question how are we handling any Exception thrown here?  I wondered why adding this IOE didn't require you to handle it upstream.

It turns out that it ends up in the ActionListener in the `ProvisionWorkflowTransportAction` where it'll silently kill that thread (and likely end up with a timeout with an incompleted future.). So a parsing exception will get completely lost.   Which in theory isn't bad if you never make a typo in your JSON... or you at least have unit testing of that JSON.

But we're not unit-testing that parsing in any way. As a test, I just changed the register local model timeout to a completely invalid value ("FAIL") and all unit tests passed.

Waiting to parse until this "sort" method means you are only testing the specific workflow steps being sorted, so in order to thoroughly vaildate our `workflowsteps.json` file we would need to create a workflow that calls every single step.  

I still think "parsing" when you read the value and not in a "sorting" method is the right answer. You read the file once and find any parsing failures then.  Adding `TimeValue.parseTimeValue(parser.text(), TIMEOUT)` in the WorkflowStepValidator line 89 as I proposed will catch that invalid parsing error:
```
  2> java.lang.IllegalArgumentException: failed to parse setting [timeout] with value [FAIL] as a time value: unit is missing or unrecognized
        at __randomizedtesting.SeedInfo.seed([AF75607C3973607E:EC2D9DC06707F913]:0)
        at org.opensearch.common.unit.TimeValue.parseTimeValue(TimeValue.java:406)
        at org.opensearch.common.unit.TimeValue.parseTimeValue(TimeValue.java:376)
        at org.opensearch.flowframework.model.WorkflowStepValidator.parse(WorkflowStepValidator.java:90)
```

Whatever solution you choose, you need to unit test reading that template to make sure typos don't slip through.

src/main/java/org/opensearch/flowframework/workflow/WorkflowProcessSorter.java
```diff
@@ -141,7 +142,7 @@ public List<ProcessNode> sortProcessNodes(Workflow workflow, String workflowId)
      * @throws Exception if validation fails
      */
     public void validate(List<ProcessNode> processNodes) throws Exception {
-        WorkflowValidator validator = WorkflowValidator.parse("mappings/workflow-steps.json");
+        WorkflowValidator validator = readWorkflowValidator(processNodes.get(0).id());
```
dbwiddis, line 145: This id is just the workflow step ID of the first node, not the workflow ID that method is expecting, so If we don't yet have a workflow ID I'd suggest just passing "new workflow" or something similar here.

src/main/java/org/opensearch/flowframework/workflow/WorkflowProcessSorter.java
```diff
@@ -244,11 +245,32 @@ public void validateGraph(List<ProcessNode> processNodes, WorkflowValidator vali
...
+        }
     }
 
-    private TimeValue parseTimeout(WorkflowNode node) {
-        String timeoutValue = (String) node.userInputs().getOrDefault(NODE_TIMEOUT_FIELD, NODE_TIMEOUT_DEFAULT_VALUE);
+    /**
+     * A method for parsing workflow timeout value.
+     * The value could be parsed from node NODE_TIMEOUT_FIELD, the timeout field in workflow-step.json,
+     * or the default NODE_TIMEOUT_DEFAULT_VALUE
+     * @param node the workflow nde
```
dbwiddis, line 266: Typo :)

src/main/java/org/opensearch/flowframework/workflow/WorkflowProcessSorter.java
```diff
@@ -244,11 +245,32 @@ public void validateGraph(List<ProcessNode> processNodes, WorkflowValidator vali
...
+     * @return the timeout value
+     */
+    protected TimeValue parseTimeout(WorkflowNode node) {
+        WorkflowValidator validator = readWorkflowValidator(node.id());
+        String nodeTimeoutValue = Optional.ofNullable(validator.getWorkflowStepValidators().get(node.type()).getTimeout())
+            .orElse(NODE_TIMEOUT_DEFAULT_VALUE);
+        String timeoutValue = (String) node.userInputs().getOrDefault(NODE_TIMEOUT_FIELD, nodeTimeoutValue);
         String fieldName = String.join(".", node.id(), USER_INPUTS_FIELD, NODE_TIMEOUT_FIELD);
         TimeValue timeValue = TimeValue.parseTimeValue(timeoutValue, fieldName);
         if (timeValue.millis() < 0) {
```
dbwiddis, line 276: I'm OK with requiring positive values.  Note that the `TimeValue` class does provide an explicit exception for -1 as a 'magic number' which could potentially be allowed as a special-case "no timeout" value (if documented).

However, as strongly suggested elsewhere I think this parsing needs to be moved to the file reading code and properly unit tested.

dbwiddis (changes requested):

src/main/java/org/opensearch/flowframework/model/WorkflowStepValidator.java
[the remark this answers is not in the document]
jackiehanyang: I don't like the approach of parsing the timeout value as `TimeValue` in `WorkflowStepValidator` immediately because we currently parse the timeout field in `WorkflowNode` as `String` type. This will cause inconsistency and make the `parseTimeout(WorkflowNode node)` method awkward. Since the `timeout` attribute in `WorkflowValidator` is now a `TimeValue type`, we will need to cast it back to a String type to work with the String type `timeout` field in `WorkflowNode`. This doesn't have any functional impact, and we are catching every possible null and exception though.

But I do agree it's better to parse any time-related value to a `TimeValue` immediately when we can. So I made this change in the latest revision. Maybe we should open an issue to straighten out all timeout/time-related fields across the project to ensure they are all parsed into a TimeValue type instead of Strings.

src/main/java/org/opensearch/flowframework/model/WorkflowNode.java
```diff
@@ -50,7 +52,7 @@ public class WorkflowNode implements ToXContentObject {
     /** The field defining the timeout value for this node */
     public static final String NODE_TIMEOUT_FIELD = "node_timeout";
     /** The default timeout value if the template doesn't override it */
-    public static final String NODE_TIMEOUT_DEFAULT_VALUE = "15s";
+    public static final TimeValue NODE_TIMEOUT_DEFAULT_VALUE = new TimeValue(10, SECONDS);
```
dbwiddis, line 55: This is good for this PR.  I'm thinking I'd like to repurpose the `WORKFLOW_REQUEST_TIMEOUT` setting to be used here (and it's already a `TimeValue` too)... but we can handle that later.

src/main/java/org/opensearch/flowframework/model/WorkflowStepValidator.java
```diff
@@ -78,11 +91,22 @@ public static WorkflowStepValidator parse(XContentParser parser) throws IOExcept
                         requiredPlugins.add(parser.text());
                     }
                     break;
+                case TIMEOUT:
+                    try {
+                        timeout = TimeValue.parseTimeValue(parser.text(), TIMEOUT);
+                    } catch (IllegalArgumentException e) {
+                        logger.error("Failed to parse TIMEOUT value for field [" + fieldName + "]", e);
```
dbwiddis, line 98: Nit: use substitutions instead of concatenation.
```suggestion
                        logger.error("Failed to parse TIMEOUT value for field [{}]", fieldName, e);
```

dbwiddis (approved):
LGTM with a nit and few outstanding comments from prior review.

Issue #315, opened 2023-12-22
[DISCUSS] Open Questions for docs / APIs


dbwiddis:
While drafting the API documentation (#311) I encountered a few questions that I need team feedback on.

1. Is `use_case` an enum or user-defined field?  I think we originally identified it as an enum for use by the front-end.  Whatever it is, we need to add code for it and document it.
   - I've inserted some words to this effect.
2. I need to add params to the documentation. 
   - I've done this
3. I think given our limited number of steps for 2.12 the "Sample Workflow Steps" (#314) may be sufficient for workflow step documentation.  But we really need to improve that in later releases. Open to thoughts on:
   - Do we need individual API docs for each workflow step or can we just reference ML Commons API, or some mix?
      - I've linked to individual APIs for all cases and documented differences
   - Do we need an API that returns all the workflow steps and required/optional inputs (essentially the [content of this json file](https://github.com/opensearch-project/flow-framework/blob/main/src/main/resources/mappings/workflow-steps.json) as an API)?
      - I think we do
      - https://github.com/opensearch-project/flow-framework/issues/364

amitgalitz:
I think `use_case` should be an enum based on pre-defined cases, we still don't really have those yet though I think.

I think we can add links to the API that the steps will be using

PR #325, opened 2023-12-27
Use current timestamps for registering agent

### LINKED ISSUE #324: [BUG] Register Agent step incorrectly requires current timestamps
### What is the bug?

The Register Agent step includes "created time" and "last updated time" as required inputs.

Register an agent and have to provide those inputs.

Leave the inputs blank and have them auto-populated by code for the current `time.getEpochSecond()`.

https://github.com/opensearch-project/ml-commons/blob/50788de0bcf7a9509da6d2dabb23aab59d5a843d/plugin/src/test/java/org/opensearch/ml/rest/RegisterAgentTransportActionTests.java#L79-L99

dbwiddis:
### Description

Registering an agent normally uses the current time for creation and last updated. These fields should not be provided by a user.

There is a mismatch between the required fields in `workflowstep.json` and the Required/Optional fields in the Workflow Step itself (and no null handling for omission of some fields). This should probably be reviewed in a bit more detail.

### Issues Resolved

Fixes #324

### DIFF
--- a/src/main/resources/mappings/workflow-steps.json
+++ b/src/main/resources/mappings/workflow-steps.json
@@ -141,8 +141,6 @@
             "tools",
             "parameters",
             "memory",
-            "created_time",
-            "last_updated_time",
             "app_type"
         ],
         "outputs":[
--- a/src/main/resources/mappings/workflow-steps.json
+++ b/src/main/resources/mappings/workflow-steps.json
@@ -137,11 +137,7 @@
     "register_agent": {
         "inputs":[
             "name",
-            "type",
-            "tools",
-            "parameters",
-            "memory",
-            "app_type"
+            "type"
         ],
         "outputs":[
             "agent_id"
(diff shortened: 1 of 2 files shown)


src/main/resources/mappings/workflow-steps.json
```diff
@@ -137,13 +137,7 @@
     "register_agent": {
         "inputs":[
             "name",
-            "type",
-            "tools",
-            "parameters",
-            "memory",
-            "created_time",
-            "last_updated_time",
-            "app_type"
```
ohltyler, line 146: Catching up on these JSON files. Where do we persist the set of optional fields, besides `optionalKeys` in the `WorkflowStep` itself?

ohltyler (approved):

src/main/resources/mappings/workflow-steps.json
dbwiddis: Good question. There was talk of putting optional fields in this JSON file. There was also talk of getting rid of this file and creating an API to return all this stuff programmatically.  We have done neither.  

For now, though, this file is used for validation and these fields need to go.

We do need to figure out how else to document these.  See point 3 in https://github.com/opensearch-project/flow-framework/issues/315

Issue #345, opened 2024-01-02
[FEATURE] Complete Register Local Model Implementation


dbwiddis:
### Is your feature request related to a problem?

The sparse model tool requires the Register Local Model step to be completed.  It is currently missing some changes made only to the Register Remote Model step such as updating state index, and handling deploy=true.  Its integration tests are currently disabled due to these shortcomings.

### What solution would you like?
 - [ ] Update the step to record its registered model in the state index
 - [ ] Update the step to handle the boolean deploy = true parameter (see https://github.com/opensearch-project/flow-framework/pull/340 for necessary steps)
 - [ ] Enable its integration test and add a new one for the deploy=true variant

PR #350, opened 2024-01-02
Adds deploy model flag support for local model registration, fixes integration tests

joshpalis:
### Description
This PR achieves multiple things : 

- Enables register local model and deploy model integration test
- Adds support for setting the deploy flag to true for local model registration
- Modifies the `AbstractRetryableWorkflowStep` `retryableGetMLTask` to accept an action listener and moves completing the future/ simulating model deployment within the step classes themselves
- Fixes the `DeprovisionWorkflowTransportAction` so that the deprovision sequence is ascertained from the resources created, rather than the provision sequence
- Adds a new 1 step template to test out register local model and deployment 
- Mitigates ML Circuit breaker issues due to memory intensive tasks (due to multiple local / remote model registration tests). These tests have been reduced to just 2 tests for local and remote model registration. Github runners have limited memory/storage

Resources Created output for the single step register local model template : 
```
 curl -i -XGET "localhost:9200/_plugins/_flow_framework/workflow/Cs_q0YwBbQBv2_MrXO2l/_status?all=true&pretty"
HTTP/1.1 200 OK
content-length: 603

{
  "workflow_id" : "Cs_q0YwBbQBv2_MrXO2l",
  "state" : "COMPLETED",
  "provisioning_progress" : "DONE",
  "provision_start_time" : 1704328870019,
  "provision_end_time" : 1704328885091,
  "resources_created" : [
    {
      "workflow_step_name" : "register_local_model",
      "workflow_step_id" : "workflow_step_1",
      "resource_type" : "model_id",
      "resource_id" : "Dc_q0YwBbQBv2_MruO2z"
    },
    {
      "workflow_step_name" : "deploy_model",
      "workflow_step_id" : "workflow_step_1-deploy",
      "resource_type" : "model_id",
      "resource_id" : "Dc_q0YwBbQBv2_MruO2z"
    }
  ]
}
```

### Issues Resolved
Fixes https://github.com/opensearch-project/flow-framework/issues/345


src/test/java/org/opensearch/flowframework/rest/FlowFrameworkRestApiIT.java
```diff
@@ -68,8 +69,6 @@ public void testSearchWorkflows() throws Exception {
     }
 
     public void testCreateAndProvisionLocalModelWorkflow() throws Exception {
-        /*- Local model registration is not yet fully complete.  Commenting this test out until it is.
-         * https://github.com/opensearch-project/flow-framework/issues/305
 
         // Using a 3 step template to create a model group, register a remote model and deploy model
```
dbwiddis, line 73: Nit: change comment here to 'local'

dbwiddis (commented):
Let's make sure we keep the "retry taskId until we get a resource ID" and "update resources in index" functionality in appropriate classes.

joshpalis:
Some flaky test failures for local model registration, curious if deploying multiple local models is the root cause of this. I'll check if undeploying the local model before proceeding with additional tests would resolve this: 
```
 Caused by: org.opensearch.flowframework.exception.FlowFrameworkException: Deploy model failed with error : {"XNnALvoTRJO3CUYOHLYvUA":"Memory Circuit Breaker is open, please check your resources!"}
»  	at org.opensearch.flowframework.workflow.AbstractRetryableWorkflowStep.lambda$retryableGetMlTask$5(AbstractRetryableWorkflowStep.java:161) ~[?:?]
»  	at org.opensearch.core.action.ActionListener$1.onResponse(ActionListener.java:82) ~[opensearch-core-3.0.0-SNAPSHOT.jar:3.0.0-SNAPSHOT]
»  	at org.opensearch.ml.client.MachineLearningNodeClient.lambda$getMLTaskResponseActionListener$9(MachineLearningNodeClient.java:302) ~[?:?]
»  	at org.opensearch.core.action.ActionListener$1.onResponse(ActionListener.java:82) ~[opensearch-core-3.0.0-SNAPSHOT.jar:3.0.0-SNAPSHOT]
»  	at org.opensearch.ml.client.MachineLearningNodeClient.lambda$wrapActionListener$18(MachineLearningNodeClient.java:374) ~[?:?]
»  	at org.opensearch.core.action.ActionListener$1.onResponse(ActionListener.java:82) ~[opensearch-core-3.0.0-SNAPSHOT.jar:3.0.0-SNAPSHOT]
»  	at org.opensearch.action.support.TransportAction$1.onResponse(TransportAction.java:113) ~[opensearch-3.0.0-SNAPSHOT.jar:3.0.0-SNAPSHOT]
»  	at org.opensearch.action.support.TransportAction$1.onResponse(TransportAction.java:107) ~[opensearch-3.0.0-SNAPSHOT.jar:3.0.0-SNAPSHOT]
»  	at org.opensearch.ml.action.tasks.GetTaskTransportAction.lambda$doExecute$0(GetTaskTransportAction.java:67) ~[?:?]
```

src/main/java/org/opensearch/flowframework/workflow/AbstractRetryableWorkflowStep.java
```diff
@@ -96,16 +98,50 @@ protected void retryableGetMlTask(
                                     id,
                                     ActionListener.wrap(updateResponse -> {
                                         logger.info("successfully updated resources created in state index: {}", updateResponse.getIndex());
-                                        future.complete(
-                                            new WorkflowData(
-                                                Map.ofEntries(
-                                                    Map.entry(resourceName, id),
-                                                    Map.entry(REGISTER_MODEL_STATUS, response.getState().name())
-                                                ),
+                                        if (workflowStep == "Local model registration" && Boolean.TRUE.equals(deploy)) {
```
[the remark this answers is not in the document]
joshpalis: Perhaps a simpler way would to pass an action listener as part of the retryableGetMlTask. This would allow us to handle completing the future / simulating model deployment within the step class after the action listener's `onResponse` is invoked.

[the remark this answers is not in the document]
joshpalis: I have added the modifications we discussed in the latest commit

joshpalis:
Still observing flaky test failures, to mitigate this I have set the native memory threshold from 90 to 100 to prevent the circuit breaker from opening ([Documentation](https://opensearch.org/docs/latest/ml-commons-plugin/cluster-settings/#set-native-memory-threshold)), however the logs still show that the breaker is open. Will continue to look into this : 
```
ERROR][o.o.f.t.ProvisionWorkflowTransportAction] [integTest-0] Provisioning failed for workflow: LK-n0YwBmExBUdkP7lAh
�  java.util.concurrent.CompletionException: java.util.concurrent.ExecutionException: org.opensearch.flowframework.exception.FlowFrameworkException: Memory Circuit Breaker is open, please check your resources!
�  	at java.base/java.util.concurrent.CompletableFuture.reportJoin(CompletableFuture.java:413) ~[?:?]
�  	at java.base/java.util.concurrent.CompletableFuture.join(CompletableFuture.java:2118) ~[?:?]
�  	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596) ~[?:?]
```

src/main/java/org/opensearch/flowframework/transport/DeprovisionWorkflowTransportAction.java
```diff
@@ -151,64 +127,43 @@ protected void doExecute(Task task, WorkflowRequest request, ActionListener<Work
...
-                String deprovisionStep = getDeprovisionStepByWorkflowStep(stepName);
-                // Unimplemented steps presently return null, so skip
-                if (deprovisionStep == null) {
-                    return null;
-                }
-                // New ID is old ID with deprovision added
-                String deprovisionStepId = pn.id() + DEPROVISION_SUFFIX;
-                return new ProcessNode(
+        List<ProcessNode> deprovisionProcessSequence = new ArrayList<>();
+        for (Map.Entry<String, ResourceCreated> resource : resourceMap.entrySet()) {
```
dbwiddis, line 138: The entryset's order with this map is nondeterministic.  The order of deprovisioning is somewhat robust to failed attempts with retries, but we can reduce those retries by deprovisioning in the reverse order we provisioned (see `Collections.reverse()` at line 168 which reversed the process node sequence).

However we do know the order of entries in the resources created list is the creation order so we can preserve that. We just need to make sure we preserve map insertion order (see next comment where we create the map).

src/main/java/org/opensearch/flowframework/transport/DeprovisionWorkflowTransportAction.java
```diff
@@ -103,46 +100,25 @@ public DeprovisionWorkflowTransportAction(
...
         // Stash thread context to interact with system index
         try (ThreadContext.StoredContext context = client.threadPool().getThreadContext().stashContext()) {
-            client.get(getRequest, ActionListener.wrap(response -> {
+            client.execute(GetWorkflowStateAction.INSTANCE, getStateRequest, ActionListener.wrap(response -> {
                 context.restore();
+                // Get a map of step id to created resources
+                final Map<String, ResourceCreated> resourceMap = response.getWorkflowState()
+                    .resourcesCreated()
+                    .stream()
+                    .collect(Collectors.toMap(ResourceCreated::workflowStepId, Function.identity()));
```
dbwiddis, line 114: Preserve the order here:
```suggestion
                    .collect(Collectors.toMap(ResourceCreated::workflowStepId, Function.identity(), (u, m) -> u, LinkedHashMap::new));
```

dbwiddis (commented):
LGTM with one strong suggestion for improvement.

src/main/java/org/opensearch/flowframework/transport/DeprovisionWorkflowTransportAction.java
```diff
@@ -151,64 +127,43 @@ protected void doExecute(Task task, WorkflowRequest request, ActionListener<Work
...
-                if (deprovisionStep == null) {
-                    return null;
-                }
-                // New ID is old ID with deprovision added
-                String deprovisionStepId = pn.id() + DEPROVISION_SUFFIX;
-                return new ProcessNode(
+        List<ProcessNode> deprovisionProcessSequence = new ArrayList<>();
+        for (Map.Entry<String, ResourceCreated> resource : resourceMap.entrySet()) {
+            String workflowStepId = resource.getKey();
+            ResourceCreated resourceCreated = resource.getValue();
```
dbwiddis, line 140: Actually on second look, I think you're losing a deprovision step here.   Your example in the comments shows the same step id and resource but different step names.  That results in only one entry in the map.  But you need to both undeploy and delete (unregister) the model ID.

src/main/java/org/opensearch/flowframework/transport/DeprovisionWorkflowTransportAction.java
dbwiddis, line 114: Consider appending step name to step ID with some delimiter, to make the pair unique?

src/main/java/org/opensearch/flowframework/transport/DeprovisionWorkflowTransportAction.java
joshpalis: Ah I forgot to update the example in the description, but the step names are modified when the local/remote model registration also deploys. 

Here's an example from running the register remote model template with the deploy flag, the deploy step has a workflow step ID of `workflow_step_2-deploy` : 
```
{
  "workflow_id" : "BM_n0YwBbQBv2_Mr8O06",
  "state" : "COMPLETED",
  "provisioning_progress" : "DONE",
  "provision_start_time" : 1704328720539,
  "provision_end_time" : 1704328721409,
  "resources_created" : [
    {
      "workflow_step_name" : "create_connector",
      "workflow_step_id" : "workflow_step_1",
      "resource_type" : "connector_id",
      "resource_id" : "Bc_o0YwBbQBv2_Mrce3E"
    },
    {
      "workflow_step_name" : "register_remote_model",
      "workflow_step_id" : "workflow_step_2",
      "resource_type" : "model_id",
      "resource_id" : "CM_o0YwBbQBv2_Mrc-2e"
    },
    {
      "workflow_step_name" : "deploy_model",
      "workflow_step_id" : "workflow_step_2-deploy",
      "resource_type" : "model_id",
      "resource_id" : "CM_o0YwBbQBv2_Mrc-2e"
    }
  ]
}
```

src/main/java/org/opensearch/flowframework/transport/DeprovisionWorkflowTransportAction.java
joshpalis: Answered above

src/main/java/org/opensearch/flowframework/workflow/AbstractRetryableWorkflowStep.java
```diff
@@ -71,13 +69,15 @@ protected AbstractRetryableWorkflowStep(
...
+     * @param mlTaskListener the ML Task Listener
      */
     protected void retryableGetMlTask(
         String workflowId,
         String nodeId,
         CompletableFuture<WorkflowData> future,
         String taskId,
-        String workflowStep
+        String workflowStep,
+        ActionListener<MLTask> mlTaskListener
```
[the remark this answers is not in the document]
joshpalis: Ah thats a good point, it's no longer used, I'll go ahead and remove this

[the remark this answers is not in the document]
joshpalis: We currently use this to check if the future is done for every retry and to cancel the future if needed. I think its ok to keep this here

src/main/java/org/opensearch/flowframework/transport/DeprovisionWorkflowTransportAction.java
dbwiddis: Ah, I see! This should be fine unless someone has similar steps named "foo" and "foo-deploy" :)

src/main/java/org/opensearch/flowframework/transport/DeprovisionWorkflowTransportAction.java
dbwiddis: > Answered above

Yeah, I see that the map is OK, but looking at it all again, it seems to be overcomplicated.

We already have `response.getWorkflowState().resourcesCreated()` that we are using to create this map (it ends up as the `values()` after this change).

But later we are iterating the entry set just to extract that same iteration of `resourcesCreated`, which already has all the pieces we want (ID without having to modify it, step name/type, and resource id).

joshpalis: I have removed the resource map from the deprovision workflow action and instead passed the list of resource created

src/main/java/org/opensearch/flowframework/workflow/RegisterLocalModelStep.java
```diff
@@ -176,10 +155,80 @@ public void onFailure(Exception e) {
...
+                        // Registered Model Resource has been updated
+                        String resourceName = getResourceByWorkflowStep(getName());
+                        String id = getResourceId(mlTask);
+
+                        if (Boolean.TRUE.equals(deploy)) {
+
+                            // Simulate Model deployment step and update resources created
+                            flowFrameworkIndicesHandler.updateResourceInStateIndex(
+                                currentNodeInputs.getWorkflowId(),
+                                currentNodeId + "-deploy",
```
owaiskazi19, line 187: Let's create a constant for the suffix.

src/test/java/org/opensearch/flowframework/FlowFrameworkRestTestCase.java
```diff
@@ -119,8 +119,19 @@ public void setUpSettings() throws Exception {
             );
             assertEquals(200, response.getStatusLine().getStatusCode());
 
+            // Set ML native memory threshold to 100 to avoid opening the circuit breaker during tests
+            response = TestHelpers.makeRequest(
+                client(),
+                "PUT",
+                "_cluster/settings",
+                null,
+                "{\"persistent\":{\"plugins.ml_commons.native_memory_threshold\":100}}",
```
owaiskazi19, line 128: Should this be `transient` setting for the cluster?

owaiskazi19 (commented):
Look good overall with minor comments.

src/main/java/org/opensearch/flowframework/workflow/RegisterLocalModelStep.java
dbwiddis: Actually, I'm not sure we need a suffix in the state if we're not using the map version anymore which previously had a key collision.

If we do keep a suffix I'd prefer a random one that is unlikely to match anything a user might have on a different step.

src/test/java/org/opensearch/flowframework/FlowFrameworkRestTestCase.java
joshpalis: It can be either, transient vs persistent really only matters during cluster restarts. In our case, we just spin up the cluster once and run all our tests before terminating the cluster

dbwiddis (approved):
LGTM.  Approving with a comment that I'm not sure an id-suffix is needed any more.  If it is, it's fine (or make it a constant or auto-generated).

src/main/java/org/opensearch/flowframework/workflow/RegisterLocalModelStep.java
joshpalis: That's a good point Dan, let me test out if the suffix is needed, if not then I'll remove this

src/main/java/org/opensearch/flowframework/workflow/AbstractRetryableWorkflowStep.java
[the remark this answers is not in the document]
joshpalis: Not sure if having another listener would work here since we may need to cancel the future

owaiskazi19 (approved):

owaiskazi19:
Did the CI miss 2 checks? We have 21 in total

PR #353, opened 2024-01-03
Get installed plugins from local node and add a timeout

### LINKED ISSUE #352: [BUG] Validation of installed plugins blocks forever
### What is the bug?

The check for installed plugins may block indefinitely on some clusters.

https://github.com/opensearch-project/flow-framework/blob/dd32d55a4d24f06980f051b0409368dbcc676202/src/main/java/org/opensearch/flowframework/workflow/WorkflowProcessSorter.java#L180

Create a 2-node cluster on macOS using main (3.0.0-SNAPSHOT) branches.

Validation fails after some timeout with 409 error identifying the cause.

```
"opensearch[node-1][transport_worker][T#4]" #47 [36355] daemon prio=5 os_prio=31 cpu=7494.81ms elapsed=12333.24s tid=0x000000012e55f200 nid=36355 waiting on condition  [0x00000001796c5000]
   java.lang.Thread.State: WAITING (parking)
	at jdk.internal.misc.Unsafe.park(java.base@21.0.1/Native Method)
	- parking to wait for  <0x00000007e91cb998> (a java.util.concurrent.CompletableFuture$Signaller)
	at java.util.concurrent.locks.LockSupport.park(java.base@21.0.1/LockSupport.java:221)
	at java.util.concurrent.CompletableFuture$Signaller.block(java.base@21.0.1/CompletableFuture.java:1864)
	at java.util.concurrent.ForkJoinPool.unmanagedBlock(java.base@21.0.1/ForkJoinPool.java:3780)
	at java.util.concurrent.ForkJoinPool.managedBlock(java.base@21.0.1/ForkJoinPool.java:3725)
	at java.util.concurrent.CompletableFuture.waitingGet(java.base@21.0.1/CompletableFuture.java:1898)
	at java.util.concurrent.CompletableFuture.get(java.base@21.0.1/CompletableFuture.java:2072)
	at org.opensearch.flowframework.workflow.WorkflowProcessSorter.validatePluginsInstalled(WorkflowProcessSorter.java:180)
	at org.opensearch.flowframework.workflow.WorkflowProcessSorter.validate(WorkflowProcessSorter.java:145)
	at org.opensearch.flowframework.transport.CreateWorkflowTransportAction.validateWorkflows(CreateWorkflowTransportAction.java:266)
	at org.opensearch.flowframework.transport.CreateWorkflowTransportAction.doExecute(CreateWorkflowTransportAction.java:102)
	at org.opensearch.flowframework.transport.CreateWorkflowTransportAction.doExecute(CreateWorkflowTransportAction.java:50)
	at org.opensearch.action.support.TransportAction$RequestFilterChain.proceed(TransportAction.java:218)
	at org.opensearch.action.support.TransportAction.execute(TransportAction.java:188)
	at org.opensearch.action.support.TransportAction.execute(TransportAction.java:107)
	at org.opensearch.client.node.NodeClient.executeLocally(NodeClient.java:110)
	at org.opensearch.client.node.NodeClient.doExecute(NodeClient.java:97)
```

dbwiddis:
### Description

Refactors the installed plugin validation check to more closely match `_cat/plugins`:
 - Only queries the local node state to get the localNodeId
 - Adds a timeout to prevent blocking synchronously

### Issues Resolved

Fixes #352 

### DIFF
--- a/src/main/java/org/opensearch/flowframework/workflow/WorkflowProcessSorter.java
+++ b/src/main/java/org/opensearch/flowframework/workflow/WorkflowProcessSorter.java
@@ -10,9 +10,9 @@ package org.opensearch.flowframework.workflow;
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
-import org.opensearch.action.admin.cluster.node.info.NodeInfo;
 import org.opensearch.action.admin.cluster.node.info.NodesInfoRequest;
 import org.opensearch.action.admin.cluster.node.info.PluginsAndModules;
+import org.opensearch.action.admin.cluster.state.ClusterStateRequest;
 import org.opensearch.client.Client;
 import org.opensearch.cluster.service.ClusterService;
 import org.opensearch.common.settings.Settings;
@@ -79,9 +79,9 @@ public class WorkflowProcessSorter {
         this.workflowStepFactory = workflowStepFactory;
         this.threadPool = threadPool;
         this.maxWorkflowSteps = MAX_WORKFLOW_STEPS.get(settings);
+        clusterService.getClusterSettings().addSettingsUpdateConsumer(MAX_WORKFLOW_STEPS, it -> maxWorkflowSteps = it);
         this.clusterService = clusterService;
         this.client = client;
-        clusterService.getClusterSettings().addSettingsUpdateConsumer(MAX_WORKFLOW_STEPS, it -> maxWorkflowSteps = it);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/main/java/org/opensearch/flowframework/workflow/WorkflowProcessSorter.java
```diff
@@ -153,31 +153,45 @@ public void validate(List<ProcessNode> processNodes) throws Exception {
...
-            installedPluginsFuture.complete(installedPlugins);
-
-        }, exception -> {
-            logger.error("Failed to retrieve installed plugins");
-            installedPluginsFuture.completeExceptionally(exception);
+        final CompletableFuture<List<String>> installedPluginsFuture = new CompletableFuture<>();
+
+        ClusterStateRequest clusterStateRequest = new ClusterStateRequest();
+        clusterStateRequest.clear().nodes(true).local(true);
+        client.admin().cluster().state(clusterStateRequest, ActionListener.wrap(stateResponse -> {
```
joshpalis, line 160: Can we use `  String localNodeId = clusterService.state().getNodes().getLocalNodeId()` instead of sending a cluster state request?

joshpalis (approved):
LGTM but had a question

amitgalitz (approved):
LGTM, should we add an issue to add multi node integ tests? Also test this manually on multiple node cluster

src/main/java/org/opensearch/flowframework/workflow/WorkflowProcessSorter.java
```diff
@@ -153,31 +153,45 @@ public void validate(List<ProcessNode> processNodes) throws Exception {
...
-
-            installedPluginsFuture.complete(installedPlugins);
-
-        }, exception -> {
-            logger.error("Failed to retrieve installed plugins");
-            installedPluginsFuture.completeExceptionally(exception);
+        final CompletableFuture<List<String>> installedPluginsFuture = new CompletableFuture<>();
+
+        ClusterStateRequest clusterStateRequest = new ClusterStateRequest();
+        clusterStateRequest.clear().nodes(true).local(true);
```
owaiskazi19, line 159: Do we need to add `clusterManagerNodeTimeout` for the clusterStateRequest as well? Or is it fine to ignore it since we are working with the local node?

src/main/java/org/opensearch/flowframework/workflow/WorkflowProcessSorter.java
dbwiddis: Of note, `clusterService.localNode()` internally calls that whole chain and is shorter, but...

What you wrote is what the existing code was, and was possibly causing a deadlock in a multinode cluster.  I'm not sure exactly the cause yet, but this seems the most suspicious.  The `state()` call coded as an `AtomicReference` meaning it has to contend with locking, etc. of the state by the update applier function to be sure the most recent state is returned, all of which depends on a ton of extra information on the cluster that we don't need.

Since all we need is the local node id, we can restrict the cluster state request to `local()` so it fetches from the local node and does not require to fetch from the cluster manager (or have the current state of all the nodes updated). We also restrict it to `nodes()` to only get the nodes info, for a shorter response.

Finally, the way I wrote it is how it's implemented in `_cat/plugins`, so given the unknowns with the `state()` method I feel it's safer.

src/main/java/org/opensearch/flowframework/workflow/WorkflowProcessSorter.java
dbwiddis: Correct, since we're restricting it to local we don't need the manager timeout.  And I've added our own timeout wrapping this whole wait, as a failsafe.

dbwiddis:
> LGTM, should we add an issue to add multi node integ tests? Also test this manually on multiple node cluster

It's being tested multinode in end-to-end testing which is how we found this issue to start with.

But we could change `numNodes` in our existing integ test CI.

src/main/java/org/opensearch/flowframework/workflow/WorkflowProcessSorter.java
joshpalis: Ah good points, this makes sense to me

PR #371, opened 2024-01-05
Include previous node substitutions in nested lists and maps

### LINKED ISSUE #368: [FEATURE] Handle substitutions in nested user inputs map

The `RAGTool` requires two different `model_id` inputs which will come from two different previous nodes.
```json
  {
    "type": "RAGTool",
    "description": "A description of the tool",
    "parameters": {
      "embedding_model_id": "uUiKRowBTbhm9gMcl6RR",
      "inference_model_id": "k_yZWowBQGkb8loYTqch",
      "index": "my_test_data",
      "embedding_field": "embedding",
      "source_field": [
        "text"
      ],
      "input": "${parameters.question}",
      "prompt": "\n\nHuman:You are a professional data analysist. You will always answer question based on the given context first. If the answer is not directly shown in the context, you will analyze the data and find the answer. If you don't know the answer, just say don't know. \n\n Context:\n${parameters.output}\n\nHuman:${parameters.question}\n\nAssistant:"
    }
  }
```
This is similar to the case of the Agent step requiring an `llm.model_id` except we need to uniquely handle more than one possible input.

We already handle substitutions at the `user_inputs` level using the format `${{step_id.output_field}}` so if these models were user inputs this would work:
```
      "embedding_model_id": "${{deploy_embedding_model_step_id.model_id}}",
      "inference_model_id": "${{deploy_inference_model_step_id.model_id}}",
```
Unfortunately these values are nested in the `parameters` map in which the `ToolStep` does not process substitutions.  Further, it does look for a `model_id` sometimes but uses a `findFirst()` in a stream.

Special casing this tool similar to how we special-cased the agent `llm.model_id`.

dbwiddis:
### Description

Extends the substitution pattern `${{ previousNode.key }}` to work inside nested maps and lists in the user inputs.

### Issues Resolved

Fixes #368 

### DIFF
--- a/src/main/java/org/opensearch/flowframework/util/ParseUtils.java
+++ b/src/main/java/org/opensearch/flowframework/util/ParseUtils.java
@@ -30,12 +30,14 @@ import java.io.IOException;
 import java.time.Instant;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Optional;
 import java.util.Set;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 
 import static org.opensearch.core.xcontent.XContentParserUtils.ensureExpectedToken;
 import static org.opensearch.flowframework.common.CommonValue.PARAMETERS_FIELD;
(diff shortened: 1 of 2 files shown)


ohltyler (approved):
LGTM!

joshpalis (approved):

amitgalitz (approved):

owaiskazi19 (approved):
LGTM

Issue #365, opened 2024-01-05
[FEATURE/DOCS] Properly handle model group deletion


dbwiddis:
### Is your feature request related to a problem?

Users can use the `ModelGroup` step to create a model group, but there's no corresponding deletion.   However, [the documentation](https://opensearch.org/docs/latest/ml-commons-plugin/api/model-group-apis/delete-model-group/) indicates that a model group is automatically deleted when the last model in it is deleted (and you can't delete it until then).

There appears to be an edge case of creating a model group (empty) and then wanting to delete it, but this seems an extremely unlikely scenario to encounter in a workflow environment.

### What solution would you like?

Pick one:
1. add a workflow step to delete a model group, recognizing it will often return a deletion status of "not found". (May be preferable for easier tracking of removing the resource id)
2. update the deprovision workflow to use the "noop" step to "deprovision" a model group (probably just a quick change in the `WorkflowResources` enum). (May be preferable if we just want to remove the resource id anyway.)

PR #376, opened 2024-01-05
Added a noop workflow step to delete model group

owaiskazi19:
### Description
Added a noop workflow step to delete model group. The reason of the noop step is that the model group is deleted on its own when there are no model in it. To handle the null case in deprovisioning for register model group step, created this noop workflow step to avoid any conflict with the resources created.

### Issues Resolved
Fixes https://github.com/opensearch-project/flow-framework/issues/365

### DIFF
--- a/src/main/java/org/opensearch/flowframework/workflow/ModelGroupStep.java
+++ b/src/main/java/org/opensearch/flowframework/workflow/ModelGroupStep.java
@@ -48,7 +48,7 @@ public class ModelGroupStep implements WorkflowStep {
     private final FlowFrameworkIndicesHandler flowFrameworkIndicesHandler;
 
     /** The name of this step, used as a key in the template and the {@link WorkflowStepFactory} */
-    public static final String NAME = "register_model_group";
+    public static final String NAME = "model_group";
 
     /**
      * Instantiate this class
--- a/src/main/java/org/opensearch/flowframework/common/WorkflowResources.java
+++ b/src/main/java/org/opensearch/flowframework/common/WorkflowResources.java
@@ -17,6 +17,7 @@ import org.opensearch.flowframework.workflow.CreateIndexStep;
 import org.opensearch.flowframework.workflow.CreateIngestPipelineStep;
 import org.opensearch.flowframework.workflow.DeleteAgentStep;
 import org.opensearch.flowframework.workflow.DeleteConnectorStep;
+import org.opensearch.flowframework.workflow.DeleteModelGroupStep;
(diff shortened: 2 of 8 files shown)


ohltyler (commented):
Few minor comments. A high-level question: why is this a noop instead of doing the actual [deletion](https://opensearch.org/docs/latest/ml-commons-plugin/api/model-group-apis/delete-model-group/)? Could you add some context in the PR description?

src/main/java/org/opensearch/flowframework/workflow/ModelGroupStep.java
```diff
@@ -48,7 +48,7 @@ public class ModelGroupStep implements WorkflowStep {
     private final FlowFrameworkIndicesHandler flowFrameworkIndicesHandler;
 
     /** The name of this step, used as a key in the template and the {@link WorkflowStepFactory} */
-    public static final String NAME = "register_model_group";
+    public static final String NAME = "model_group";
```
amitgalitz, line 51: are we moving away from having the `register`/`create` prefixes?

joshpalis:
Just a heads up, flaky integration tests will be resolved by https://github.com/opensearch-project/flow-framework/issues/377

src/main/java/org/opensearch/flowframework/workflow/ModelGroupStep.java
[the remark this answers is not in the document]
owaiskazi19: Misunderstood. Instead of changing the file name from `ModelGroupStep` to `RegisterModelGroupStep`, I changed the above. Reverted the change.

owaiskazi19: Answered above.

owaiskazi19:
> Few minor comments. A high-level question: why is this a noop instead of doing the actual [deletion](https://opensearch.org/docs/latest/ml-commons-plugin/api/model-group-apis/delete-model-group/)? Could you add some context in the PR description?

Added more details.

amitgalitz (approved):

ohltyler (approved):
LGTM! I assume failing tests are the flaky ones mentioned.

dbwiddis (commented):
So this all works but why do we need a whole class that's a clone of `NoopStep`?

Why not just do this?
```java
REGISTER_MODEL_GROUP(RegisterModelGroupStep.NAME, WorkflowResources.MODEL_GROUP_ID, NoopStep.NAME),
```

PR #381, opened 2024-01-07
Change max retries to retry duration, refactor settings for consistency

### LINKED ISSUE #330: [FEATURE] Overhaul WORKFLOW_REQUEST_TIMEOUT setting
Coming from https://github.com/opensearch-project/flow-framework/pull/317#discussion_r1437288166

We have a setting `WORKFLOW_REQUEST_TIMEOUT` whose only purpose is to provide a timeout for the Create/Update API.  This is someone redundant as there is already a systemwide OpenSearch REST timeout setting, and our create/provision APIs aren't expected to take more than a few hundred milliseconds.

We also have configurable step timeouts, a per-step timeout just implemented in #317 that we control in our workflow-steps.json file, and a per-step timeout override that users may provide with a `node_timeout` setting in their `user_inputs` section.  However, the default if neither of those is specified is hard-coded.

Change the `WORKFLOW_REQUEST_TIMEOUT` to `WORKFLOW_STEP_DEFAULT_TIMEOUT` and use it as the fallback that is currently hardcoded.

Leaving things as they are and documenting a hardcoded value that we could easily configure.   

Adding another timeout setting and documenting why we have a separate one for a call that always returns immediately anyway.

image

### LINKED ISSUE #338: [FEATURE] Replace/augment MAX_GET_TASK_REQUEST_RETRY with a calculation based on timeout

The retryable workflow step abstract class relies on a constant number of retries, `MAX_GET_TASK_REQUEST_RETRY`.  This isn't very useful for two reasons:
 - it might timeout long before the user's specified timeout and then do nothing
 - if it retries past the timeout, it'll just get cancelled, and may as well not exist

Any timeout setting should have some reference to the step's timeout.

Pick one:
 - Keep the `MAX_GET_TASK_REQUEST_RETRY` setting as is, but calculate the delay between retries based on the step timeout
 - Replace `MAX_GET_TASK_REQUEST_RETRY` with a duration between retries setting and calculate retries based on the step timeout and that duration
 - Eliminate `MAX_GET_TASK_REQUEST_RETRY` and just retry at a fixed duration (possibly set by the user) until the timeout
 - Some other idea I haven't thought of

dbwiddis:
### Description

1. Changes retries based on steps to a duration-based retry that will time out.
2. Moves the max workflows and request retry settings to the settings object for consistency with all the other settings.
3. Applies the request timeout setting to all searches. It was only used in the max workflow search.

### Issues Resolved

Fixes #330 
Fixes #338 


src/main/java/org/opensearch/flowframework/common/FlowFrameworkSettings.java
```diff
@@ -82,27 +85,31 @@ public FlowFrameworkSettings(ClusterService clusterService, Settings settings) {
...
+        this.retryDuration = TASK_REQUEST_RETRY_DURATION.get(settings);
         this.maxWorkflowSteps = MAX_WORKFLOW_STEPS.get(settings);
+        this.maxWorkflows = MAX_WORKFLOWS.get(settings);
+        this.requestTimeout = WORKFLOW_REQUEST_TIMEOUT.get(settings);
         clusterService.getClusterSettings().addSettingsUpdateConsumer(FLOW_FRAMEWORK_ENABLED, it -> isFlowFrameworkEnabled = it);
-        clusterService.getClusterSettings().addSettingsUpdateConsumer(MAX_GET_TASK_REQUEST_RETRY, it -> maxRetry = it);
+        clusterService.getClusterSettings().addSettingsUpdateConsumer(TASK_REQUEST_RETRY_DURATION, it -> retryDuration = it);
         clusterService.getClusterSettings().addSettingsUpdateConsumer(MAX_WORKFLOW_STEPS, it -> maxWorkflowSteps = it);
+        clusterService.getClusterSettings().addSettingsUpdateConsumer(MAX_WORKFLOWS, it -> maxWorkflows = it);
+        clusterService.getClusterSettings().addSettingsUpdateConsumer(WORKFLOW_REQUEST_TIMEOUT, it -> requestTimeout = it);
```
owaiskazi19, line 96: Shouldn't we have these settings in the AbstractClass which was deleted below? Reason being the listener for both the settings would be listening only when a REST call is invoked rather during plugin initialization.  Trying to understand more here.

owaiskazi19 (commented):
Overall LGTM with one question

src/main/java/org/opensearch/flowframework/common/FlowFrameworkSettings.java
dbwiddis: > Shouldn't we have these settings in the AbstractClass which was deleted below?

"Should"? Probably no.  "Could?" Maybe, but that adds unnecessary transport layer comms.

Putting them in the Abstract class means you'll end up registering two different update consumers, one for each subclass instance.  Not that big a deal in this case as we know we're only instantiating each RestAction once, and updates are probably rare, but still redundant. 

Putting them in the RestAction means you have to communicate them to the TransportAction in the Request, so we had extra parameters in the WorkflowRequest.

Having them all in one place, in an injected class, makes them easy to access directly in the TransportAction and simplifies the WorkflowRequest.

owaiskazi19: Makes sense. That helps.

owaiskazi19 (approved):

src/main/java/org/opensearch/flowframework/workflow/AbstractRetryableWorkflowStep.java
```diff
@@ -74,9 +73,8 @@ protected void retryableGetMlTask(
         String workflowStep,
         ActionListener<MLTask> mlTaskListener
     ) {
-        AtomicInteger retries = new AtomicInteger();
         CompletableFuture.runAsync(() -> {
-            while (retries.getAndIncrement() < this.maxRetry && !future.isDone()) {
+            while (!future.isDone()) {
```
amitgalitz, line 77: I'm probably glancing over it but where do we take into account the give timeout for each step to stop and call the timeout now

dbwiddis: The `ProcessNode` handles the step timeouts:
https://github.com/opensearch-project/flow-framework/blob/e0f5a2a1a95c794d8a899e5aa052725871cd9a88/src/main/java/org/opensearch/flowframework/workflow/ProcessNode.java#L164-L170

Basically it starts the timeout thread, scheduled to complete exceptionally at the timeout.

Then it starts the step execute thread.

Whichever thread completes the future first wins.... if the step completes, it cancels the delay thread (if for some reason it doesn't do so fast enough, the future can only be completed once so the complete exceptionally becomes a no-op).

joshpalis (approved):

src/main/java/org/opensearch/flowframework/workflow/AbstractRetryableWorkflowStep.java
amitgalitz: I see, thanks for pointing to this

amitgalitz (approved):

Issue #377, opened 2024-01-05
[FEATURE] Separate Local Model Registration to support Custom, Pretrained and Sparse Encoding Models


joshpalis:
### Is your feature request related to a problem?
Coming from flaky integration test failures due to the memory limitations of Github action runners which facilitate integration tests ([Reference](https://github.com/opensearch-project/flow-framework/actions/runs/7427053910/job/20212796098?pr=374)), it has been determined that mitigation would require using a smaller model to test out local model registration within an integration test cluster. 

The `amazon/neural-sparse/opensearch-neural-sparse-tokenizer-v1` has been determined to be the smallest model we can use for testing ([Documentation](https://opensearch.org/docs/latest/ml-commons-plugin/pretrained-models/#sparse-encoding-models)) , however the required fields for registering a `sparse encoding model` is different than the required fields for registering a custom `text embedding model`.

In order to replace the model used in integration testing with this sparse encoding model, it is required to separate out the `RegisterLocalModelStep` into a `RegisterCustomLocalModelStep` and `RegisterSparseEncodingLocalModelStep`. Additionally, we require support for registering an `OpenSearch -provided pretained model`, which does not require a URL. 

### What solution would you like?
The `RegisterCustomLocalModelStep` relates to the following [documentation](https://opensearch.org/docs/latest/ml-commons-plugin/api/model-apis/register-model/#register-a-custom-model) and will have the following required and optional fields : 

Required keys :
- `name`
- `version`
- `model_format`
- `function_name`
- `model_content_hash_value`
- `url`
- `model_type`
- `embedding_dimension`
- `framework_type`

Optional keys :
- `description`
- `model_group_id`
- `all_config`
- `deploy`

The `RegisterSparseEncodingModelStep` relates to the following [documentation ](https://opensearch.org/docs/latest/ml-commons-plugin/api/model-apis/register-model/#register-a-pretrained-sparse-encoding-model) and will have the following required and optional fields 

Required keys :
- `name`
- `version`
- `model_format`
- `function_name`
- `model_content_hash_value`
- `url`

Optional keys :
- `description`
- `model_group_id`
- `deploy`

The `RegisterLocalPretrainedModelStep` relates to the following [documentation](https://opensearch.org/docs/latest/ml-commons-plugin/api/model-apis/register-model/#register-a-pretrained-text-embedding-model) and will have the following required and optional fields

Required keys :
- `name`
- `version`
- `model_format`

Optional keys :
- `description`
[...]

PR #383, opened 2024-01-08
Eliminate Google Guava dependency

### LINKED ISSUE #328: [FEATURE] Eliminate all use of Google Guava

Google Guava is a monolithic "kitchen sink" dependency with [recent CVEs](https://www.cvedetails.com/vulnerability-list/vendor_id-1224/product_id-52274/Google-Guava.html) and a huge (2.9 MB) jar which has tons of utility classes we don't need.

Most of our use of it was for the former Google Collections project's `ImmutableList` and `ImmutableMap` made obsolete by Java 9's `List.of` and `Map.of`, that I've removed in #327.

We still have one remaining use of it for (former Google Commons IO) code reading our mapping JSON files, 3 instances of:
```java
import com.google.common.io.Resources;

URL url = <class name>.class.getClassLoader().getResource(mapping);
return Resources.toString(url, StandardCharsets.UTF_8);
```

We should create a utility class to do this reading without using a large dependency. It's possible in pure Java by replacing `getResource()` with `getResourceAsStream()` and processing the resulting InputStream with one of many built-in ways (Scanner, BufferedReader, ByteArrayOutputStream).  

Here's some code I've hand-edited from an OpenSearch test class (`o.o.search.geo.GeoFilterIT.unZipData()`) that might work:
```java
import org.opensearch.common.io.Streams;

private static String readMappingJSON(String path) throws IOException {
    try (InputStream is = Streams.class.getResourceAsStream(path)) {
        if (is == null) {
            throw new FileNotFoundException("Resource [" + path + "] not found in classpath");
        }
        try (ByteArrayOutputStream baos = new ByteArrayOutputStream()) {
            Streams.copy(is, baos);
            return new String(baos.toByteArray(), StandardCharsets.UTF_8);
        }
    }
}
```

Although now that I see that `Streams` class I can see it has some other methods, like a `readAllLines()`.  `o.o.discovery.gce.GceMockUtils` has a `readJsonResponse` method that uses the `BufferedReader` approach, so you could do this:
```java
try (InputStream is = Streams.class.getResourceAsStream(path)) {
    if (is == null) {
        throw new FileNotFoundException("Resource [" + path + "] not found in classpath");
    }
    final StringBuilder sb = new StringBuilder();
    Streams.readAllLines(is, sb::append);
    return sb.toString();
}
```

Scanner, BufferedReader, other dependencies.

dbwiddis:
### Description

```diff
-    implementation "com.google.guava:guava:33.0.0-jre"
```

### Issues Resolved

Fixes #328


amitgalitz:
I see there is a good amount of failed tests in CI, was this occurring locally too?

joshpalis:
> I see there is a good amount of failed tests in CI, was this occurring locally too?

No this does not occur locally, this is due to memory limitations of GH action runners. Issue https://github.com/opensearch-project/flow-framework/issues/377 details my current approach to mitigate this issue

dbwiddis:
> I see there is a good amount of failed tests in CI, was this occurring locally too?

Tests were passing locally on my personal laptop where I submitted this.  Let me investigate.

dbwiddis:
Failures are caused by not finding the json files:
> java.io.FileNotFoundException: Resource [mappings/global-context.json] not found in classpath

But they are present in `src/main/resources`

Screenshot 2024-01-08 at 9 20 59 AM

Possible classloader issue from the class loading the stream, will try to fix...

dbwiddis:
Fixed. Was using relative classpath which worked with my other laptop config but apparently wasn't portable.

owaiskazi19 (approved):

amitgalitz (approved):

Issue #341, opened 2023-12-29
[BUG] RegisterRemoteModelStep should hardcode "remote" function name


dbwiddis:
### What is the bug?

The `RegisterRemoteModelStep` allows/requires the user to specify `function_name": "remote"` as one of the user inputs.
 - The step relies on this API being synchronous and directly returning a `model_id`
 - If any other function name is supplied, the API executes asynchronously and returns a `task_id`, and execution will fail.

### How can one reproduce the bug?

Provision a workflow with a `register_remote_model` step and either leave the `function_name` user input blank, or use a different function than "remote".

### What is the expected behavior?

The `RegisterRemoteModelStep` executes synchronously regardless of user input.  The function name should be hardcoded to "remote" and any user input ignored.

### Do you have any additional context?

See https://github.com/opensearch-project/ml-commons/blob/f30c3cb135c41f26c16ff2bee47a1cf566826445/plugin/src/main/java/org/opensearch/ml/action/register/TransportRegisterModelAction.java#L303

PR #384, opened 2024-01-08
Adds Register Local custom, pretrained or sparse encoding model steps

joshpalis:
### Description
This PR abstracts the local model registration step and adds the following steps (required, optional keys can be found in the related issue : 

- RegisterLocalCustomModelStep
- RegisterLocalPretrainedModelStep
- RegisterLocalSparseEncodingModelStep

Additionally, this PR aims to resolve flaky integration tests by registering a sparse encoding model, which is significantly smaller in size than the text embedding models used previously

### Issues Resolved
https://github.com/opensearch-project/flow-framework/issues/377


src/main/java/org/opensearch/flowframework/workflow/AbstractRegisterLocalModelStep.java
```diff
@@ -90,68 +84,65 @@ public CompletableFuture<WorkflowData> execute(
...
                 .modelName(modelName)
                 .version(modelVersion)
-                .modelFormat(modelFormat)
-                .hashValue(modelContentHashValue)
-                .modelConfig(modelConfig)
-                .url(url);
+                .modelFormat(MLModelFormat.from(modelFormat));
+
+            if (functionName != null) {
+                mlInputBuilder.functionName(FunctionName.from(functionName));
```
dbwiddis, line 122: Is this the key feature that defines which model we're using?  

We might want to hardcode this per-step and not collect it from the user.   See https://github.com/opensearch-project/flow-framework/issues/341 for where the "remote" function name actually triggers a synchronous code flow.

I see there's a `Function.SPARSE_ENCODING` which probably applies to that model so that one should probably be hardcoded..  Not sure how this applies to the other models (or can they use any function)?

dbwiddis (approved):
LGTM with one comment to consider.

src/main/java/org/opensearch/flowframework/workflow/AbstractRegisterLocalModelStep.java
joshpalis: The Sparse embedding models also support a `Function.SPARSE_TOKENIZE` value for the function name, it really depends on the model config that is provided with the model. Since there's multiple options, I would rather we have the user provide this

dbwiddis: Huh. The docs are confusing.  It says `function_name` is a required input of either SPARSE_ENCODING or SPARSE_TOKENIZE, but then the examples don't use that for opensearch pre-trained, or custom model.

So we might want to have some way of validating that input just for the sparse model.... but not hardcoded since it's still a user's choice.

joshpalis:
I've force pushed and ran the checks multiple times, confirmed that the tests are no longer flaky

src/main/java/org/opensearch/flowframework/workflow/AbstractRegisterLocalModelStep.java
```diff
@@ -90,68 +84,65 @@ public CompletableFuture<WorkflowData> execute(
...
+            }
+            if (modelContentHashValue != null) {
+                mlInputBuilder.hashValue(modelContentHashValue);
+            }
+            if (url != null) {
+                mlInputBuilder.url(url);
+            }
+            if (Stream.of(modelType, embeddingDimension, frameworkType).allMatch(x -> x != null)) {
```
owaiskazi19, line 130: Are these the required keys or can be optional?

src/main/java/org/opensearch/flowframework/workflow/RegisterLocalCustomModelStep.java
```diff
@@ -0,0 +1,91 @@
...
+     */
+    public RegisterLocalCustomModelStep(
+        ThreadPool threadPool,
+        MachineLearningNodeClient mlClient,
+        FlowFrameworkIndicesHandler flowFrameworkIndicesHandler,
+        FlowFrameworkSettings flowFrameworkSettings
+    ) {
+        super(threadPool, mlClient, flowFrameworkIndicesHandler, flowFrameworkSettings);
+        this.mlClient = mlClient;
+        this.flowFrameworkIndicesHandler = flowFrameworkIndicesHandler;
```
owaiskazi19, line 59: Why are we setting `mlClient` and `flowFrameworkIndicesHandler` here since not using anywhere in the class? It's only required by Abstract class and passed in `super` already.

src/main/java/org/opensearch/flowframework/workflow/RegisterLocalPretrainedModelStep.java
```diff
@@ -0,0 +1,73 @@
...
+     */
+    public RegisterLocalPretrainedModelStep(
+        ThreadPool threadPool,
+        MachineLearningNodeClient mlClient,
+        FlowFrameworkIndicesHandler flowFrameworkIndicesHandler,
+        FlowFrameworkSettings flowFrameworkSettings
+    ) {
+        super(threadPool, mlClient, flowFrameworkIndicesHandler, flowFrameworkSettings);
+        this.mlClient = mlClient;
+        this.flowFrameworkIndicesHandler = flowFrameworkIndicesHandler;
```
owaiskazi19, line 51: Same as above.

src/main/java/org/opensearch/flowframework/workflow/RegisterLocalSparseEncodingModelStep.java
```diff
@@ -0,0 +1,77 @@
...
+     */
+    public RegisterLocalSparseEncodingModelStep(
+        ThreadPool threadPool,
+        MachineLearningNodeClient mlClient,
+        FlowFrameworkIndicesHandler flowFrameworkIndicesHandler,
+        FlowFrameworkSettings flowFrameworkSettings
+    ) {
+        super(threadPool, mlClient, flowFrameworkIndicesHandler, flowFrameworkSettings);
+        this.mlClient = mlClient;
+        this.flowFrameworkIndicesHandler = flowFrameworkIndicesHandler;
```
owaiskazi19, line 55: Here too.

src/test/java/org/opensearch/flowframework/rest/FlowFrameworkRestApiIT.java
```diff
@@ -70,15 +70,20 @@ public void testSearchWorkflows() throws Exception {
 
     public void testCreateAndProvisionLocalModelWorkflow() throws Exception {
         // Using a 1 step template to register a local model and deploy model
-        Template template = TestHelpers.createTemplateFromFile("registerlocalmodel-deployflag.json");
+        Template template = TestHelpers.createTemplateFromFile("register-deplylocalsparseencodingmodel.json");
```
owaiskazi19, line 73:
```suggestion
        Template template = TestHelpers.createTemplateFromFile("register-deploylocalsparseencodingmodel.json");
```

owaiskazi19 (commented):
Overall looks good with few comments

src/main/java/org/opensearch/flowframework/workflow/AbstractRegisterLocalModelStep.java
joshpalis: They can be optional or required depending on the implementing sub-class. In this case, these fields are only applicable for local custom model registration. We check for all the possible keys here though in this abstract class, if they're null, then its essentially no-op

owaiskazi19 (approved):

Issue #364, opened 2024-01-05
[FEATURE] Add an API to return validation parameters


dbwiddis:
### Is your feature request related to a problem?

Users do not have an easy way to validate workflow steps, particularly in understanding what parameters are part of the output (e.g., `model_id`) and which inputs are required vs. optional (other than by laboriously visiting original API docs).  We should allow users to easily identify these requirements via API.

In addition, the planned front-end will need access to this same information in order to enable real-time validation in the drag-and-drop interface.

### What solution would you like?

Create a new REST API to return content equivalent to the [workflow-steps.json](https://github.com/opensearch-project/flow-framework/blob/main/src/main/resources/mappings/workflow-steps.json) file.

Allow an optional parameter for the workflow step type which returns only the content of that portion of the file.

### What alternatives have you considered?

1. Force users to navigate documentation.
2. Point users to the JSON file on GitHub.
3. Document every single step on opensearch.org.

### Do you have any additional context?

We already parse the contents of this file into a `WorkflowValidator` class, which internally contains a `Map<String, WorkflowStepValidator>`.
1. The API can simply return the contents of these objects, which simply requires `toXContent()` methods.
2. We could also avoid the potential issues with parsing this file (e.g., users experiencing runtime validation checks rather than compile time checks) by simply hardcoding the `WorkflowStepValidator` map.  We can then remove the `parse` methods and all the complexity with exceptions that they bring.

owaiskazi19:
Request: `GET /_plugins/_flow_framework/workflow/_step`

Response:
```
{
  "undeploy_model": {
    "inputs": [
      "model_id"
    ],
    "outputs": [
      "success"
    ],
    "required_plugins": [
      "opensearch-ml"
    ]
  },
  "register_model_group": {
    "inputs": [
      "name"
    ],
    "outputs": [
      "model_group_id",
      "model_group_status"
    ],
    "required_plugins": [
      "opensearch-ml"
    ]
  },
  "register_remote_model": {
    "inputs": [
      "name",
      "connector_id"
    ],
    "outputs": [
      "model_id",
      "register_model_status"
    ],
    "required_plugins": [
      "opensearch-ml"
    ]
  },
  "register_local_sparse_encoding_model": {
    "inputs": [
      "name",
      "version",
      "model_format",
      "function_name",
      "model_content_hash_value",
      "url"
    ],
    "outputs": [
      "model_id",
      "register_model_status"
    ],
    "required_plugins": [
[...]

PR #394, opened 2024-01-09
GET API to fetch workflow-step.json

owaiskazi19:
### Description
This PR achieves the below:
- REST API that will return content equivalent to the [workflow-steps.json](https://github.com/opensearch-project/flow-framework/blob/main/src/main/resources/mappings/workflow-steps.json) file.
- Removed `create_index` and `create_ingest_pipeline` from the json
- Added tests

Request:
```
GET /_plugins/_flow_framework/workflow/_steps
```

Refer [this](https://github.com/opensearch-project/flow-framework/issues/364#issuecomment-1883979469) for response.

### Issues Resolved
Fixes https://github.com/opensearch-project/flow-framework/issues/364


src/main/java/org/opensearch/flowframework/rest/RestGetWorkflowStepAction.java
```diff
@@ -0,0 +1,95 @@
...
+    private static final String GET_WORKFLOW_STEP_ACTION = "get_workflow_step";
+    private static final Logger logger = LogManager.getLogger(RestGetWorkflowStepAction.class);
+    private FlowFrameworkSettings flowFrameworkFeatureEnabledSetting;
+
+    /**
+     * Instantiates a new RestGetWorkflowStepAction
+     * @param flowFrameworkFeatureEnabledSetting Whether this API is enabled
+     */
+    public RestGetWorkflowStepAction(FlowFrameworkSettings flowFrameworkFeatureEnabledSetting) {
+        this.flowFrameworkFeatureEnabledSetting = flowFrameworkFeatureEnabledSetting;
```
joshpalis, line 48: just a nit : Let's just call this `flowFrameworkSettings`

src/main/java/org/opensearch/flowframework/rest/RestGetWorkflowStepAction.java
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    @Override
+    public String getName() {
+        return GET_WORKFLOW_STEP_ACTION;
+    }
+
+    @Override
+    public List<Route> routes() {
+        return List.of(new Route(RestRequest.Method.GET, String.format(Locale.ROOT, "%s/%s", WORKFLOW_URI, "_step")));
```
joshpalis, line 58: `_step` doesn't seem like the best name here since we will be returning definitions for all the steps

src/main/java/org/opensearch/flowframework/transport/GetWorkflowStepTransportAction.java
```diff
@@ -0,0 +1,49 @@
...
+     */
+    @Inject
+    public GetWorkflowStepTransportAction(TransportService transportService, ActionFilters actionFilters) {
+        super(GetWorkflowStepAction.NAME, transportService, actionFilters, WorkflowRequest::new);
+    }
+
+    @Override
+    protected void doExecute(Task task, WorkflowRequest request, ActionListener<GetWorkflowStepResponse> listener) {
+        try {
+            listener.onResponse(new GetWorkflowStepResponse(WorkflowValidator.parse("mappings/workflow-steps.json")));
```
joshpalis, line 43: I see we're still retrieving the definitions from a json file, perhaps it might be better to have these definitions defined in a class, something like the Workflow Resources class. But this is just a suggestion for improvement, won't block this PR

joshpalis (approved):
Approved, but with some small comments and a suggestion

src/main/java/org/opensearch/flowframework/rest/RestGetWorkflowStepAction.java
owaiskazi19: `_steps`, `_workflows`? Take your pick

src/main/java/org/opensearch/flowframework/transport/GetWorkflowStepTransportAction.java
```diff
@@ -0,0 +1,49 @@
...
+     * @param transportService the transport service
+     * @param actionFilters action filters
+     */
+    @Inject
+    public GetWorkflowStepTransportAction(TransportService transportService, ActionFilters actionFilters) {
+        super(GetWorkflowStepAction.NAME, transportService, actionFilters, WorkflowRequest::new);
+    }
+
+    @Override
+    protected void doExecute(Task task, WorkflowRequest request, ActionListener<GetWorkflowStepResponse> listener) {
```
dbwiddis, line 41: Why do we need a `WorkflowRequest` for this?  We never even use the values in the request and it forces you to instantiate a `new WorkflowRequest(null, null)`.

src/main/java/org/opensearch/flowframework/rest/RestGetWorkflowStepAction.java
```diff
@@ -0,0 +1,95 @@
...
+        try {
+            if (!flowFrameworkFeatureEnabledSetting.isFlowFrameworkEnabled()) {
+                throw new FlowFrameworkException(
+                    "This API is disabled. To enable it, update the setting [" + FLOW_FRAMEWORK_ENABLED.getKey() + "] to true.",
+                    RestStatus.FORBIDDEN
+                );
+            }
+
+            WorkflowRequest workflowRequest = new WorkflowRequest(null, null);
+            return channel -> client.execute(GetWorkflowStepAction.INSTANCE, workflowRequest, ActionListener.wrap(response -> {
```
dbwiddis, line 72:
```suggestion
            ActionRequest request = new ActionRequest() {
                @Override
                public ActionRequestValidationException validate() {
                    return null;
                }
            };
            return channel -> client.execute(GetWorkflowStepAction.INSTANCE, request, ActionListener.wrap(response -> {
```

src/main/java/org/opensearch/flowframework/rest/RestGetWorkflowStepAction.java
dbwiddis: `_steps`.  Did we have one where we can return only the one for a specific step name?

src/main/java/org/opensearch/flowframework/transport/GetWorkflowStepTransportAction.java
dbwiddis: Yeah, we should improve this in the future. Another issue is that we're reading/parsing this multiple times just to answer the REST query.  JSON text file parsed to xcontent, parsed to String in the response, parsed back to xcontent to be returned to the user.  A single writeable class could just output the xcontent once.

src/main/java/org/opensearch/flowframework/rest/RestGetWorkflowStepAction.java
joshpalis: +1 to `_steps`

owaiskazi19: > `_steps`. Did we have one where we can return only the one for a specific step name?

Sure. Not yet.

src/main/java/org/opensearch/flowframework/transport/GetWorkflowStepTransportAction.java
owaiskazi19: Changed it to `ActionRequest`.

dbwiddis (approved):
LGTM for an overall API.  If we are able to pass a parameter (yeah that means creating a new request type to send the parameter) to limit it to a single request type that'd be ideal.  That can be in a follow-up PR if desired.

PR #411, opened 2024-01-17
Fix process node zero timeout and handle not found exceptions

dbwiddis:
### Description

Fixes a bug in deprovisioning where if deprovisioning failed the first time, the retry `ProcessNode` had no timeout and blocked threads.

Also handles exceptionally completed futures with an `OpenSearchStatusException` with `NOT_FOUND` status as successful deprovisions.

### DIFF
--- a/src/main/java/org/opensearch/flowframework/transport/DeprovisionWorkflowTransportAction.java
+++ b/src/main/java/org/opensearch/flowframework/transport/DeprovisionWorkflowTransportAction.java
@@ -11,14 +11,15 @@ package org.opensearch.flowframework.transport;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.opensearch.ExceptionsHelper;
+import org.opensearch.OpenSearchStatusException;
 import org.opensearch.action.support.ActionFilters;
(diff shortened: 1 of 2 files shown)


amitgalitz:
Thanks for making this fix! Been trying to look at it for a while with no luck finding root cause. I was wondering if you understood why this was happening only on multi node consistently

amitgalitz (approved):

owaiskazi19 (approved):
Looks good. Thanks for the fix @dbwiddis

src/main/java/org/opensearch/flowframework/transport/DeprovisionWorkflowTransportAction.java
```diff
@@ -138,7 +141,7 @@ private void executeDeprovisionSequence(
                     new WorkflowData(Map.of(getResourceByWorkflowStep(stepName), resource.resourceId()), workflowId, deprovisionStepId),
                     Collections.emptyList(),
                     this.threadPool,
-                    TimeValue.ZERO
+                    flowFrameworkSettings.getRequestTimeout()
```
owaiskazi19, line 144: This really caused the chaos :)

Issue #410, opened 2024-01-17
[BUG] Flow-framework build failed because no opensearch-security dependency


zelinh:
### What is the bug?
Currently I noticed the distribution bundle build for 2.12.0 failed on flow-framework plugin because it can't find opensearch-security dependency after this introduced https://github.com/opensearch-project/flow-framework/blob/f27f2d21a0cc9e3686e6063cb5e95998d1b7bdac/build.gradle#L164
```
* Where:
Build file '/tmp/tmpda1j575p/flow-framework/build.gradle' line: 348

* What went wrong:
A problem occurred evaluating root project 'opensearch-flow-framework'.
> Could not resolve all files for configuration ':zipArchive'.
   > Could not find org.opensearch.plugin:opensearch-security:2.12.0.0.
     Searched in the following locations:
       - https://repo.maven.apache.org/maven2/org/opensearch/plugin/opensearch-security/2.12.0.0/opensearch-security-2.12.0.0.pom
       - file:/home/ci-runner/.m2/repository/org/opensearch/plugin/opensearch-security/2.12.0.0/opensearch-security-2.12.0.0.pom
       - https://aws.oss.sonatype.org/content/repositories/snapshots/org/opensearch/plugin/opensearch-security/2.12.0.0/opensearch-security-2.12.0.0.pom
       - https://plugins.gradle.org/m2/org/opensearch/plugin/opensearch-security/2.12.0.0/opensearch-security-2.12.0.0.pom
     Required by:
         project :

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.
> Get more help at https://help.gradle.org.

BUILD FAILED in 11s
```
More error context can be found [here](https://build.ci.opensearch.org/blue/rest/organizations/jenkins/pipelines/distribution-build-opensearch/runs/9197/nodes/69/steps/360/log/?start=0).

There are two approaches that we think could resolve this:
1. Since it says there are test time dependency, when we run build, we should exclude these test dependencies.
You could create a customized build script just like how k-NN did here. https://github.com/opensearch-project/k-NN/blob/7900dbb339ab56a56a09cf07c50381aeba084cbb/scripts/build.sh#L121 and disable any test time dependency using `-x integTest`
You could copy our default build script from https://github.com/opensearch-project/opensearch-build/blob/main/scripts/default/opensearch/build.sh to your repo and made customization on that.

[...]

PR #418, opened 2024-01-17
Excluded integTest from build script to handle the security plugin dependency

owaiskazi19:
### Description
Excluded integTest from build script to handle the security plugin dependency.

Refer https://github.com/opensearch-project/neural-search/pull/543 for more details.

### Issues Resolved
Fixes https://github.com/opensearch-project/flow-framework/issues/410


amitgalitz (approved):

joshpalis (approved):

build.gradle
```diff
@@ -183,7 +184,7 @@ ext{
     configureSecurityPlugin = { OpenSearchCluster cluster ->
 
         // Retrieve Security Plugin Zip from zipArchive
-        configurations.zipArchive.asFileTree.each {
+        configurations.secureIntegTestPluginArchive.asFileTree.each {
             if(it.name.contains("opensearch-security")) {
```
joshpalis, line 188: Just a nit, we wont need this name check since only security is in the `secureIntegTestPluginArchive`

build.gradle
```diff
@@ -347,7 +348,6 @@ testClusters.integTest {
 
     // Installs all registered zipArchive dependencies on integTest cluster nodes except security
     configurations.zipArchive.asFileTree.each {
-        if(!it.name.contains("opensearch-security")) {
```
prudhvigodithi, line 350: Hey @owaiskazi19 instead of new `secureIntegTestPluginArchive` configuration, I assume just changing the `if` statement to `if(it.name.contains("opensearch-security"))` should work. Can you please check?

owaiskazi19: @prudhvigodithi here we have to install all the plugins except security plugin. `zipArchive` would have security plugin in it.

vibrantvarun: Yes, +1 to @owaiskazi19 . Initially we kept it how @prudhvigodithi you mentioned above. But the problem, here we just need security plugin to run integ tests, therefore zip archive tries to find it at the compile time also. Therefore, we need to change the tag to securePluginTestArchive

prudhvigodithi: Thanks @vibrantvarun and @owaiskazi19 I'm afraid introducing new configuration `secureIntegTestPluginArchive`  might never be called. Also I noticed  the logic is called in `ext` (extra properties) where as I see similar code in `integTest`, If not needed we can completely remove the in `ext`.
```
testClusters.integTest {
    testDistribution = "ARCHIVE"

    // Optionally install security
    if (System.getProperty("security.enabled") != null && System.getProperty("security.enabled") == "true") {
        configureSecurityPlugin(testClusters.integTest)
    }

    // Installs all registered zipArchive dependencies on integTest cluster nodes except security
    configurations.zipArchive.asFileTree.each {
            plugin(provider(new Callable<RegularFile>(){
                    @Override
                    RegularFile call() throws Exception {
                        return new RegularFile() {
                            @Override
                            File getAsFile() {
                                return it
                            }
                        }
                    }
                })
            )

    }
   ```

Issue #409, opened 2024-01-16
[BUG] Multi Node failures


amitgalitz:
### What is the bug?
We have experienced a few bugs lately when running flow-framework on a multi node setting. 
1. we encountered is when we run the template below and all resources are created successfully, however we get a timeout exception from the completableFuture.  
2. We have all seen lots of issues with deprovisioning on a multi node cluster with the the cluster hanging up and getting stuck, this is probably due to not having all the future complete either successfully or exceptionally. (https://github.com/opensearch-project/flow-framework/actions/runs/7507094368/job/20441779804) 

### How can one reproduce the bug?
Template to reproduce first issue mentioned above:
```
{
  "name": "test-template",
  "description": "template description,
  "use_case": "REGISTER_MODEL",
  "version": {
    "template": "1.0.0",
    "compatibility": ["2.12.0", "3.0.0"]
  },
  "workflows": {
    "provision": {
      "user_params": {},
      "nodes": [
        {
          "id": "create_test_connector",
          "type": "create_connector",
          "previous_node_inputs": {},
          "user_inputs": {
            "version": "1",
            "name": "test connector",
            "protocol": "aws_sigv4",
            "description": "test connector",
            "actions": [
              {
                "headers": {
                  "x-amz-content-sha256": "required",
                  "content-type": "application/json"
                },
                "method": "POST",
                "request_body": "{}",
                "action_type": "predict",
                "url": "test.url"
              }
            ],
            "credential": {
              "access_key": "123",
              "secret_key": "1234"
            },
            "parameters": {
              "endpoint": "test-endpoint",
              "content_type": "application/json",
              "auth": "Sig_V4"
            }
          }
        },
        {
          "id": "register_test_model",
          "type": "register_remote_model",
          "previous_node_inputs": {
            "create_test_connector": "connector_id"
          },
          "user_inputs": {
            "description": "test model",
            "name": "test-model"
          }
        },
        {
          "id": "deploy_test_model",
          "type": "deploy_model",
          "previous_node_inputs": {
            "register_test_model": "model_id"
          }
        }
      ]
    }
  }
}
```

### What is the expected behavior?
Cluster shouldn't be stuck on deprovisioning request and we shouldn't have inconsistency between the workflow status and resources that were created. 

[...]

PR #419, opened 2024-01-18
Replace all CompletableFutures with PlainActionFutures

owaiskazi19:
### Description
Replace all CompletableFutures with PlainActionFutures
Except 
- `CompletableFuture.runAsync`

### Issues Resolved
Fixes https://github.com/opensearch-project/flow-framework/issues/409

### DIFF
--- a/src/main/java/org/opensearch/flowframework/workflow/ProcessNode.java
+++ b/src/main/java/org/opensearch/flowframework/workflow/ProcessNode.java
@@ -10,10 +10,12 @@ package org.opensearch.flowframework.workflow;
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.opensearch.action.support.PlainActionFuture;
 import org.opensearch.common.unit.TimeValue;
 import org.opensearch.threadpool.Scheduler.ScheduledCancellable;
 import org.opensearch.threadpool.ThreadPool;
 
+import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -39,7 +41,7 @@ public class ProcessNode {
     private final ThreadPool threadPool;
     private final TimeValue nodeTimeout;
 
-    private final CompletableFuture<WorkflowData> future = new CompletableFuture<>();
+    private final PlainActionFuture<WorkflowData> future = PlainActionFuture.newFuture();
 
     /**
      * Create this node linked to its executing process, including input data and any predecessor nodes.
@@ -109,7 +111,7 @@ public class ProcessNode {
      * @return A future indicating the processing state of this node.
      * Returns {@code null} if it has not begun executing, should not happen if a workflow is sorted and executed topologically.
      */
-    public CompletableFuture<WorkflowData> future() {
+    public PlainActionFuture<WorkflowData> future() {
         return future;
     }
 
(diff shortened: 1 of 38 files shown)


owaiskazi19:
@zane-neo mind taking a look.

dbwiddis (commented):
LGTM!   A few suggestions:
 - We use the completed future for `WorkflowData.EMPTY` so often we should make it a constant (probalby in the `WorkflowData` class
 - You can really streamline the `ProcessNode` collection of predecessor future values.  The previous multi-step was only there to enable a varargs array for `allOf()`.
 - Replace all `future.isDone()` with a test for `future.getState()` matching `COMPLETED`.
 - Double-check the exception handling everywhere you've replaced `join()` with either `get()` or `actionGet()` as it's different.

src/main/java/org/opensearch/flowframework/workflow/ProcessNode.java
```diff
@@ -139,51 +141,53 @@ public TimeValue nodeTimeout() {
      * @return this node's future.
      * This is returned immediately, while process execution continues asynchronously.
      */
-    public CompletableFuture<WorkflowData> execute() {
+    public PlainActionFuture<WorkflowData> execute() {
         if (this.future.isDone()) {
             throw new IllegalStateException("Process Node [" + this.id + "] already executed.");
         }
 
         CompletableFuture.runAsync(() -> {
```
zane-neo, line 149: In fact I don't have preference between `CompletableFuture` and `PlainActionFuture`, the bottleneck I meant is that the runAsync or other run methods in CompletableFuture uses ForkjoinPool which could be a small number, we should provide a customer thread pool for CompletableFuture to run tasks, e.g. `CompletableFuture.runAsync(() -> {yourTask}, Executors.myExcutor())`

dbwiddis: We do use our own thread pool executor for the [two times we use `runAsync()`](https://github.com/search?q=repo%3Aopensearch-project%2Fflow-framework%20runasync&type=code)
https://github.com/opensearch-project/flow-framework/blob/bf2cad068996c4aee9f38352ef3dab4a04c1823b/src/main/java/org/opensearch/flowframework/workflow/ProcessNode.java#L188

and 

https://github.com/opensearch-project/flow-framework/blob/bf2cad068996c4aee9f38352ef3dab4a04c1823b/src/main/java/org/opensearch/flowframework/workflow/AbstractRetryableWorkflowStep.java#L131

Are there any other bottlenecks you see?

The main benefit of this replacement, as I understood it, is removing (blocking) `.join()` or `.get()` calls from `CompletableFuture` with their equivalent methods in `PlainActionFuture` which extends `ActionListener` and thus uses the `onResponse` and `onFailure` callbacks which I presume work much better in a multithreaded/multichannel environment.

src/test/java/org/opensearch/flowframework/workflow/CreateConnectorStepTests.java
```diff
@@ -130,7 +130,7 @@ public void testCreateConnectorFailure() throws IOException {
 
         verify(machineLearningNodeClient).createConnector(any(MLCreateConnectorInput.class), any());
 
-        assertTrue(future.isCompletedExceptionally());
+        assertTrue(future.isDone());
```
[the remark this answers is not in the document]
owaiskazi19: `getState()` is not a public method.

src/main/java/org/opensearch/flowframework/workflow/ProcessNode.java
owaiskazi19: @zane-neo any other suggestion?

.github/workflows/test_security.yml
```diff
@@ -16,7 +16,7 @@ jobs:
   integ-test-with-security-linux:
     strategy:
       matrix:
-        java: [11, 17, 21]
+        java: [11, 17,  21.0.1]
```
dbwiddis, line 19: Yay, security tests pass now!

src/test/java/org/opensearch/flowframework/workflow/CreateConnectorStepTests.java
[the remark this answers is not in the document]
dbwiddis: I guess we're stuck with `isDone()`.

dbwiddis (approved):
LGTM!

src/main/java/org/opensearch/flowframework/workflow/ProcessNode.java
```diff
@@ -139,51 +139,45 @@ public TimeValue nodeTimeout() {
...
-            List<CompletableFuture<WorkflowData>> predFutures = predecessors.stream().map(p -> p.future()).collect(Collectors.toList());
             try {
-                if (!predecessors.isEmpty()) {
-                    CompletableFuture<Void> waitForPredecessors = CompletableFuture.allOf(predFutures.toArray(new CompletableFuture<?>[0]));
-                    waitForPredecessors.join();
-                }
-
                 logger.info("Starting {}.", this.id);
```
dbwiddis, line 149: Already approved but one quick note. This log entry should be after the `actionGet()` call so it doesn't post until the previous nodes have all finished.

dbwiddis:
Integ tests are failing in JDK 21, probably need all our CI updated to force 21.0.1

.github/workflows/test_security.yml
dbwiddis: I've applied this change in #429 so you can just rebase to main and leave it unchanged

joshpalis: PR https://github.com/opensearch-project/flow-framework/pull/432 fixes our multi-node integration tests, perhaps lets merge that PR in first prior to this PR so we can gauge if these changes have any effect.

joshpalis (approved):
Changes LGTM

PR #421, opened 2024-01-19
Change thread pool to a ScalingExecutorBuilder

### LINKED ISSUE #420: [FEATURE] Make thread pool size a configurable conatant

The thread pool size is hard coded and may not scale well to large loads.  Users should have the ability to increase the thread pool size.

See #265 for possible impacts of a too-low limit.

Change the `FixedExecutorBuilder` to a `ResizableExecutorBuilder` which permits users to update the `size` and `queue_size`.  Add appropriate values to draft documentation.

See usage here:

https://github.com/opensearch-project/OpenSearch/blob/904c9a99987c5bb2b2db02a9e2c911af51f23616/server/src/main/java/org/opensearch/threadpool/ThreadPool.java#L244

dbwiddis:
### Description

Changes Flow Framework thread pool to use a ScalingExecutorBuilder
 - the `FixedExecutorBuilder` establishes a maximum queue size which is arbitrary and not scalable
 - the `ResizableExecutorBuilder` allowing users to manually change this isn't used much
 - the `ScalingExecutorBuilder` is widely used across OpenSearch and plugins and is clearly the preferred option here unless we know better. (Which we don't.)

Matches ["generic" thread pool type](https://github.com/opensearch-project/OpenSearch/blob/e265355e8cf8399443b84104032af075953f1245/server/src/main/java/org/opensearch/threadpool/ThreadPool.java#L238), with different number of processors since it's more narrowly focused.

Uses [similar threadpool to AD](https://github.com/opensearch-project/anomaly-detection/blob/7c0ce4cb554a05230ecf5365b903219f357997bf/src/main/java/org/opensearch/timeseries/TimeSeriesAnalyticsPlugin.java#L845-L853) except with shorter keepalive to correspond to [most OpenSearch thread pool times](https://github.com/opensearch-project/OpenSearch/blob/e265355e8cf8399443b84104032af075953f1245/server/src/main/java/org/opensearch/threadpool/ThreadPool.java#L251-L254), and uses full number of allocated processors (smaller # than generic).

Allows user to change # of processors per node via settings.

### Issues Resolved

Fixes #420 

### DIFF
--- a/src/main/java/org/opensearch/flowframework/FlowFrameworkPlugin.java
+++ b/src/main/java/org/opensearch/flowframework/FlowFrameworkPlugin.java
@@ -18,6 +18,7 @@ import org.opensearch.common.settings.IndexScopedSettings;
 import org.opensearch.common.settings.Setting;
 import org.opensearch.common.settings.Settings;
 import org.opensearch.common.settings.SettingsFilter;
+import org.opensearch.common.unit.TimeValue;
 import org.opensearch.common.util.concurrent.OpenSearchExecutors;
 import org.opensearch.core.action.ActionResponse;
 import org.opensearch.core.common.io.stream.NamedWriteableRegistry;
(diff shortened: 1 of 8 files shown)


joshpalis (approved):

amitgalitz (approved):

amitgalitz:
btw just for reference I think we used fixed as that was used by ml-commons here: https://github.com/opensearch-project/ml-commons/blob/41fac826c8fd1182f83c1f3f6b301bf17022eb2f/plugin/src/main/java/org/opensearch/ml/plugin/MachineLearningPlugin.java#L784

jackiehanyang (approved):

PR #432, opened 2024-01-21
Re-enabled multi-node integration tests and Fix integration test set up

joshpalis:
### Description
This PR re-enables multi-node integration tests for JDK 21.0.1 and aims to resolve flaky integration tests.

### Issues Resolved
Part of https://github.com/opensearch-project/flow-framework/issues/409

### DIFF
--- a/src/test/java/org/opensearch/flowframework/FlowFrameworkRestTestCase.java
+++ b/src/test/java/org/opensearch/flowframework/FlowFrameworkRestTestCase.java
@@ -273,6 +273,7 @@ public abstract class FlowFrameworkRestTestCase extends OpenSearchRestTestCase {
                 // Do not reset ML/Flow Framework encryption index as this is needed to encrypt connector credentials
                 if (indexName != null
                     && !".opendistro_security".equals(indexName)
+                    && !".plugins-ml-model-group".equals(indexName)
                     && !".plugins-ml-config".equals(indexName)
                     && !".plugins-flow-framework-config".equals(indexName)) {
                     adminClient().performRequest(new Request("DELETE", "/" + indexName));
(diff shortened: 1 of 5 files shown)


joshpalis:
Security tests are failing due to the misconfiguration of the admin client when attempting to wipe all indices after each test run. The admin client incorrectly does not have the permissions to delete system indices. Will handle after non-security integration / multi-node tests pass consistently

src/test/java/org/opensearch/flowframework/FlowFrameworkRestTestCase.java
```diff
@@ -187,7 +203,7 @@ protected static void deleteIndexWithAdminClient(String name) throws IOException
 
     // Utility fn for checking if an index exists. Should only be used when not allowed in a regular context
     // (e.g., checking existence of system indices)
-    protected static boolean indexExistsWithAdminClient(String indexName) throws IOException {
+    public static boolean indexExistsWithAdminClient(String indexName) throws IOException {
```
dbwiddis, line 206: For my own education why do we need to use `adminClient()` here (and elsewhere) instead of `client()`?  

This method looks exactly like the superclass `indexExists()` method except using the `adminClient`.   The javadoc for the inherited `adminClient()` says it should only be used for cleanup (in the superclass it's only used with DELETE calls).   Is it because we're manipulating system indices?

dbwiddis (approved):
LGTM!

src/test/java/org/opensearch/flowframework/FlowFrameworkRestTestCase.java
joshpalis: Yes, since this class supports both security enabled and disabled integration tests, in both cases, the admin client will always have permissions to search/delete system indices. As for why we're manipulating system indices, due to the node connectivity/drop issues we were seeing in the failed CI runs, I suspected the issue was lack of storage, so I thought deleting all created/populated indices after each test would mitigate this issue. 

On the same note, the reason why we exclude the ML config index from this clean up is because of the cron job that initializes the index and encryption key. If we delete it every single time, then each test in which we create a connector would need to wait for the next interval of the cron job to run in order to pass.

owaiskazi19 (approved):
LGTM

owaiskazi19:
@joshpalis this PR would need a manual backport.

Issue #442, opened 2024-01-24
[CI] macOS Integ Test Flakiness


dbwiddis:
### Is your feature request related to a problem?

Integ tests are flaky on macOS.  Given that failures have primarily been resource-constrained, and that macOS runners are really VMs and tend to ve more limiting, this is not a surprise.

### What solution would you like?

Pick one:
 - Include longer timeouts that reduce/eliminate issues with macOS tests.  We're already wating 4-6 minutes for tests to complete, adding an extra minute to avoid a "retry" that adds another 4-6 minutes is a win.
 - Use JNA `Platform` class to test whether we're on macOS and add longer time for that platform
 - Use a `bash` script in the github action to automatically retry N times.  See [this example](https://github.com/oshi/oshi/blob/01e192a28e5d267a8f360f82d830a6f5fbe48939/.github/workflows/unix.yaml#L99) from a project I maintain.  This (with 2 tries) is consistent with OpenSearch Jenkins retries.

### What alternatives have you considered?

Investigating whether macOS is officially supported and if not, using GitHub's `allow_failure: true` on macOS tests.  I don't like this option but it's the easiest.

### Do you have any additional context?

macOS is a fantastic operating system but it is specifically designed to run on Apple hardware and not in a VM.

owaiskazi19:
How about doing something similar to core https://github.com/opensearch-project/OpenSearch/issues/2547? We can use the https://blog.gradle.org/gradle-flaky-test-retry-plugin plugin to perform the retry.

dbwiddis:
Hey @kgantchev thanks for the idea. I've actually taken advantage of these runners on other projects I maintain.

All OpenSearch project repo settings are centrally managed so getting this app approved has to go through these organization admins.  We'd need to come up with a justification for the need to do this testing to propose adding it, but I am not sure it's really necessary here (and given the scale of this organization's projects, I doubt we'd fit under your free offering.)

our primary user base uses cloud servers on hardware optimized for that purpose (flexible storage and memory choices, no need for displays), while Apple Silicon is primarily optimized for the end user market (integrated on-chip memory and GPU).  So our investments in testing hardware/software need to focus on the overwhelming majority of users.

kgantchev:
> I've actually taken advantage of these runners on other projects I maintain.  

Awesome! I'm glad it has come in handy.  

> ...  
[...]

PR #449, opened 2024-01-24
Retry flaky tests

owaiskazi19:
### Description
Retry flaky tests using `org.gradle.test-retry` plugin.

### Issues Resolved
Closes https://github.com/opensearch-project/flow-framework/issues/442

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 ## [Unreleased]
 ### Added
 - Github workflow for changelog verification ([#440](https://github.com/opensearch-project/flow-framework/pull/440))
+- Retry flaky tests ([#449](https://github.com/opensearch-project/flow-framework/pull/449))
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,6 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 ## [Unreleased]
 ### Added
 - Github workflow for changelog verification ([#440](https://github.com/opensearch-project/flow-framework/pull/440))
-- Retry flaky tests ([#449](https://github.com/opensearch-project/flow-framework/pull/449))
 
 ### Changed
 
--- a/build.gradle
+++ b/build.gradle
@@ -410,11 +410,11 @@ task integTestRemote(type: RestIntegTestTask) {
 // test retry configuration
 subprojects {
     apply plugin: "org.gradle.test-retry"
-    tasks.withType(Test).configureEach {
+    tasks.withType(RestIntegTestTask).configureEach {
         retry {
             failOnPassedAfterRetry = false
             maxRetries = 3
-            maxFailures = 10
+            maxFailures = 5
         }
     }
 }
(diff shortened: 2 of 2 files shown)


dbwiddis (commented):
Cool app. 

We do not need to apply it to all tests.  Just IT.  And let’s limit to one retry.

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 ## [Unreleased]
 ### Added
 - Github workflow for changelog verification ([#440](https://github.com/opensearch-project/flow-framework/pull/440))
+- Retry flaky tests ([#449](https://github.com/opensearch-project/flow-framework/pull/449))
```
dbwiddis, line 10: I don't think we need to add CI tweaks to our changelog. This is for user-facing changes in the plugin.  Also we don't have to really start adding anything until we cut the 2.12 branch at code freeze.

build.gradle
```diff
@@ -405,6 +406,19 @@ task integTestRemote(type: RestIntegTestTask) {
...
 
+
+// test retry configuration
+subprojects {
+    apply plugin: "org.gradle.test-retry"
+    tasks.withType(Test).configureEach {
+        retry {
+            failOnPassedAfterRetry = false
+            maxRetries = 3
+            maxFailures = 10
```
dbwiddis, line 417: Let's make this a lower number.  Typically we only should see 1 or 2 failures getting retried in a test run, so this value should definitely be less than the max # of tests we run (7) and probably closer to 3.

owaiskazi19: Changed it to 1 and 3 respectively.

owaiskazi19: Looks like 1 and 3 didn't work. Will try incrementing `maxRetries.`

joshpalis (approved):
Just a small comment, but looks got to me. I'm still not sure why we're encountering these multi-node failures on JDK 21.0.2 but not JDK 21.0.1

build.gradle
```diff
@@ -405,6 +406,19 @@ task integTestRemote(type: RestIntegTestTask) {
     }
 }
 
+
+// test retry configuration
```
joshpalis, line 410: Can we add a TODO here for a long term fix, I'm just concerned that we'll miss catching some flaky issues that may help us identify some gaps/ improvements in the future. But as a short term fix, this works wonderfully.

joshpalis (approved):
Just a small comment, but looks got to me. I'm still not sure why we're encountering these multi-node failures on JDK 21.0.2 but not JDK 21.0.1

build.gradle
joshpalis: And perhaps with this change we can reset the `fail fast` setting back to its default within our CI workflow

dbwiddis: Why do we want to reset fail fast?  It's frustrating when a flaky test somewhere kills a bunch of other tests, and we don't know which one(s) would have failed.

dbwiddis (approved):

amitgalitz:
looks good to me, but I agree with Josh that we need another issue to solve these failures, the multi node flakiness is still a little concerning

amitgalitz (approved):

build.gradle
owaiskazi19: Multi nodes on macos and windows are failing because of the runner size.

build.gradle
owaiskazi19: Incremented it to 3 and 5.

PR #458, opened 2024-01-26
Separate multi-node integ tests to their own job

dbwiddis:
### Description

Multi-node integration tests were set to run on the same server as integration tests had just run on.  On cluster startup, the nodes were reading the previous cluster state from the (dropped) single node and uses the same stored indices.

Additionally ML Commons recently made the default `true` for a feature that auto-deploys ML models after a node drop and recovery, so these recoveries were occurring.

While no guarantee separating this out will fix the flaky macOS tests, I've seen enough hints in the logs to suggest that leftover bits from previous tests are slowing startup and potentially contributing to test failures. 

### Issues Resolved

Might resolve flaky tests. At least will cut down on the debugging noise.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -79,7 +79,22 @@ jobs:
       - name: Build and Run Tests
         run: |
           ./gradlew integTest yamlRestTest
-      - name: Multi Nodes Integration Testing
-        if: matrix.java == '21'
+  integMultiNodeTest:
+    needs: [spotless, javadoc]
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+        java: [21]
+    name: Multi-Node Integ Test JDK${{ matrix.java }}, ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up JDK ${{ matrix.java }}
+        uses: actions/setup-java@v4
+        with:
+          java-version: ${{ matrix.java }}
+          distribution: temurin
+      - name: Build and Run Tests
         run: |
           ./gradlew integTest -PnumNodes=3


.github/workflows/CI.yml
```diff
@@ -79,7 +79,22 @@ jobs:
       - name: Build and Run Tests
         run: |
           ./gradlew integTest yamlRestTest
-      - name: Multi Nodes Integration Testing
-        if: matrix.java == '21'
+  integMultiNodeTest:
+    needs: [spotless, javadoc]
```
owaiskazi19, line 83: Nit: Should we also add `integTest` here? If `integTest` itself fails then we can avoid running multi node integ tests. WDYT?

.github/workflows/CI.yml
```diff
@@ -79,7 +79,22 @@ jobs:
...
+    strategy:
+      fail-fast: false
```
owaiskazi19, line 85: Should we have this strategy only for `macos` and `windows`?

owaiskazi19 (commented):
Thanks for digging into this @dbwiddis. Great RCA.

.github/workflows/CI.yml
owaiskazi19: But our CI would run for a longer time then.

dbwiddis: Yeah I thought of that and figured we could run in parallel.

.github/workflows/CI.yml
dbwiddis: When they stop being flaky we can remove this.  I don’t think it hurts to have it tho.

owaiskazi19 (approved):

joshpalis (approved):

joshpalis:
> Additionally ML Commons recently made the default true for a feature that auto-deploys ML models after a node drop and recovery, so these recoveries were occurring.

Interesting, I had thought we we're preventing the auto-redeployment of ML Models via the cluster settings [here](https://github.com/opensearch-project/flow-framework/blob/ad199e3822204f17a463c8596787f4063bb84b4c/src/test/java/org/opensearch/flowframework/FlowFrameworkRestTestCase.java#L124) during set up

dbwiddis:
> Interesting, I had thought we we're preventing the auto-redeployment of ML Models via the cluster settings [here](https://github.com/opensearch-project/flow-framework/blob/ad199e3822204f17a463c8596787f4063bb84b4c/src/test/java/org/opensearch/flowframework/FlowFrameworkRestTestCase.java#L124) during set up

Maybe that was a recent change that I missed when spelunking old logs.

But recovering indices and cluster state was happening and I'm pretty sure it impacted the timing of node startup.

Issue #461, opened 2024-01-26
[BUG] Multi-node integ tests are flaky


dbwiddis:
### What is the bug?

Multi-node integ tests are failing randomly.   The common symptom is that the cluster does not appear to be started up when the first test(s) run, which leads to 1 in 3 chance of failure if the REST request hits a node that does not contain the primary shard for a queried index.

### How can one reproduce the bug?

Occurs randomly on GitHub Actions. Latest failures :
 - https://github.com/opensearch-project/flow-framework/actions/runs/7671893205/job/20911274115
 - https://github.com/opensearch-project/flow-framework/actions/runs/7671893205/job/20911275308

### What is the expected behavior?

Passing tests

### What is your host/environment?

GitHub actions. This has occurred on all OS's.  So far we've only done multi-node tests on JDK 21, will continue to evaluate other JDKs.

### Do you have any additional context?

The common line tends to be this exception, always on `::integTest-0` node.  Other nodes have failures ~20s later, associated with cluster manager (which was probably this node).

Note the sequence and timestamps involved here: we do an HTTP request here
```
1> [2024-01-26T16:24:31,158][INFO ][o.o.f.r.FlowFrameworkRestApiIT] [testCreateAndProvisionLocalModelWorkflow] before test
  2> org.opensearch.client.ResponseException: method [POST], host [http://127.0.0.1:37145], URI [/_plugins/_flow_framework/workflow/S90ER40BNKhClHx_ZtQu/_provision], status line [HTTP/1.1 500 Internal Server Error]
    {"error":"Cannot invoke \"String.getBytes(java.nio.charset.Charset)\" because \"src\" is null"}
        at __randomizedtesting.SeedInfo.seed([3366CB2459279EBB:C09AA671C3648979]:0)
        at app//org.opensearch.client.RestClient.convertResponse(RestClient.java:376)
        at app//org.opensearch.client.RestClient.performRequest(RestClient.java:346)
        at app//org.opensearch.client.RestClient.performRequest(RestClient.java:321)
        at app//org.opensearch.flowframework.TestHelpers.makeRequest(TestHelpers.java:116)
        at app//org.opensearch.flowframework.TestHelpers.makeRequest(TestHelpers.java:89)
        at app//org.opensearch.flowframework.TestHelpers.makeRequest(TestHelpers.java:78)
        at app//org.opensearch.flowframework.FlowFrameworkRestTestCase.provisionWorkflow(FlowFrameworkRestTestCase.java:381)
[...]

PR #465, opened 2024-01-28
Add more thread pools

dbwiddis:
### Description

Spent several more hours today digging into the Multi-node integ test failures.  In summary:
 - The first REST call to other nodes sometimes times out.  Subsequent calls always seem to succeed.  Added a single retry in the TestHelpers class that has caught all of these instances.
 - Workflow provisioning is flakily delayed by ML Commons create connector.  With a shorter timeout, we just get a failed provisioning.  Adding a longer timeout results in continuing on to deploy model but a failure "Deploy model failed with error : {"smSrrtMdQgKo8ZLMpnTxtA":"The ML encryption master key has not been initialized yet. Please retry after waiting for 10 seconds."}". This passes the second time with the gradle retry.

Playing around with the thread pool showed it was very limiting.  Split out provisioning, deprovisioning, and retry onto separate thread pools.

I'm confident that this flakiness is due to a lot of setup work that is required in ML Commons and takes longer on the GitHub runners due to smaller numbers of threads available.   Testing on a real cluster with ML Commons set up hasn't shown any issues.

### Issues Resolved

More tweaks to fixes of #461

### DIFF
--- a/src/main/java/org/opensearch/flowframework/FlowFrameworkPlugin.java
+++ b/src/main/java/org/opensearch/flowframework/FlowFrameworkPlugin.java
@@ -187,21 +187,21 @@ public class FlowFrameworkPlugin extends Plugin implements ActionPlugin {
             new ScalingExecutorBuilder(
                 WORKFLOW_THREAD_POOL,
                 1,
-                Math.max(1, OpenSearchExecutors.allocatedProcessors(settings) - 1),
+                Math.max(2, OpenSearchExecutors.allocatedProcessors(settings) - 1),
                 TimeValue.timeValueMinutes(1),
                 FLOW_FRAMEWORK_THREAD_POOL_PREFIX + WORKFLOW_THREAD_POOL
             ),
             new ScalingExecutorBuilder(
                 PROVISION_WORKFLOW_THREAD_POOL,
                 1,
-                Math.max(1, OpenSearchExecutors.allocatedProcessors(settings) - 1),
+                Math.max(4, OpenSearchExecutors.allocatedProcessors(settings) - 1),
                 TimeValue.timeValueMinutes(5),
                 FLOW_FRAMEWORK_THREAD_POOL_PREFIX + PROVISION_WORKFLOW_THREAD_POOL
             ),
             new ScalingExecutorBuilder(
                 DEPROVISION_WORKFLOW_THREAD_POOL,
                 1,
-                Math.max(1, OpenSearchExecutors.allocatedProcessors(settings) - 1),
+                Math.max(2, OpenSearchExecutors.allocatedProcessors(settings) - 1),
                 TimeValue.timeValueMinutes(1),
                 FLOW_FRAMEWORK_THREAD_POOL_PREFIX + DEPROVISION_WORKFLOW_THREAD_POOL
             )
--- a/src/main/java/org/opensearch/flowframework/FlowFrameworkPlugin.java
+++ b/src/main/java/org/opensearch/flowframework/FlowFrameworkPlugin.java
@@ -73,7 +73,9 @@ import java.util.Collection;
 import java.util.List;
 import java.util.function.Supplier;
 
+import static org.opensearch.flowframework.common.CommonValue.DEPROVISION_WORKFLOW_THREAD_POOL;
 import static org.opensearch.flowframework.common.CommonValue.FLOW_FRAMEWORK_THREAD_POOL_PREFIX;
+import static org.opensearch.flowframework.common.CommonValue.PROVISION_WORKFLOW_THREAD_POOL;
 import static org.opensearch.flowframework.common.CommonValue.WORKFLOW_THREAD_POOL;
 import static org.opensearch.flowframework.common.FlowFrameworkSettings.FLOW_FRAMEWORK_ENABLED;
 import static org.opensearch.flowframework.common.FlowFrameworkSettings.MAX_WORKFLOWS;
@@ -185,9 +187,23 @@ public class FlowFrameworkPlugin extends Plugin implements ActionPlugin {
             new ScalingExecutorBuilder(
                 WORKFLOW_THREAD_POOL,
                 1,
-                OpenSearchExecutors.allocatedProcessors(settings),
-                TimeValue.timeValueMinutes(5),
+                Math.max(1, OpenSearchExecutors.allocatedProcessors(settings) - 1),
+                TimeValue.timeValueMinutes(1),
                 FLOW_FRAMEWORK_THREAD_POOL_PREFIX + WORKFLOW_THREAD_POOL
+            ),
+            new ScalingExecutorBuilder(
+                PROVISION_WORKFLOW_THREAD_POOL,
+                1,
+                Math.max(1, OpenSearchExecutors.allocatedProcessors(settings) - 1),
+                TimeValue.timeValueMinutes(5),
+                FLOW_FRAMEWORK_THREAD_POOL_PREFIX + PROVISION_WORKFLOW_THREAD_POOL
+            ),
+            new ScalingExecutorBuilder(
+                DEPROVISION_WORKFLOW_THREAD_POOL,
+                1,
+                Math.max(1, OpenSearchExecutors.allocatedProcessors(settings) - 1),
+                TimeValue.timeValueMinutes(1),
+                FLOW_FRAMEWORK_THREAD_POOL_PREFIX + DEPROVISION_WORKFLOW_THREAD_POOL
             )
         );
     }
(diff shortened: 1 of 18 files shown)


dbwiddis:
One Flaky failure on this PR first run: it's the awaiting for the ML config index that took over a minute.  Bumped the max failures up to 4 since there are 3 such calls/test failures for this.

amitgalitz:
Thanks for making these changes!
quick clarifying question:

>I'm confident that this flakiness is due to a lot of setup work that is required in ML Commons and takes longer on the GitHub runners due to smaller numbers of threads available. Testing on a real cluster with ML Commons set up hasn't shown any issues.

When you say "testing on a real cluster with ml commons set up", do you mean a local multi node cluster with flow framework and ml commons and using the flow-framework APIs?

dbwiddis:
Summary of specific things that happen on startup that this change works around:
1. REST client initiation is done by a superclass of the RestApiIT in a `@Before` method.  Immediately starting tests doesn't permit this to complete on other nodes, resulting in one category of flaky tests.  Waiting more than about 5 seconds before starting the tests seems to avoid this.  This is fixed with both a single retry on makeRequest, and a longer startup delay.  I wouldn't normally include that retry in main code but I think it's fine in a test helper class.
2. Create connector call sometimes takes a long time due to external calls.  The existing default 10s delay was insufficient. Increased the default delay.
3. Creation of `.ml-commons-config` index doesn't happen until (1) ML Commons plugin init which starts a cron job (2) second cycle of the job running at 10s increments, meaning 20 seconds after node startup.  Due to very few processors to run threads, if the full RestApiIT is executed it consumes most of the processing and this init can take over 60 seconds.  Created a one-time 25s delay before executing the tests to give ML Commons time to init.

joshpalis (approved):

src/test/java/org/opensearch/flowframework/rest/FlowFrameworkRestApiIT.java
```diff
@@ -39,6 +42,18 @@
...
 
+    private static AtomicBoolean waitToStart = new AtomicBoolean(true);
+
+    @Before
+    public void waitToStart() throws Exception {
+        // ML Commons cron job runs every 10 seconds and takes 20+ seconds to initialize .plugins-ml-config index
+        // Delay on the first attempt for 25 seconds to allow this initialization and prevent flaky tests
+        if (waitToStart.getAndSet(false)) {
+            CountDownLatch latch = new CountDownLatch(1);
+            latch.await(25, TimeUnit.SECONDS);
```
owaiskazi19, line 53: Should we do just `latch.await()` here, we don't have to define a specific timeout then?

src/main/java/org/opensearch/flowframework/FlowFrameworkPlugin.java
```diff
@@ -185,9 +187,23 @@ public List<ExecutorBuilder<?>> getExecutorBuilders(Settings settings) {
...
                 FLOW_FRAMEWORK_THREAD_POOL_PREFIX + WORKFLOW_THREAD_POOL
+            ),
+            new ScalingExecutorBuilder(
+                PROVISION_WORKFLOW_THREAD_POOL,
+                1,
+                Math.max(1, OpenSearchExecutors.allocatedProcessors(settings) - 1),
+                TimeValue.timeValueMinutes(5),
+                FLOW_FRAMEWORK_THREAD_POOL_PREFIX + PROVISION_WORKFLOW_THREAD_POOL
+            ),
+            new ScalingExecutorBuilder(
```
owaiskazi19, line 201: I feel like we are having couple of `ScalingExecutorBuilder` for our plugin. 
Even [ml-commons](https://github.com/opensearch-project/ml-commons/blob/main/plugin/src/main/java/org/opensearch/ml/plugin/MachineLearningPlugin.java#L772) has FixedExecutorBuilders because of which we have to handle the flaky tests. 
IMO we should make `PROVISION_WORKFLOW_THREAD_POOL` and `DEPROVISION_WORKFLOW_THREAD_POOL` as `FixedExecutorBuilders` to avoid spawning of threads.

owaiskazi19 (commented):
LGTM overall with few questions.

src/test/java/org/opensearch/flowframework/rest/FlowFrameworkRestApiIT.java
dbwiddis: We'd never time out with `.await()`.  This is just a fancy alternative to `Thread.sleep(25000)` which would do the same thing.

src/main/java/org/opensearch/flowframework/FlowFrameworkPlugin.java
dbwiddis: Fixed executors will still spawn the threads in exactly the same way Scaling will, the difference is that all the spawned threads will stay alive and idle at the max.    

WIth scaling, after threads are idle for the time value, they will shut down.    

Pictorially, assuming 4 processors the max size is 3.

Fixed:   1 -> 2 -> 3 -> stays forever with 3 idle threads
Scaling: 1 -> 2 -> 3 -> stays for timeout of oldest thread -> 2 -> stays for timeout of oldest thread -> 1 idle thread

dbwiddis: WIth fixed we have to set a max queue size and it will fail there.  With scaling the queue can grow indefinitely.

dbwiddis: AD uses ScalingExecutorBuilders. They are recommended for non-steady workloads, which Flow Framework is.  It will be idle most of the time and thread pools should release unused idle threads.

Fixed Executor Builders are used for predictable steady-state workloads where threads need to always be ready to execute.

dbwiddis:
Still getting failed ml config index exists errors even with this delay.  I'm not sure what else we can do. 

```
Tests with failures:
 - org.opensearch.flowframework.rest.FlowFrameworkRestApiIT.testFailedUpdateWorkflow
 - org.opensearch.flowframework.rest.FlowFrameworkRestApiIT.testCreateAndProvisionRemoteModelWorkflow
 - org.opensearch.flowframework.rest.FlowFrameworkRestApiIT.testCreateAndProvisionAgentFrameworkWorkflow
```

joshpalis:
Seems the process node tests are leaking threads : 
```
  2>    1) Thread[id=113, name=opensearch[org.opensearch.flowframework.workflow.ProcessNodeTests][[timer]], state=TIMED_WAITING, group=TGRP-ProcessNodeTests]
  2>    2) Thread[id=114, name=opensearch[org.opensearch.flowframework.workflow.ProcessNodeTests][opensearch_provision_workflow][T#1], state=WAITING, group=TGRP-ProcessNodeTests]
  2>    3) Thread[id=116, name=opensearch[org.opensearch.flowframework.workflow.ProcessNodeTests][opensearch_provision_workflow][T#2], state=WAITING, group=TGRP-ProcessNodeTests]
  2>    4) Thread[id=115, name=opensearch[org.opensearch.flowframework.workflow.ProcessNodeTests][scheduler][T#1], state=WAITING, group=TGRP-ProcessNodeTests]
  2>    5) Thread[id=112, name=TEST-ProcessNodeTests.testNode-seed#[2316B57D3BD543DC], state=WAITING, group=TGRP-ProcessNodeTests]
```

dbwiddis:
> Seems the process node tests are leaking threads :
> 
> ```
>   2>    1) Thread[id=113, name=opensearch[org.opensearch.flowframework.workflow.ProcessNodeTests][[timer]], state=TIMED_WAITING, group=TGRP-ProcessNodeTests]
>   2>    2) Thread[id=114, name=opensearch[org.opensearch.flowframework.workflow.ProcessNodeTests][opensearch_provision_workflow][T#1], state=WAITING, group=TGRP-ProcessNodeTests]
>   2>    3) Thread[id=116, name=opensearch[org.opensearch.flowframework.workflow.ProcessNodeTests][opensearch_provision_workflow][T#2], state=WAITING, group=TGRP-ProcessNodeTests]
>   2>    4) Thread[id=115, name=opensearch[org.opensearch.flowframework.workflow.ProcessNodeTests][scheduler][T#1], state=WAITING, group=TGRP-ProcessNodeTests]
>   2>    5) Thread[id=112, name=TEST-ProcessNodeTests.testNode-seed#[2316B57D3BD543DC], state=WAITING, group=TGRP-ProcessNodeTests]
> ```

Yeah, saw that.  The "scheduler" thread would have interrupted that...

Not sure that's really a problem or just where the code happened to be when the whole test suite timed out for some other reason.

In any case, I saw a lot of simplification we could do in process node after switching to action future and put that in.

dbwiddis:
Tested locally with threadpool max of 1 and failed on too long provisioning, so also increased the minimum "max" scale for the pools even with a small number of processors.  Seemed to make the tests go a lot faster, too....

owaiskazi19 (approved):

PR #466, opened 2024-01-29
Prevent provisioning an already-provisioned workflow

### LINKED ISSUE #445: [DISCUSS] Handling duplicate calls to provision API

If a user calls the provision API twice using the same workflow, its provisioned resources are duplicated.  This can be confusing as the combination of workflow step name and workflow step ID is no longer unique, and choosing the correct resource ID requires guesswork.   Also the state changes to COMPLETED when the first workflow finishes, even though more resources are being added.

Additionally this results in the additional copy (or copies) consuming cluster resources.

Prevent a workflow from being provisioned if its status is `PROVISIONING`.   This status may need to be checked multiple times (like double-checked-locking) to prevent the race condition where two provision calls both see `NOT_STARTED` before one of them begins provisioning.

While addressing this, we also should discuss how to handle `FAILED`.  We could prevent provisioning and require users to use the deprovision API first (basically only allow provisioning from the `NOT_STARTED` state) or we could try to "pick up from where we left off", adding logic to skip provisioining steps if the resource already exists (complex and brittle).

Leaving the API as is, and/or encouraging use of the create API with the provisioning parameter, which does not have this shortcoming.

We should probably handle this similarly to trying to create the same Anomaly Detector twice.

dbwiddis:
### Description

Prevents provisioning a workflow whose state is anything other than NOT_STARTED.

Clarified some associated method/variable naming.

### Issues Resolved

Fixes #445 

### DIFF
--- a/src/main/java/org/opensearch/flowframework/indices/FlowFrameworkIndicesHandler.java
+++ b/src/main/java/org/opensearch/flowframework/indices/FlowFrameworkIndicesHandler.java
@@ -412,8 +412,8 @@ public class FlowFrameworkIndicesHandler {
         }
         doesTemplateExist(documentId, templateExists -> {
             if (templateExists) {
-                isWorkflowProvisioned(documentId, workflowIsProvisioned -> {
-                    if (workflowIsProvisioned) {
+                isWorkflowNotStarted(documentId, workflowIsNotStarted -> {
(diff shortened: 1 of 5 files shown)


owaiskazi19 (approved):
LGTM

joshpalis (approved):

PR #477, opened 2024-01-30
Fixing security tests

joshpalis:
### Description
Enabled security integration tests using `-Dhttps` flag. Now security enabled integration tests will run with the following commands 

- ./gradlew integTest -Dhttps=true -Duser=admin -Dpassword=admin
- ./gradlew integTest -Dsecurity.enabled=true

### Issues Resolved
Resolves https://github.com/opensearch-project/flow-framework/issues/469

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -216,6 +216,9 @@ integTest {
         is_https = is_https == null ? 'true' : is_https
         user = user == null ? 'admin' : user
         password = password == null ? 'admin' : password
+        System.setProperty("https", is_https)
+        System.setProperty("user", user)
+        System.setProperty("password", password)
     }
     systemProperty('https', is_https)
     systemProperty('user', user)
@@ -229,14 +232,14 @@ integTest {
     }
 
     // Exclude integration tests that require security plugin
-    if (System.getProperty("security.enabled") == null || System.getProperty("security.enabled") == "false") {
+    if (System.getProperty("https") == null || System.getProperty("https") == "false") {
         filter {
             excludeTestsMatching "org.opensearch.flowframework.rest.FlowFrameworkSecureRestApiIT"
         }
     }
 
     // Include only secure integration tests in security enabled clusters
-    if (System.getProperty("security.enabled") != null && System.getProperty("security.enabled") == "true") {
+    if (System.getProperty("https") != null && System.getProperty("https") == "true") {
         filter {
             includeTestsMatching "org.opensearch.flowframework.rest.FlowFrameworkSecureRestApiIT"
             excludeTestsMatching "org.opensearch.flowframework.rest.FlowFrameworkRestApiIT"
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -229,14 +232,14 @@ integTest {
...
-    if (System.getProperty("security.enabled") == null || System.getProperty("security.enabled") == "false") {
+    if (System.getProperty("https") == null || System.getProperty("https") == "false") {
         filter {
             excludeTestsMatching "org.opensearch.flowframework.rest.FlowFrameworkSecureRestApiIT"
         }
     }
 
     // Include only secure integration tests in security enabled clusters
-    if (System.getProperty("security.enabled") != null && System.getProperty("security.enabled") == "true") {
+    if (System.getProperty("https") != null && System.getProperty("https") == "true") {
```
dbwiddis, line 242: Is there a way to do `equalsIgnoreCase()` here (and line 275)?

owaiskazi19:
Thanks for addressing this @joshpalis.
QQ: How does security tests work for other plugins with `security.enabled` flag?

dbwiddis:
In the issue summary, add "Fixes" or "Resolves" prior to the issue number you're fixing.  That will auto-link the issue to close it when the PR is merged.

build.gradle
joshpalis: I dont think its necessary, all documentation specifically points to using `https` and no other plugin allows for `HTTPS`

dbwiddis (approved):
LGTM!  Is there a way to re-trigger the build that autocut the issue to verify the fix?

joshpalis:
> QQ: How does security tests work for other plugins with security.enabled flag?

Other plugins such as KNN only pull in the security plugin when the `security.enabled` flag is set. However they do not filter tests based on this flag. In our case, when running with security enabled, we only include the secure integration tests. The issue was that the distribution build job only passes the `https` flag, the `security.enabled` flag is something that we added for our plugin, so our non-security enabled tests were not filtered out. These changes ensure that non-security tests are filtered out based on the `https` flag instead of the `security.enabled` flag

owaiskazi19:
> Other plugins such as KNN only pull in the security plugin when the `security.enabled` flag is set. However they do not filter tests based on this flag. In our case, when running with security enabled, we only include the secure integration tests. The issue was that the distribution build job only passes the `https` flag, the `security.enabled` flag is something that we added for our plugin, so our non-security enabled tests were not filtered out. These changes ensure that non-security tests are filtered out based on the `https` flag instead of the `security.enabled` flag

The reason KNN and other plugins work with `security-enabled` flag is because they have the same integ tests running for security plugin enabled and disabled case. Whereas we have different tests for both. Instead, we should have the same test cases by only changing the url of the cluster to `http` or `https`. 
Also, the https flag is set [here](https://github.com/opensearch-project/opensearch-build/blob/main/scripts/default/integtest.sh#L105) from the build repo.
I am fine merging this for 2.12 but for the next release we can incorporate https://github.com/opensearch-project/flow-framework/issues/478.

owaiskazi19 (approved):

owaiskazi19:
> LGTM! Is there a way to re-trigger the build that autocut the issue to verify the fix?

Builds run once a day.

owaiskazi19:
> In the issue summary, add "Fixes" or "Resolves" prior to the issue number you're fixing. That will auto-link the issue to close it when the PR is merged.

For autocut issue we should avoid using fixes or closes because if the build passes, autocut bot itself closes the issue.

dbwiddis:
> For autocut issue we should avoid using fixes or closes because if the build passes, autocut bot itself closes the issue.

Or we can close it and autocut will reopen it when it fails again...

PR #482, opened 2024-01-31
Filter out system indices from deletion for integration tests

### LINKED ISSUE #469: [AUTOCUT] Integration Test failed for flow-framework: 2.12.0 tar distribution
The integration test failed at distribution level for component flow-framework<br>Version: 2.12.0<br>Distribution: tar<br>Architecture: x64<br>Platform: linux<br><br>Please check the logs: https://build.ci.opensearch.org/job/integ-test/7355/display/redirect<br><br> * Test-report manifest:*<br> - https://ci.opensearch.org/ci/dbc/integ-test/2.12.0/9282/linux/x64/tar/test-results/7355/integ-test/test-report.yml <br><br> _Note: Steps to reproduce, additional logs and other files can be found within the above test-report manifest. <br>Instructions of this test-report manifest can be found [here](https://github.com/opensearch-project/opensearch-build/tree/main/src/report_workflow#guide-on-test-report-manifest-from-ci)._

joshpalis:
### Description

The autocut issue shows that the error comes from attempting to delete protected system indices registered [here](https://github.com/opensearch-project/security/blob/2dab119387212d5047062100f799934b7ffbffea/tools/install_demo_configuration.sh#L388). I suspect that setting is applied during the distribution build, but it was ommitted here in our tests in order to delete these indices (in an attempt to mitigate flaky tests that come from lack of storage). This PR adds those settings back and filters out the ML/ Flow Framework system indices from deletion
```
2> REPRODUCE WITH: ./gradlew ':integTest' --tests "org.opensearch.flowframework.rest.FlowFrameworkSecureRestApiIT.testCreateWorkflowWithReadAccess" -Dtests.seed=D479F832B8590D9D -Dtests.security.manager=false -Dtests.jvm.argline="-XX:TieredStopAtLevel=1 -XX:ReservedCodeCacheSize=64m" -Dtests.locale=zh-Hans-SG -Dtests.timezone=Pacific/Tahiti -Druntime.java=21

  2> org.opensearch.client.ResponseException: method [DELETE], host [https://localhost:9200/], URI [/.plugins-ml-model-group], status line [HTTP/1.1 403 Forbidden]

    {"error":{"root_cause":[{"type":"security_exception","reason":"no permissions for [] and User [name=admin, backend_roles=[admin], requestedTenant=null]"}],"type":"security_exception","reason":"no permissions for [] and User [name=admin, backend_roles=[admin], requestedTenant=null]"},"status":403}
```

### Issues Resolved
#469 

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -216,6 +216,9 @@ integTest {
         is_https = is_https == null ? 'true' : is_https
         user = user == null ? 'admin' : user
         password = password == null ? 'admin' : password
+        System.setProperty("https", is_https)
+        System.setProperty("user", user)
+        System.setProperty("password", password)
     }
     systemProperty('https', is_https)
     systemProperty('user', user)
@@ -229,14 +232,14 @@ integTest {
     }
 
     // Exclude integration tests that require security plugin
-    if (System.getProperty("security.enabled") == null || System.getProperty("security.enabled") == "false") {
+    if (System.getProperty("https") == null || System.getProperty("https") == "false") {
         filter {
             excludeTestsMatching "org.opensearch.flowframework.rest.FlowFrameworkSecureRestApiIT"
         }
     }
 
     // Include only secure integration tests in security enabled clusters
-    if (System.getProperty("security.enabled") != null && System.getProperty("security.enabled") == "true") {
+    if (System.getProperty("https") != null && System.getProperty("https") == "true") {
         filter {
             includeTestsMatching "org.opensearch.flowframework.rest.FlowFrameworkSecureRestApiIT"
             excludeTestsMatching "org.opensearch.flowframework.rest.FlowFrameworkRestApiIT"
@@ -269,7 +272,7 @@ testClusters.integTest {
     testDistribution = "ARCHIVE"
 
     // Optionally install security
-    if (System.getProperty("security.enabled") != null && System.getProperty("security.enabled") == "true") {
+    if (System.getProperty("https") != null && System.getProperty("https") == "true") {
         // Retrieve Security Plugin Zip from zipArchive
         configurations.secureIntegTestPluginArchive.asFileTree.each {
             if(it.name.contains("opensearch-security")) {
@@ -376,6 +379,9 @@ task integTestRemote(type: RestIntegTestTask) {
         is_https = is_https == null ? 'true' : is_https
         user = user == null ? 'admin' : user
         password = password == null ? 'admin' : password
+        System.setProperty("https", is_https)
+        System.setProperty("user", user)
+        System.setProperty("password", password)
     }
     systemProperty('https', is_https)
     systemProperty('user', user)
--- a/build.gradle
+++ b/build.gradle
@@ -333,7 +333,8 @@ testClusters.integTest {
                 '".plugins-ml-config", ' +
                 '".plugins-ml-connector", ' +
                 '".plugins-ml-model-group", ' +
-                '".plugins-ml-model", ".plugins-ml-task", ' +
+                '".plugins-ml-model", ' +
+                '".plugins-ml-task", ' +
                 '".plugins-ml-conversation-meta", ' +
                 '".plugins-ml-conversation-interactions", ' +
                 '".plugins-flow-framework-config", ' +
--- a/build.gradle
+++ b/build.gradle
@@ -336,21 +336,6 @@ testClusters.integTest {
                 '".plugins-ml-model", ".plugins-ml-task", ' +
                 '".plugins-ml-conversation-meta", ' +
                 '".plugins-ml-conversation-interactions", ' +
-                '".opendistro-alerting-config", ' +
-                '".opendistro-alerting-alert*", ' +
-                '".opendistro-anomaly-results*", ' +
-                '".opendistro-anomaly-detector*", ' +
-                '".opendistro-anomaly-checkpoints", ' +
-                '".opendistro-anomaly-detection-state", ' +
-                '".opendistro-reports-*", ' +
-                '".opensearch-notifications-*", ' +
-                '".opensearch-notebooks", ' +
-                '".opensearch-observability", ' +
-                '".ql-datasources", ' +
-                '".opendistro-asynchronous-search-response*", ' +
-                '".replication-metadata-store", ' +
-                '".opensearch-knn-models", ' +
-                '".geospatial-ip2geo-data*", ' +
                 '".plugins-flow-framework-config", ' +
                 '".plugins-flow-framework-templates", ' +
                 '".plugins-flow-framework-state"' +
--- a/build.gradle
+++ b/build.gradle
@@ -329,6 +329,33 @@ testClusters.integTest {
         setting("plugins.security.authcz.admin_dn", "\n- CN=kirk,OU=client,O=client,L=test, C=de")
         setting('plugins.security.restapi.roles_enabled', '["all_access", "security_rest_api_access"]')
         setting('plugins.security.system_indices.enabled', "true")
+        setting('plugins.security.system_indices.indices', '[' +
+                '".plugins-ml-config", ' +
+                '".plugins-ml-connector", ' +
+                '".plugins-ml-model-group", ' +
+                '".plugins-ml-model", ".plugins-ml-task", ' +
+                '".plugins-ml-conversation-meta", ' +
+                '".plugins-ml-conversation-interactions", ' +
+                '".opendistro-alerting-config", ' +
+                '".opendistro-alerting-alert*", ' +
+                '".opendistro-anomaly-results*", ' +
+                '".opendistro-anomaly-detector*", ' +
+                '".opendistro-anomaly-checkpoints", ' +
+                '".opendistro-anomaly-detection-state", ' +
+                '".opendistro-reports-*", ' +
+                '".opensearch-notifications-*", ' +
+                '".opensearch-notebooks", ' +
+                '".opensearch-observability", ' +
+                '".ql-datasources", ' +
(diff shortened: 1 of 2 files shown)


joshpalis:
@dbwiddis In non-security enabled integration /multi-node tests these should still be deleted, I can modify this such that this filtering only happens when the https flag is set

build.gradle
```diff
@@ -329,6 +329,33 @@ testClusters.integTest {
         setting("plugins.security.authcz.admin_dn", "\n- CN=kirk,OU=client,O=client,L=test, C=de")
         setting('plugins.security.restapi.roles_enabled', '["all_access", "security_rest_api_access"]')
         setting('plugins.security.system_indices.enabled', "true")
+        setting('plugins.security.system_indices.indices', '[' +
```
dbwiddis, line 332: I'm not sure how adding this list here relates to limiting what we delete in the test case class.   We're mostly adding stuff that we will ignore, and a bunch of indices unrelated to our tests that won't change.

If we need this set, Is there a way to fetch this programmatically?  Copy-pasting [a list from another plugin](https://github.com/opensearch-project/security/blob/2dab119387212d5047062100f799934b7ffbffea/tools/install_demo_configuration.sh#L388C3-L388C62) that may change in the future is brittle.   It looks like that script adds them to the `opensearch.yml` file can we read them from there?  Or better yet at run-time from the settings?

If we keep this as-is, please add a comment saying where these come from so future-you and future-me know when a test breaks where to re-sync this list.

src/test/java/org/opensearch/flowframework/FlowFrameworkRestTestCase.java
```diff
@@ -295,8 +295,9 @@ protected void wipeAllODFEIndices() throws IOException {
                 // Do not reset ML/Flow Framework encryption index as this is needed to encrypt connector credentials
                 if (indexName != null
                     && !".opendistro_security".equals(indexName)
-                    && !".plugins-ml-config".equals(indexName)
-                    && !".plugins-flow-framework-config".equals(indexName)) {
+                    && !indexName.startsWith(".opensearch")
+                    && !indexName.startsWith(".plugins-flow-framework")
+                    && !indexName.startsWith(".plugins-ml")) {
```
dbwiddis, line 300: This comprises most of the list above and pretty much all of the indices that our tests modify, so I'm missing the value in deleting them in between tests.

build.gradle
joshpalis: This list comes from the security demo configuration [here](https://github.com/opensearch-project/security/blob/2dab119387212d5047062100f799934b7ffbffea/tools/install_demo_configuration.sh#L388) which is applied to a security enabled integration test cluster. We have to manually set this setting since we're not using the demo script.

src/test/java/org/opensearch/flowframework/FlowFrameworkRestTestCase.java
joshpalis: I have added a recent commit which only does this filtering when the https flag is set. In regular integration test runs in which we do not have security enabled, these indices will be deleted.

dbwiddis:
> In non-security enabled integration /multi-node tests these should still be deleted, 

Can we more narrowly target what we actually need to delete?  The `.ml-commons-config` index takes 20 seconds to recreate itself and has caused all sorts of flakiness that I worked around in https://github.com/opensearch-project/flow-framework/pull/465/commits/1fb3140b64249928492f71aa779c4b1649935e68 so I'd really like to leave that untouched.  Debug-logging cat-indices between tests could show us which index is filling up space and we can just delete that one.

I am still not sure we're really being disk-limited here and if we are let's find a better way than wiping everything.

> I can modify this such that this filtering only happens when the https flag is set

Definitely let's include that as part of the logic, but I'm wondering here if we need security-enabled multi-node tests?

build.gradle
owaiskazi19: Can we try something like CCR plugin [here](https://github.com/opensearch-project/cross-cluster-replication/blob/3ea239b5b308b8c053511fb2da79ac37f12b2340/build.gradle#L685) to run the demo script? Otherwise, we would need to keep the track of the system indices in the demo file and update our build.gradle.

joshpalis: CCR applies the security config in the same way we are [here](https://github.com/opensearch-project/cross-cluster-replication/blob/3ea239b5b308b8c053511fb2da79ac37f12b2340/build.gradle#L573), though they have omitted protected indices other than their own.

src/test/java/org/opensearch/flowframework/FlowFrameworkRestTestCase.java
dbwiddis: Is it possible to just add this part and NOT add all the other stuff in the build?

joshpalis: It's possible, but that would leave the risk of facing this issue again in the future since our security configuration would not have protected indices. Since the distribution build uses this configuration, we should have it too to catch this issue early on.

dbwiddis (approved):

build.gradle
owaiskazi19: Yes, that's what we were trying to do with just our indices and ml-commons. I am not sure if adding other plugins indices would be the right way to go around here.

joshpalis: Ill remove them

src/test/java/org/opensearch/flowframework/FlowFrameworkRestTestCase.java
```diff
@@ -292,17 +292,27 @@ protected void wipeAllODFEIndices() throws IOException {
...
                 String indexName = (String) index.get("index");
-                // Do not reset ML/Flow Framework encryption index as this is needed to encrypt connector credentials
-                if (indexName != null
-                    && !".opendistro_security".equals(indexName)
-                    && !".plugins-ml-config".equals(indexName)
-                    && !".plugins-flow-framework-config".equals(indexName)) {
-                    adminClient().performRequest(new Request("DELETE", "/" + indexName));
+                if (indexName != null) {
+                    // Filter out system index deletion only when running security enabled tests
+                    if (isHttps() && isSystemIndex(indexName)) {
```
dbwiddis, line 297: This conditional can be wrapped into the one above (negating it).

src/test/java/org/opensearch/flowframework/FlowFrameworkRestTestCase.java
```diff
@@ -292,17 +292,27 @@ protected void wipeAllODFEIndices() throws IOException {
...
+                        continue;
+                    }
+                    // Do not reset ML/Flow Framework encryption index as this is needed to encrypt connector credentials
+                    if (!".plugins-ml-config".equals(indexName) && !".plugins-flow-framework-config".equals(indexName)) {
```
dbwiddis, line 301: we'll never get here since these are both system indices, right?

dbwiddis (approved):

src/test/java/org/opensearch/flowframework/FlowFrameworkRestTestCase.java
joshpalis: In a non-security enabled cluster this will ensure the encryption indices are still filtered out

build.gradle
```diff
@@ -329,6 +329,18 @@ testClusters.integTest {
...
+                '".plugins-ml-config", ' +
+                '".plugins-ml-connector", ' +
+                '".plugins-ml-model-group", ' +
+                '".plugins-ml-model", ".plugins-ml-task", ' +
```
owaiskazi19, line 336:
```suggestion
                '".plugins-ml-model",  ' +
                ".plugins-ml-task",
```


or if we are using all the system indices of ml-commons here we can do `".plugins-ml-*`. Similarly for flow framework indices.

joshpalis: I can make the suggested change, but I'd rather not use `.plugins-ml-*` since we're trying to emulate the actual configuration of the demo installation config

owaiskazi19 (approved):
Let's get this in and not resolve the issue. If this fix is able to get a successful build, then autocut bot would resolve it's own. Thanks @joshpalis

dbwiddis:
> Let's get this in and not resolve the issue. If this fix is able to get a successful build, then autocut bot would resolve on it's own. 

The build is not even testing flow framework on 3.0.0 branch.  Does it not test if the issue is open?  Or is it only testing on 2.12.0 which runs daily?

We closed it earlier and it auto-reopened the issue when it failed, so I know that path works.

owaiskazi19:
> The build is not even testing flow framework on 3.0.0 branch. Does it not test if the issue is open? Or is it only testing on 2.12.0 which runs daily?
Since we didn't add flow framework to 3.0.0 [manifest](https://github.com/opensearch-project/opensearch-build/blob/main/manifests/3.0.0/opensearch-3.0.0.yml). It's not running for 3.0. We have added it only for [2.12](https://github.com/opensearch-project/opensearch-build/blob/main/manifests/2.12.0/opensearch-2.12.0.yml#L167).

> We closed it earlier and it auto-reopened the issue when it failed, so I know that path works.
If the build passes for 2.12 which runs once a day, the issue will be auto-closed.

dbwiddis:
Latest 2.12.0 run did not update for flow-framework:
ARM: https://build.ci.opensearch.org/blue/organizations/jenkins/integ-test/detail/integ-test/7417/pipeline/
x64: https://build.ci.opensearch.org/blue/organizations/jenkins/integ-test/detail/integ-test/7418/pipeline/

dbwiddis:
OK, looks like it finally ran and closed the issue.

PR #498, opened 2024-02-06
Add sample templates

### LINKED ISSUE #356: [DISCUSS] Provide sample template to users without needing docs

Customers would like to use a sample template without copying and pasting from documentation.

There are many alternatives. We should decide on one that will support front end search/integration.
 - REST API (list / query sample templates)?
 - Folder on the class path?
 - A better idea?

dbwiddis:
### Description

Adds sample templates for users to use as a basis for their workflows

### Issues Resolved

Resolves #356 

### DIFF
--- a/sample-templates/README.md
+++ b/sample-templates/README.md
@@ -1,0 +1,33 @@
+## Flow Framework Sample Templates
+
+This folder contains sample workflow templates that can be used with Flow Framework.
+
+Each template is provided in both YAML and JSON format with identical functionality.
+The YAML templates include comments which give more insight into the template's usage.
+Use the corresponding `Content-Type` (`application/yaml` or `application/json`) when providing them as the body of a REST request.
+
+You will need to update the `credentials` field with appropriate API keys.
+
+To create a workflow and provision the resources:
+
+```
+POST /_plugins/_flow_framework/workflow?provision=true
+{ template as body }
+```
+
+This will return a `workflow_id`. To get the IDs of created resources, call the workflow status API.
+
+```
+GET /_plugins/_flow_framework/workflow/<workflow_id>/_status
+```
+
+For the Query Assist Agent API, the `agent_id` of the `root_agent` can be used to query it.
+
+```
+POST /_plugins/_ml/agents/_<agent_id>/_execute
+{
+  "parameters": {
+    "question": "How many 5xx logs do I have?"
+  }
+}
+```
--- a/sample-templates/deploy-openai-model.json
+++ b/sample-templates/deploy-openai-model.json
@@ -22,7 +22,8 @@
             "protocol": "http",
             "parameters": {
               "endpoint": "api.openai.com",
-              "model": "gpt-3.5-turbo"
+              "model": "gpt-3.5-turbo",
+              "response_filter": "$.choices[0].message.content"
             },
             "credential": {
               "openAI_key": "PUT_YOUR_API_KEY_HERE"
--- a/sample-templates/deploy-openai-model.json
+++ b/sample-templates/deploy-openai-model.json
@@ -1,0 +1,59 @@
+{
+  "name": "Deploy OpenAI Model",
+  "description": "Deploy a model using a connector to OpenAI",
+  "use_case": "PROVISION",
+  "version": {
+    "template": "1.0.0",
+    "compatibility": [
+      "2.12.0",
+      "3.0.0"
+    ]
+  },
+  "workflows": {
+    "provision": {
+      "nodes": [
+        {
+          "id": "create_openai_connector",
+          "type": "create_connector",
+          "user_inputs": {
+            "name": "OpenAI Chat Connector",
+            "description": "The connector to public OpenAI model service for GPT 3.5",
+            "version": "1",
+            "protocol": "http",
+            "parameters": {
+              "endpoint": "api.openai.com",
+              "model": "gpt-3.5-turbo"
+            },
+            "credential": {
+              "openAI_key": "PUT_YOUR_API_KEY_HERE"
+            },
+            "actions": [
+              {
+                "action_type": "predict",
+                "method": "POST",
+                "url": "https://${parameters.endpoint}/v1/chat/completions"
+              }
+            ]
+          }
+        },
+        {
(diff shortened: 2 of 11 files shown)


sample-templates/deploy-sagemaker-mistral-model.json
```diff
@@ -0,0 +1,64 @@
...
+              "service_name": "sagemaker"
+            },
+            "actions": [
+              {
+                "action_type": "predict",
+                "method": "POST",
+                "headers": {
+                  "content-type": "application/json"
+                },
+                "url": "https://runtime.sagemaker.us-east-1.amazonaws.com/endpoints/PUT_YOUR_ENDPOINT_HERE/invocations",
```
joshuali925, line 38: this is a custom sagemaker endpoint, it shouldn't be limit to mistral?

sample-templates/deploy-openai-model.json
```diff
@@ -0,0 +1,59 @@
...
+        {
+          "id": "create_openai_connector",
+          "type": "create_connector",
+          "user_inputs": {
+            "name": "OpenAI Chat Connector",
+            "description": "The connector to public OpenAI model service for GPT 3.5",
+            "version": "1",
+            "protocol": "http",
+            "parameters": {
+              "endpoint": "api.openai.com",
```
joshuali925, line 24:
```suggestion
              "endpoint": "api.openai.com",
              "response_filter": "$.choices[0].message.content",
```

sample-templates/query-assist-agent.json
```diff
@@ -0,0 +1,299 @@
...
+            "register_t2ppl_model": "model_id"
+          },
+          "user_inputs": {
+            "type": "PPLTool",
+            "name": "TransferQuestionToPPLAndExecuteTool",
+            "description": "Use this tool to transfer natural language to generate PPL and execute PPL to query inside. Use this tool after you know the index name, otherwise, call IndexRoutingTool first. The input parameters are: {index:IndexName, question:UserQuestion}",
+            "parameters": {
+              "response_filter": "$.completion",
+              "execute": false,
+              "prompt": "Below is an instruction that describes a task, paired with the index and corresponding fields that provides further context. Write a response that appropriately completes the request.\n\n### Instruction:\nI have an opensearch index with fields in the following. Now I have a question: ${indexInfo.question} Can you help me generate a PPL for that?\n\n### Index:\n${indexInfo.indexName}\n\n### Fields:\n${indexInfo.mappingInfo}\n\n### Response:\n"
```
joshuali925, line 116: the claude prompt is here: https://github.com/opensearch-project/skills/pull/184/files#diff-778297aed573eb9f6ee67ce363e0159c095a9d6bd8ebebf8313aa90bae242249R2

once the PR is merged, this line can be changed to 
```suggestion
              "model_type": "claude"
```

joshuali925 (commented):
do we also need to give query assist with openai? if so i can provide something

i'll update the flow template in https://github.com/opensearch-project/dashboards-assistant/blob/0de2c9be98a673775441680370933b0ae9f7421e/GETTING_STARTED_GUIDE.md to point to this once this is merged, or is there another better reference?

dbwiddis:
> do we also need to give query assist with openai? if so i can provide something

Good question.  Can we do it with only openai and without ppl tool (sagemaker)?   I figured the Bedrock+Sagemaker went together well.

sample-templates/deploy-sagemaker-mistral-model.json
dbwiddis: Can you recommend what to put here?   In the YAML I had comments about the user needing to change it, JSON doesn't allow comments.

joshuali925: i think this can be something general like "Sagemaker Custom Model"

joshuali925:
> Can we do it with only openai and without ppl tool (sagemaker)? I figured the Bedrock+Sagemaker went together well.

We can do it with openai model (same one you submitted) and `PPLTool` that uses the openai model, `PPLTool` is needed but not through sagemaker

dbwiddis:
> We can do it with openai model (same one you submitted) and `PPLTool` that uses the openai model, `PPLTool` is needed but not through sagemaker

OK I'll modify the whole template to only use OpenAI!

sample-templates/README.md
```diff
@@ -0,0 +1,33 @@
+## Flow Framework Sample Templates
+
+This folder contains sample workflow templates that can be used with Flow Framework.
+
```
amitgalitz, line 4: I know that we have some comments in the YAML template as you point below, but should we maybe have some info here just explaining the template more in depth or will this be in the documentation website?

For example as you also mention in YAML but to explain the edgs vs using previous nodes inputs.
Explaining that the type refers to whatever step and ID is set by user and can be anything, that user_input here corresponds to that API's payload.

It can still be a little overwhelming looking at the big template payload all together

sample-templates/deploy-openai-model.json
amitgalitz: @joshuali925 just wondering is this a new optional parameter on connector API? I don't see it any of the examples on documentation website

sample-templates/README.md
dbwiddis: It's all explained in detail in the actual documentation on opensearch.org including a tutorial building it up piece by piece.  

https://github.com/opensearch-project/documentation-website/pull/6257/files/ba09447a5b3a48975a1759c4d16fa5b24d627bc0..8a1dca85a60a5cc6ce6d808c4cc6627c8b565636#diff-1007a4ccfce8e754061e7bf9a56298cf9a0832a39497bc09757c8b36a62292e9

owaiskazi19 (approved):

joshpalis (approved):

sample-templates/deploy-openai-model.json
joshuali925: see https://github.com/opensearch-project/ml-commons/blob/79b9dfd7eeef051fbf9d0d35242052433b83b297/docs/tutorials/agent_framework/build_your_own_chatbot.md#31-create-agent, it's the same parameter for agent/tools

PR #515, opened 2024-02-09
Improve provisioning exceptions with step ID and hide exception details

dbwiddis:
### Description

Handles exceptions during the provisioning workflow better:
 - includes the workflow step ID that caused the exception
 - Includes only the exception class type but not the exception message details


joshpalis (approved):

owaiskazi19 (approved):
Looks fine for the quick fix

Issue #395, opened 2024-01-10
[FEATURE] Move workflow-steps.json to a Java class/enum


amitgalitz:
### Is your feature request related to a problem?
Currently we store lots of information regarding the workflow steps in a json file, we parse this json file for validation and other use cases. We also will soon add a new API that fetches the information in this json file. Because this json file is parsed in runtime, we can't use any of the information here during compile time to help our code be more stable. 

### What solution would you like?
We should store the information in either an Enum or a class. Some of the benefits are:
1. Type safe with predefined constants
2. compile time checking, reducing potential runtime errors and overhead for test cases
3. less processing time spent on parsing json

For example the code could look something like this:
```
  CREATE_CONNECTOR(
        "create_connector",
        Arrays.asList("name ", "description", "version", "protocol", "parameters", "credential", "actions"),
        Arrays.asList("connector_id"),
        "connector_id"
    ),

    REGISTER_REMOTE_MODEL(
        "register_remote_model",
        Arrays.asList("name ", "function_name", "connector_id"),
        Arrays.asList("model_id", "register_model_status"),
                "model_id"
                );
...

    WorkflowStepsDefinition(String workflowStep, List<String> inputs, List<String> outputs, @Nullable String resourceCreated) {
        this.workflowStep = workflowStep;
        this.inputs = List.copyOf(inputs);
        this.outputs = List.copyOf(outputs);
        this.resourceCreated = resourceCreated;
    }
    
```

dbwiddis:
If we make it an enum we should integrate it with the workflow step factory so there's one place for people to add new capability.  New enum value, along with the different components.

PR #523, opened 2024-02-15
Moved workflow-steps.json to Enum

owaiskazi19:
### Description
Moved workflow-steps.json to Enum and removed additional parsing required for the workflow-steps.json. Also, removed the json file.
 
~TODO:~
~1. Fix tests~

### Issues Resolved
Resolves https://github.com/opensearch-project/flow-framework/issues/395


src/main/java/org/opensearch/flowframework/workflow/WorkflowStepFactory.java
```diff
@@ -75,6 +88,231 @@ public WorkflowStepFactory(
...
+     * Enum encapsulating the different step names, their inputs, outputs, required plugin and timeout of the step
+     */
+
+    public enum WorkflowSteps {
+
+        NOOP("noop", Collections.emptyList(), Collections.emptyList(), Collections.emptyList(), null),
+
+        CREATE_CONNECTOR(
+            "create_connector",
+            Arrays.asList("name", "description", "version", "protocol", "parameters", "credential", "actions"),
```
joshpalis, line 101: Can we use the CommonValue strings here in these definitions?

src/main/java/org/opensearch/flowframework/workflow/WorkflowStepFactory.java
```diff
@@ -75,6 +88,231 @@ public WorkflowStepFactory(
...
+        };
+
+        /**
+         * Gets the timeout based on the workflowStep.
+         * @param workflowStep workflow step type
+         * @return the resource that will be created
+         * @throws FlowFrameworkException if workflow step doesn't exist in enum
+         */
+        public static TimeValue getTimeoutByWorkflowType(String workflowStep) throws FlowFrameworkException {
+            if (workflowStep != null && !workflowStep.isEmpty()) {
```
dbwiddis, line 235: o.o.core.common.Strings has you covered here:
```suggestion
            if (!Strings.isNullOrEmpty(workflowStep) {
```

src/main/java/org/opensearch/flowframework/workflow/WorkflowStepFactory.java
```diff
@@ -75,6 +88,231 @@ public WorkflowStepFactory(
...
+                for (WorkflowSteps mapping : values()) {
+                    if (workflowStep.equals(mapping.getWorkflowStep())) {
+                        return mapping.getTimeout();
+                    }
+                }
+            }
+            logger.error("Unable to find workflow timeout for step: {}", workflowStep);
+            throw new FlowFrameworkException("Unable to find workflow timeout for step: " + workflowStep, RestStatus.BAD_REQUEST);
```
dbwiddis, line 243: Suggeest you create the string once and use it in both these places.  See #524 for dozens of examples.

src/main/java/org/opensearch/flowframework/workflow/WorkflowStepFactory.java
```diff
@@ -75,6 +88,231 @@ public WorkflowStepFactory(
         stepMap.put(DeleteAgentStep.NAME, () -> new DeleteAgentStep(mlClient));
     }
 
+    /**
+     * Enum encapsulating the different step names, their inputs, outputs, required plugin and timeout of the step
+     */
+
+    public enum WorkflowSteps {
```
dbwiddis, line 95: If you want to leave the stepMap above as-is, there's no reason to have this enum in this class, and it can be put in its own class.  (We could change the map key to the enum value and make it an EnumMap.)

If you want to keep it in this class, we can add a`Supplier<WorkflowStep>` field to each enum value (which would be a good use of the `workflowStep` name you've presently given to a String...) and get rid of the map.

src/main/java/org/opensearch/flowframework/workflow/WorkflowStepFactory.java
```diff
@@ -75,6 +88,231 @@ public WorkflowStepFactory(
...
+            return outputs;
+        }
+
+        public List<String> getRequiredPlugins() {
+            return requiredPlugins;
+        }
+
+        public TimeValue getTimeout() {
+            return timeout;
+        }
```
dbwiddis, line 222: Since these are all final/immutable, let's use record syntax rather than getters here (`workflowStep()`, `inputs()`, etc.)

src/main/java/org/opensearch/flowframework/workflow/WorkflowStepFactory.java
```diff
@@ -75,6 +88,231 @@ public WorkflowStepFactory(
...
+    public enum WorkflowSteps {
+
+        NOOP("noop", Collections.emptyList(), Collections.emptyList(), Collections.emptyList(), null),
+
+        CREATE_CONNECTOR(
+            "create_connector",
+            Arrays.asList("name", "description", "version", "protocol", "parameters", "credential", "actions"),
+            Arrays.asList("connector_id"),
+            Arrays.asList("opensearch-ml"),
+            new TimeValue(60, SECONDS)
```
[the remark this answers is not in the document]
owaiskazi19: I think you meant `TimeValue.timeValueSeconds(60)`

src/main/java/org/opensearch/flowframework/workflow/WorkflowStepFactory.java
owaiskazi19: This was really a great idea. I removed the stepMap by adding a new field in the enum but just initialized the map in the constructor to avoid breaking anything.

src/main/java/org/opensearch/flowframework/workflow/ToolStep.java
```diff
@@ -31,6 +31,7 @@
 /**
  * Step to register a tool for an agent
  */
+@SuppressWarnings("unchecked")
```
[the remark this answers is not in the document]
owaiskazi19: Yes, it's possible and pushed the change.

dbwiddis (approved):
LGTM! Thanks for taking the suggestions, I like how you did the map.

joshpalis (approved):

amitgalitz:
Should follow up here be to combine the WorkflowResources to this too?

owaiskazi19:
> Should follow up here be to combine the WorkflowResources to this too?

WorkflowResources is not part of this issue. Can you create a separate issue for the same?

PR #524, opened 2024-02-16
Refactored logging for consistency

dbwiddis:
### Description

Adds consistent treatment for log error messages across the repo.  
 - Standardizes the log message to match the value returned to the user (without exception details) 
 - Adds exception stack traces in unexpected error situations
 - Adds more details (steps, ids, etc) when possible.

### DIFF
--- a/src/main/java/org/opensearch/flowframework/indices/FlowFrameworkIndicesHandler.java
+++ b/src/main/java/org/opensearch/flowframework/indices/FlowFrameworkIndicesHandler.java
@@ -180,19 +180,20 @@ public class FlowFrameworkIndicesHandler {
                 @SuppressWarnings("deprecation")
                 ActionListener<CreateIndexResponse> actionListener = ActionListener.wrap(r -> {
                     if (r.isAcknowledged()) {
-                        logger.info("create index:{}", indexName);
+                        logger.info("create index: {}", indexName);
                         internalListener.onResponse(true);
                     } else {
                         internalListener.onResponse(false);
                     }
                 }, e -> {
-                    logger.error("Failed to create index {}", indexName, e);
-                    internalListener.onFailure(new FlowFrameworkException(e.getMessage(), ExceptionsHelper.status(e)));
+                    String errorMessage = "Failed to create index " + indexName;
+                    logger.error(errorMessage);
+                    internalListener.onFailure(new FlowFrameworkException(errorMessage, ExceptionsHelper.status(e)));
                 });
                 CreateIndexRequest request = new CreateIndexRequest(indexName).mapping(mapping).settings(indexSettings);
                 client.admin().indices().create(request, actionListener);
             } else {
-                logger.debug("index:{} is already created", indexName);
+                logger.debug("index: {} is already created", indexName);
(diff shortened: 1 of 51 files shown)


amitgalitz (dismissed):
Let’s review the reason for all the stack trace logging removal again. Would be very hard to debug code without this

dbwiddis:
Flipping to draft mode for further discussion.  I like the standardization that this PR does, and we can decide when we want to add `, exception` to `log.error()` or similar and apply it consistently.

dbwiddis:
@amitgalitz can you please re-review?

> Merging is blocked
> Merging can be performed automatically once the requested changes are addressed.

src/main/java/org/opensearch/flowframework/indices/FlowFrameworkIndicesHandler.java
```diff
@@ -180,19 +180,20 @@ public void initFlowFrameworkIndexIfAbsent(FlowFrameworkIndex index, ActionListe
...
+                        logger.info("create index: {}", indexName);
                         internalListener.onResponse(true);
                     } else {
                         internalListener.onResponse(false);
                     }
                 }, e -> {
-                    logger.error("Failed to create index {}", indexName, e);
-                    internalListener.onFailure(new FlowFrameworkException(e.getMessage(), ExceptionsHelper.status(e)));
+                    String errorMessage = "Failed to create index " + indexName;
+                    logger.error(errorMessage);
```
amitgalitz, line 190: do we not want to log the stack trace in cases like this so we can get more information on why the create index failed?

dbwiddis: Good catch. I did a global search-and-replace of `logger.error(errorMessage)` in the transport actions but missed those outside it.

src/main/java/org/opensearch/flowframework/workflow/DeleteConnectorStep.java
```diff
@@ -69,8 +69,9 @@ public void onResponse(DeleteResponse deleteResponse) {
 
             @Override
             public void onFailure(Exception e) {
-                logger.error("Failed to delete connector");
-                deleteConnectorFuture.onFailure(new FlowFrameworkException(e.getMessage(), ExceptionsHelper.status(e)));
+                String errorMessage = "Failed to delete connector";
```
amitgalitz, line 72: should we add connector ID here? similar for other deletion steps like agent ID and etc.

src/main/java/org/opensearch/flowframework/workflow/UndeployModelStep.java
```diff
@@ -82,8 +82,9 @@ public void onResponse(MLUndeployModelsResponse mlUndeployModelsResponse) {
 
             @Override
             public void onFailure(Exception e) {
-                logger.error("Failed to unldeploy model");
-                undeployModelFuture.onFailure(new FlowFrameworkException(e.getMessage(), ExceptionsHelper.status(e)));
+                String errorMessage = "Failed to undeploy model";
```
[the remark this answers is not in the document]
amitgalitz: or should we relay on ml-commons stack trace if we there is some exception from there client?

amitgalitz (approved):
Thanks for making all these changes across the repo!

Issue #522, opened 2024-02-14
[FEATURE] Add a WorkflowStep for calling external REST APIs


dbwiddis:
### Is your feature request related to a problem?

Coming from https://github.com/opensearch-project/observability/issues/1805

> This API is designed to orchestrate multiple steps, each corresponding to a separate API call to a core API within OpenSearch. The primary goal of the Workflow API is to facilitate complex data preparation processes involving enrichment, aggregation, reindexing, and other transformations.
> 
> By enabling users to define multi-step workflows, this feature aims to simplify the preparation of raw data for visualization, thereby providing deeper insights into system status and health.

An example workflow step for an API call is included in the linked RFC:

```
"steps": [
{
  "name": "add_timestamp_field_to_otel_spans_index",
  "method": "PUT",
  "endpoint": "/${index_name}",
  "body": { ... }
 },
 ....
]
```

### What solution would you like?

Create Workflow Steps which enable calling REST APIs.   

The code already exists in our integration test classes, which could be refactored into appropriate steps in the main source tree:
https://github.com/opensearch-project/flow-framework/blob/13f672e1f210473802b292204bdd558963c9b871/src/test/java/org/opensearch/flowframework/TestHelpers.java#L81-L90

This is called with parameters like this:
https://github.com/opensearch-project/flow-framework/blob/13f672e1f210473802b292204bdd558963c9b871/src/test/java/org/opensearch/flowframework/FlowFrameworkRestTestCase.java#L135-L143

I think the best approach here is to write three new Steps:
1. A `CreateHttpClientStep` which creates an appropriate `client()` to make the call.  This would be set up very similar to how the `ToolStep` is created, using the workflow parameters to configure a `Client` object.  
2. A 'RestApiStep` which  would reference the client in its previous node inputs, and take other params for the actual REST call.
3. A `CloseHttpClientStep` to clean up the client after the workflow finishes executing.

### What alternatives have you considered?

We could write workflow steps for every OpenSearch API (or at least the most common ones) and use the associated Transport Actions.

### Do you have any additional context?

While making REST calls is less efficient than the transport calls, they enable calls outside of OpenSearch as well.  

REST client calls may also be needed for future migration to serverless.

While relatively easy to do this with an HTTP Client, supporting HTTPS and preserving appropriate headers may introduce some security complexities.

YANG-DB:
This looks great 👍 
[...]

PR #530, opened 2024-02-20
Add an HttpHost Workflow Step

dbwiddis:
### Description

Adds a WorkflowStep configuring an HttpHost.  This is similar to the ToolStep in that it simply reflects a configuration which can be re-used in many follow-on steps.

### Issues Resolved

Part of #522 

### DIFF
--- a/src/main/java/org/opensearch/flowframework/workflow/HttpHostStep.java
+++ b/src/main/java/org/opensearch/flowframework/workflow/HttpHostStep.java
@@ -1,0 +1,103 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+package org.opensearch.flowframework.workflow;
+
+import org.apache.hc.core5.http.HttpHost;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.opensearch.action.support.PlainActionFuture;
+import org.opensearch.core.rest.RestStatus;
+import org.opensearch.flowframework.exception.FlowFrameworkException;
+import org.opensearch.flowframework.util.ParseUtils;
+
+import java.util.Collections;
+import java.util.Locale;
+import java.util.Map;
+import java.util.Set;
+
+import static org.opensearch.flowframework.common.CommonValue.HOSTNAME_FIELD;
+import static org.opensearch.flowframework.common.CommonValue.HTTP_HOST_FIELD;
+import static org.opensearch.flowframework.common.CommonValue.PORT_FIELD;
+import static org.opensearch.flowframework.common.CommonValue.SCHEME_FIELD;
+
+/**
+ * Step to register parameters for an HTTP Connection to a Host
+ */
+public class HttpHostStep implements WorkflowStep {
+
+    private static final Logger logger = LogManager.getLogger(HttpHostStep.class);
+    PlainActionFuture<WorkflowData> hostFuture = PlainActionFuture.newFuture();
+    static final String NAME = "http_host";
+
+    @Override
+    public PlainActionFuture<WorkflowData> execute(
+        String currentNodeId,
+        WorkflowData currentNodeInputs,
+        Map<String, WorkflowData> outputs,
+        Map<String, String> previousNodeInputs
+    ) {
+        Set<String> requiredKeys = Set.of(SCHEME_FIELD, HOSTNAME_FIELD, PORT_FIELD);
+        Set<String> optionalKeys = Collections.emptySet();
+
+        try {
+            Map<String, Object> inputs = ParseUtils.getInputsFromPreviousSteps(
+                requiredKeys,
+                optionalKeys,
+                currentNodeInputs,
+                outputs,
+                previousNodeInputs
+            );
+
+            String scheme = validScheme(inputs.get(SCHEME_FIELD));
+            String hostname = inputs.get(HOSTNAME_FIELD).toString();
+            int port = validPort(inputs.get(PORT_FIELD));
+
+            HttpHost httpHost = new HttpHost(scheme, hostname, port);
+
+            hostFuture.onResponse(
+                new WorkflowData(
+                    Map.ofEntries(Map.entry(HTTP_HOST_FIELD, httpHost)),
+                    currentNodeInputs.getWorkflowId(),
+                    currentNodeInputs.getNodeId()
+                )
+            );
+
+            logger.info("Http Host registered successfully {}", httpHost);
+
+        } catch (FlowFrameworkException e) {
+            hostFuture.onFailure(e);
+        }
+        return hostFuture;
+    }
+
+    private String validScheme(Object o) {
+        String scheme = o.toString().toLowerCase(Locale.ROOT);
+        if ("http".equals(scheme) || "https".equals(scheme)) {
+            return scheme;
+        }
+        throw new FlowFrameworkException("http_host scheme must be http or https", RestStatus.BAD_REQUEST);
+    }
+
+    private int validPort(Object o) {
+        try {
+            int port = Integer.parseInt(o.toString());
+            if ((port & 0xffff0000) != 0) {
+                throw new FlowFrameworkException("http_host port number must be between 0 and 65535", RestStatus.BAD_REQUEST);
+            }
+            return port;
+        } catch (NumberFormatException e) {
+            throw new FlowFrameworkException("http_host port must be a number between 0 and 65535", RestStatus.BAD_REQUEST);
+        }
+    }
+
+    @Override
+    public String getName() {
+        return NAME;
+    }
+}
--- a/src/main/java/org/opensearch/flowframework/workflow/HttpHostStep.java
+++ b/src/main/java/org/opensearch/flowframework/workflow/HttpHostStep.java
@@ -40,9 +40,12 @@ public class HttpHostStep implements WorkflowStep {
         String currentNodeId,
         WorkflowData currentNodeInputs,
         Map<String, WorkflowData> outputs,
-        Map<String, String> previousNodeInputs
+        Map<String, String> previousNodeInputs,
+        Map<String, String> params
     ) {
         Set<String> requiredKeys = Set.of(SCHEME_FIELD, HOSTNAME_FIELD, PORT_FIELD);
+        // TODO Possibly add credentials fields here
+        // See ML Commons MLConnectorInput class and its usage
         Set<String> optionalKeys = Collections.emptySet();
 
         try {
@@ -51,11 +54,12 @@ public class HttpHostStep implements WorkflowStep {
                 optionalKeys,
                 currentNodeInputs,
                 outputs,
-                previousNodeInputs
+                previousNodeInputs,
+                params
             );
 
             String scheme = validScheme(inputs.get(SCHEME_FIELD));
-            String hostname = inputs.get(HOSTNAME_FIELD).toString();
+            String hostname = validHostName(inputs.get(HOSTNAME_FIELD));
             int port = validPort(inputs.get(PORT_FIELD));
 
             HttpHost httpHost = new HttpHost(scheme, hostname, port);
@@ -82,6 +86,14 @@ public class HttpHostStep implements WorkflowStep {
             return scheme;
         }
         throw new FlowFrameworkException("http_host scheme must be http or https", RestStatus.BAD_REQUEST);
+    }
+
+    private String validHostName(Object o) {
+        // TODO Add validation:
+        // Prevent use of localhost or private IP address ranges
+        // See ML Commons MLHttpClientFactory.java methods for examples
+        // Possibly consider an allowlist of addresses
+        return o.toString();
     }
 
     private int validPort(Object o) {
--- a/src/test/java/org/opensearch/flowframework/model/WorkflowValidatorTests.java
+++ b/src/test/java/org/opensearch/flowframework/model/WorkflowValidatorTests.java
@@ -11,15 +11,17 @@ package org.opensearch.flowframework.model;
 import org.opensearch.flowframework.common.FlowFrameworkSettings;
 import org.opensearch.flowframework.indices.FlowFrameworkIndicesHandler;
 import org.opensearch.flowframework.workflow.WorkflowStepFactory;
+import org.opensearch.flowframework.workflow.WorkflowStepFactory.WorkflowSteps;
 import org.opensearch.ml.client.MachineLearningNodeClient;
 import org.opensearch.test.OpenSearchTestCase;
 import org.opensearch.threadpool.ThreadPool;
 
 import java.io.IOException;
 import java.util.ArrayList;
-import java.util.HashMap;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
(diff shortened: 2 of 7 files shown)


src/main/java/org/opensearch/flowframework/workflow/HttpHostStep.java
```diff
@@ -0,0 +1,103 @@
...
+                currentNodeInputs,
+                outputs,
+                previousNodeInputs
+            );
+
+            String scheme = validScheme(inputs.get(SCHEME_FIELD));
+            String hostname = inputs.get(HOSTNAME_FIELD).toString();
+            int port = validPort(inputs.get(PORT_FIELD));
+
+            HttpHost httpHost = new HttpHost(scheme, hostname, port);
```
owaiskazi19, line 61: Instead of HttpHost workflow step, what do you think about creating RestHightLevelClient workflow step? Anyway, we need to pass this step to `RestClientBuilder.builder()`. So why not do it in the same workflow step?

src/test/java/org/opensearch/flowframework/model/WorkflowValidatorTests.java
```diff
@@ -37,67 +39,12 @@ public void setUp() throws Exception {
...
-        workflowStepValidators.put(
-            WorkflowStepFactory.WorkflowSteps.DELETE_AGENT.getWorkflowStepName(),
-            WorkflowStepFactory.WorkflowSteps.DELETE_AGENT.getWorkflowStepValidator()
-        );
-        workflowStepValidators.put(
-            WorkflowStepFactory.WorkflowSteps.NOOP.getWorkflowStepName(),
-            WorkflowStepFactory.WorkflowSteps.NOOP.getWorkflowStepValidator()
-        );
+        Map<String, WorkflowStepValidator> workflowStepValidators = Arrays.stream(WorkflowSteps.values())
+            .collect(Collectors.toMap(WorkflowSteps::getWorkflowStepName, WorkflowSteps::getWorkflowStepValidator));
```
owaiskazi19, line 43: :+1:

src/main/java/org/opensearch/flowframework/workflow/HttpHostStep.java
dbwiddis: > Instead of HttpHost workflow step, what do you think about creating RestHightLevelClient workflow step?

I have gone back and forth many times between these options, and written/deleted so much code that I think I can speak authoritatively here, and settled on separating it out this way.  Some reasons:
 - The client is a resource that needs to be closed, so is actually created in the same step as the REST API call.  Having a create client step introduces the complexity of closing it later.
 - Calling an external client requires this step; calling the cluster itself actually doesn't need an external configuration as it can be determined from the nodes themselves (which is how our integ tests do it).

i could wait and submit this all as part of a larger integrated setup so you could see why I chose it that way but I'm trying to submit incrementally here to make review easier.

owaiskazi19: The above 3 reason makes sense. To understand the design from what you are thinking `HttpHost` would map to `ToolStep` and `RestClient` -> `AgentStep`, right?

What type of Clients would we have? Would we also have external clients wokflow step? If we will just have one client then having HttpHost in the ClientStep would be a better way to go. In short, all the depends on the number of client steps we will have.

The other concern I think, with the `RestAPIStep` and communicating plugins on the REST layer would introduce the same security issues we face for SDK. WDYT?

dbwiddis: > The above 3 reason makes sense. To understand the design from what you are thinking `HttpHost` would map to `ToolStep` and `RestClient` -> `AgentStep`, right?

Not quite.  
 - Yes, `HttpHostStep` is similar to `ToolStep` (in fact that's the class I copied and edited) in that it's a simple "configuration to create an object" class.  
   - It may be optional for http client calling the cluster.
 - There will be no "step" for the client. The actual client used is an implementation detail.
 - There will be a `RestApiStep` containing the details of an `HttpRequest` (method, endpoint, params, entity (body), and headers (which can include credentials).
   - I may later create a credentials step to separate those out, or maybe add them to the HttpHost step later.

> What type of Clients would we have? Would we also have external clients wokflow step? 

This is an implementation detail that the user doesn't need to know about. Options include:
   - HighLevelRestClient is the easiest option but is on path to deprecation
   - Apache has a RestClient
   - The OpenSearch Java Client is an option
   - Java has a built in HttpClient

It will be part of the RestAPI step as a closeable resource in a try-with-resources block.  Code will look something like this:
```java
    try (RestClient client = RestClient
            .builder(httpHost)
            .setHttpClientConfigCallback(httpClientBuilder -> {
                return httpClientBuilder
                        .setSSLContext(sslcontext)
                        .setDefaultCredentialsProvider(credentialsProvider);
            })
            // maybe other things here
            .build()) {
        // substitute the RestApiStep parameters in the Request here
        final Response response = client.performRequest(new Request("GET", "/_cluster/health"));
        // ... etc.
    }
```

> If we will just have one client then having HttpHost in the ClientStep would be a better way to go. In short, all the depends on the number of client steps we will have.

There will just be one RestApiStep with client entirely internal.  My thought was to reduce the duplication of those 3 elements.   Note that we could replace that host with the equivalent string: `https://hostname:port` and have that as the single line in the RestApiStep rather than a previous node input.  But adding credentials in the headers and picking the exact correct header name and doing all the appropriate obfuscating is something the user shouldn't have to do.

> The other concern I think, with the `RestAPIStep` and communicating plugins on the REST layer would introduce the same security issues we face for SDK. WDYT?

That's the way we eventually have to go for serverless and is an option to the existing Transport clients.  I don't think it's a concern at all for external connectors.  It's essentially already happening with some of our current steps:
 - Create connector step is similar to the http host step.... pointing to a particular API
 - Calling the predict API on a deployed model is similar to the rest api step, calling that external resource

There are definitely concerns calling the cluster over rest layer that we will need to be careful about.

owaiskazi19: Just thinking aloud here. If the only inputs we are considering are the following three fields, what are your thoughts on using these fields as inputs for the `RestApiStep`? While having an `HttpHostStep` does aid in segregating logic, I don't have a strong opinion about keeping it as a separate step.
```
SCHEME_FIELD, HOSTNAME_FIELD, PORT_FIELD
```

owaiskazi19: I see the following reasons you provided for having HttpHost as a separate step. I still think, it's fine to have this step under the same RestApiStep to reduce the workflow steps and the template size.

> The client is a resource that needs to be closed, so is actually created in the same step as the REST API call. Having a create client step introduces the complexity of closing it later.

We can have a special case of closing the client after the provisioning is done if the template includes RestApiStep.

> Calling an external client requires this step; calling the cluster itself actually doesn't need an external configuration as it can be determined from the nodes themselves (which is how our integ tests do it).

We can take the external client connection inputs as parameters to RestApiStep.

> Calling multiple APIs on an external client would have less repetition of these parameters than always including them in the rest API call itself.

Even if we have a single `HttpHostStep`, we might still have multiple REST API calls.

Required keys would include other credentials as well for the Rest call.
```java
Set<String> requiredKeys = Set.of(SCHEME_FIELD, HOSTNAME_FIELD, PORT_FIELD);
HttpHost httpHost = new HttpHost(scheme, hostname, port);
try (RestClient client = RestClient
            .builder(httpHost)
            .setHttpClientConfigCallback(httpClientBuilder -> {
                return httpClientBuilder
                        .setSSLContext(sslcontext)
                        .setDefaultCredentialsProvider(credentialsProvider);
            })
            // maybe other things here
            .build()) {
        // substitute the RestApiStep parameters in the Request here
        final Response response = client.performRequest(new Request("GET", "/_cluster/health"));
        // ... etc.
    }
```

dbwiddis: > Let me know WDYT and I will approve the PR based on that.

My intent with this PR was incremental changes to avoid a massive PR later.  I may change this later, and even consider deleting it if we find we don't need it.

As-is, the scheme/hostname/port can be replaced with a single string and parsed, so provide little value.  It's the future inclusion of credentials that I am preparing for.

My suggestion:
1. approve as-is knowing it will change later
2. close this PR and wait for the complete RestAPI implementation later, where its usage is a bit more obvious and can be more carefully evaluated.

I don't have a strong opinion either way, I'll keep this commit and iterate on it for the future RestAPI step

owaiskazi19 (approved):

src/main/java/org/opensearch/flowframework/workflow/HttpHostStep.java
```diff
@@ -0,0 +1,103 @@
...
+
+            logger.info("Http Host registered successfully {}", httpHost);
+
+        } catch (FlowFrameworkException e) {
+            hostFuture.onFailure(e);
+        }
+        return hostFuture;
+    }
+
+    private String validScheme(Object o) {
```
amitgalitz, line 79: We might need to add more validation here based on the cluster settings potentially? It depends how we want the behavior to be but for example if `plugins.security.ssl.http.enabled: true` OS only allows https, we might want to also follow that here by only allowing https. Could be other cases on FGAC where this would be better too.

src/main/java/org/opensearch/flowframework/workflow/HttpHostStep.java
```diff
@@ -0,0 +1,103 @@
...
+            Map<String, Object> inputs = ParseUtils.getInputsFromPreviousSteps(
+                requiredKeys,
+                optionalKeys,
+                currentNodeInputs,
+                outputs,
+                previousNodeInputs
+            );
+
+            String scheme = validScheme(inputs.get(SCHEME_FIELD));
+            String hostname = inputs.get(HOSTNAME_FIELD).toString();
```
amitgalitz, line 58: This might be needed later on but we potentially should have some place to register trusted hostnames, so users can't just automatically be able to put any host, same can go for port. We can do validation against this trust registry then?

src/main/java/org/opensearch/flowframework/workflow/HttpHostStep.java
amitgalitz: >Calling an external client requires this step; calling the cluster itself actually doesn't need an external configuration as it can be determined from the nodes themselves (which is how our integ tests do it).

What do you mean by not needing a client to call the cluster itself (another plugin)?

>As-is, the scheme/hostname/port can be replaced with a single string and parsed, so provide little value. It's the future inclusion of credentials that I am preparing for.

if you are including the header in the `RestAPIStep` potentially, would you have the `httphost` stay mostly constant for all other calls inside the cluster? so we don't have to repeat this in every step basically?

src/main/java/org/opensearch/flowframework/workflow/HttpHostStep.java
dbwiddis: For calls to OpenSearch itself, the user won't be providing the http host, so this whole step won't apply.  This is only for external APIs.

src/main/java/org/opensearch/flowframework/workflow/HttpHostStep.java
dbwiddis: I'll probably be adding credentials in a follow-on PR, can I address this concern then?

src/main/java/org/opensearch/flowframework/workflow/HttpHostStep.java
```diff
@@ -0,0 +1,103 @@
...
+
+    @Override
+    public PlainActionFuture<WorkflowData> execute(
+        String currentNodeId,
+        WorkflowData currentNodeInputs,
+        Map<String, WorkflowData> outputs,
+        Map<String, String> previousNodeInputs
+    ) {
+        Set<String> requiredKeys = Set.of(SCHEME_FIELD, HOSTNAME_FIELD, PORT_FIELD);
+        Set<String> optionalKeys = Collections.emptySet();
```
amitgalitz, line 46: is this just a placeholder for now? I see some other steps might not have this.

src/main/java/org/opensearch/flowframework/workflow/HttpHostStep.java
amitgalitz: Yeah it can be addressed later, by credentials do you mean headers? I assume we will need something here for host, port and endpoint. We will also need to potentially register internal endpoints too

src/main/java/org/opensearch/flowframework/workflow/HttpHostStep.java
amitgalitz: How will we do calls for internal endpoints (core and plugin APIs)

src/main/java/org/opensearch/flowframework/workflow/HttpHostStep.java
dbwiddis: Yes, either basic (username/password) or api key or other params.  See various connector APIs for examples.

src/main/java/org/opensearch/flowframework/workflow/HttpHostStep.java
dbwiddis: If credentials are added later, they may be optional.  For now this is just a simple step encoding the Apache HttpHost object.  It takes no other parameters.

amitgalitz: got it, I am still want to see the full implementation, I was thinking first we handle the steps for the OS endpoints, not sure 100% sure why not to combine all the info together in one place like connector, I am okay merging this for now though

src/main/java/org/opensearch/flowframework/workflow/HttpHostStep.java
amitgalitz: can you add a TODO for adding mechanism for validating that a host/endpoint/port has been approved

src/main/java/org/opensearch/flowframework/workflow/HttpHostStep.java
dbwiddis: > How will we do calls for internal endpoints (core and plugin APIs)

We'll fetch the target host(s) from the nodes in the cluster.  We'll pass along the same authentication header that the user sent with their original request.  If there are special considerations (Sigv4, etc.) we'll handle those similarly to how they're being handled with the ML Commons use cases.

src/main/java/org/opensearch/flowframework/workflow/HttpHostStep.java
dbwiddis: I'm still researching the "full implementation" and trying to get known pieces of it done incrementally.   

The "full implementation" will be very similar to how ML Commons does it.

For reference, this workflow step essentially abstracts the HttpHost elements and validation in this class for example: https://github.com/opensearch-project/ml-commons/blob/7b609891a6b0d4df2e1e80959bebdf6f3deb9cdd/ml-algorithms/src/main/java/org/opensearch/ml/engine/httpclient/MLHttpClientFactory.java

src/main/java/org/opensearch/flowframework/workflow/HttpHostStep.java
amitgalitz: That sounds good to me, I'm okay keeping with this, I think my initial confusion came from the fact that I thought we were first gonna focus on opensearch APIs (considered everything not in plugin as external) and later on external APIs (outside of opensearch).

src/main/java/org/opensearch/flowframework/workflow/HttpHostStep.java
amitgalitz: That sounds good to me, I overall don't see to much of an issue with this, I think though with different examples of how the ratio of httphost steps we will have and the underlying apireststeps it will make more sense. I also as I stated above thought first focus was just the other plugin APIs so less of a need for this but I understand the full vision better

Issue #456, opened 2024-01-25
[FEATURE] Allow an optional workflow_step param to the get workflow steps API.


dbwiddis:
### Is your feature request related to a problem?

Users may want to just get the required inputs for a single workflow step rather than the entire list.

### What solution would you like?

Add an optional parameter to the get workflow steps API listing one or more comma-delimeted workflow steps, and return a subset  of values for just those.

### Do you have any additional context?

This could probably be done at the same time as #395

PR #538, opened 2024-02-23
Added an optional workflow_step param to the get workflow steps API

owaiskazi19:
### Description

Added an optional workflow_step param to the get workflow steps API.

Request:
```
GET localhost:9200/_plugins/_flow_framework/workflow/_steps?workflow_step=create_connector,delete_model,deploy_model
```
Response:
```
{
    "create_connector": {
        "inputs": [
            "name",
            "description",
            "version",
            "protocol",
            "parameters",
            "credential",
            "actions"
        ],
        "outputs": [
            "connector_id"
        ],
        "required_plugins": [
            "opensearch-ml"
        ],
        "timeout": "1m"
    },
    "deploy_model": {
        "inputs": [
            "model_id"
        ],
        "outputs": [
            "model_id"
        ],
        "required_plugins": [
            "opensearch-ml"
        ],
        "timeout": "15s"
    },
    "delete_model": {
        "inputs": [
            "model_id"
        ],
        "outputs": [
            "model_id"
        ],
        "required_plugins": [
            "opensearch-ml"
        ]
    }
}
```

### Issues Resolved
Resolves https://github.com/opensearch-project/flow-framework/issues/456

### DIFF
--- a/src/main/java/org/opensearch/flowframework/common/CommonValue.java
+++ b/src/main/java/org/opensearch/flowframework/common/CommonValue.java
@@ -65,7 +65,7 @@ public class CommonValue {
     /** The field name for provision workflow within a use case template*/
     public static final String PROVISION_WORKFLOW = "provision";
     /** The field name for workflow steps. This field represents the name of the workflow steps to be fetched. */
-    public static final String STEP = "step";
+    public static final String WORKFLOW_STEP = "workflow_step";
 
     /*
      * Constants associated with plugin configuration
--- a/src/main/java/org/opensearch/flowframework/common/CommonValue.java
+++ b/src/main/java/org/opensearch/flowframework/common/CommonValue.java
@@ -65,7 +65,7 @@ public class CommonValue {
     /** The field name for provision workflow within a use case template*/
     public static final String PROVISION_WORKFLOW = "provision";
     /** The field name for workflow steps. This field represents the name of the workflow steps to be fetched. */
-    public static final String STEPS = "steps";
+    public static final String STEP = "step";
 
     /*
      * Constants associated with plugin configuration
--- a/src/main/java/org/opensearch/flowframework/common/CommonValue.java
+++ b/src/main/java/org/opensearch/flowframework/common/CommonValue.java
@@ -64,6 +64,8 @@ public class CommonValue {
     public static final String VALIDATION = "validation";
     /** The field name for provision workflow within a use case template*/
     public static final String PROVISION_WORKFLOW = "provision";
+    /** The field name for workflow steps. This field represents the name of the workflow steps to be fetched. */
+    public static final String STEPS = "steps";
 
     /*
      * Constants associated with plugin configuration
--- a/src/main/java/org/opensearch/flowframework/rest/RestGetWorkflowStepAction.java
+++ b/src/main/java/org/opensearch/flowframework/rest/RestGetWorkflowStepAction.java
@@ -31,7 +31,7 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 
-import static org.opensearch.flowframework.common.CommonValue.STEPS;
+import static org.opensearch.flowframework.common.CommonValue.STEP;
 import static org.opensearch.flowframework.common.CommonValue.WORKFLOW_URI;
 import static org.opensearch.flowframework.common.FlowFrameworkSettings.FLOW_FRAMEWORK_ENABLED;
 
@@ -65,7 +65,7 @@ public class RestGetWorkflowStepAction extends BaseRestHandler {
     @Override
     protected RestChannelConsumer prepareRequest(RestRequest request, NodeClient client) throws IOException {
 
-        String[] steps = request.paramAsStringArray(STEPS, Strings.EMPTY_ARRAY);
+        String[] steps = request.paramAsStringArray(STEP, Strings.EMPTY_ARRAY);
         try {
             if (!flowFrameworkSettings.isFlowFrameworkEnabled()) {
                 throw new FlowFrameworkException(
@@ -76,7 +76,7 @@ public class RestGetWorkflowStepAction extends BaseRestHandler {
 
             Map<String, String> params = new HashMap<>();
             for (String step : steps) {
-                params.put(step, STEPS);
+                params.put(step, STEP);
             }
 
             WorkflowRequest workflowRequest = new WorkflowRequest(null, null, params);
--- a/src/main/java/org/opensearch/flowframework/rest/RestGetWorkflowStepAction.java
+++ b/src/main/java/org/opensearch/flowframework/rest/RestGetWorkflowStepAction.java
@@ -31,7 +31,8 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 
-import static org.opensearch.flowframework.common.CommonValue.*;
+import static org.opensearch.flowframework.common.CommonValue.STEPS;
+import static org.opensearch.flowframework.common.CommonValue.WORKFLOW_URI;
 import static org.opensearch.flowframework.common.FlowFrameworkSettings.FLOW_FRAMEWORK_ENABLED;
 
 /**
@@ -75,7 +76,7 @@ public class RestGetWorkflowStepAction extends BaseRestHandler {
 
             Map<String, String> params = new HashMap<>();
             for (String step : steps) {
-                params.put(STEPS, step);
+                params.put(step, STEPS);
             }
 
             WorkflowRequest workflowRequest = new WorkflowRequest(null, null, params);
--- a/src/main/java/org/opensearch/flowframework/rest/RestGetWorkflowStepAction.java
+++ b/src/main/java/org/opensearch/flowframework/rest/RestGetWorkflowStepAction.java
@@ -11,25 +11,27 @@ package org.opensearch.flowframework.rest;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.opensearch.ExceptionsHelper;
-import org.opensearch.action.ActionRequest;
-import org.opensearch.action.ActionRequestValidationException;
 import org.opensearch.client.node.NodeClient;
 import org.opensearch.core.action.ActionListener;
+import org.opensearch.core.common.Strings;
 import org.opensearch.core.rest.RestStatus;
 import org.opensearch.core.xcontent.ToXContent;
 import org.opensearch.core.xcontent.XContentBuilder;
 import org.opensearch.flowframework.common.FlowFrameworkSettings;
 import org.opensearch.flowframework.exception.FlowFrameworkException;
 import org.opensearch.flowframework.transport.GetWorkflowStepAction;
+import org.opensearch.flowframework.transport.WorkflowRequest;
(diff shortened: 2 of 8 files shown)


src/main/java/org/opensearch/flowframework/rest/RestGetWorkflowStepAction.java
```diff
@@ -69,13 +74,13 @@ protected RestChannelConsumer prepareRequest(RestRequest restRequest, NodeClient
...
-            ActionRequest request = new ActionRequest() {
-                @Override
-                public ActionRequestValidationException validate() {
-                    return null;
-                }
-            };
-            return channel -> client.execute(GetWorkflowStepAction.INSTANCE, request, ActionListener.wrap(response -> {
+            Map<String, String> params = new HashMap<>();
+            for (String step : steps) {
+                params.put(step, WORKFLOW_STEP);
```
dbwiddis, line 79: While this works, it seems unusual to do the parsing on this side of transport, incurring the transport overhead of the map size and multiple repetitions of the "workflow_step" string (with its length).

Why not just fetch the param as-is and pass the comma-delimited string as a single parameter?  Something like
```java
Map<String, String> params = request.hasParam(WORKFLOW_STEP) 
    ? Map.of(WORKFLOW_STEP, request.param(WORKFLOW_STEP)) 
    : Collections.emptyMap();
```
Then later when you want the list:
```diff
- List<String> steps = new ArrayList<>(request.getParams().keySet());
+ List<String> steps = request.containsKey(WORKFLOW_STEP)
+     ? Arrays.asList(Strings.splitStringByCommaToArray(request.getParams().get(WORKFLOW_STEP)))
+     : Collections.emptyList();
```

src/main/java/org/opensearch/flowframework/transport/GetWorkflowStepTransportAction.java
```diff
@@ -47,11 +49,21 @@ public GetWorkflowStepTransportAction(
...
+            WorkflowValidator workflowValidator;
+            if (steps.isEmpty()) {
+                workflowValidator = this.workflowStepFactory.getWorkflowValidator();
+            } else {
+                workflowValidator = this.workflowStepFactory.getWorkflowValidatorByStep(steps);
+            }
             listener.onResponse(new GetWorkflowStepResponse(workflowValidator));
         } catch (Exception e) {
+            if (e instanceof FlowFrameworkException) {
+                logger.error("Please only use only valid step name");
```
dbwiddis, line 64: Users won't see this log message. Probably better to just say "invalid step name" here.  Or just log it inside the validator when you throw the exception, or use the exception message here.

src/main/java/org/opensearch/flowframework/workflow/WorkflowStepFactory.java
```diff
@@ -350,6 +350,27 @@ public WorkflowValidator getWorkflowValidator() {
...
+
+        for (WorkflowSteps mapping : WorkflowSteps.values()) {
+            if (steps.contains(mapping.getWorkflowStepName())) {
+                workflowStepValidators.put(mapping.getWorkflowStepName(), mapping.getWorkflowStepValidator());
+            }
+        }
+
+        if (workflowStepValidators.isEmpty()) {
+            throw new FlowFrameworkException("Please only use only valid step name", RestStatus.BAD_REQUEST);
+        }
```
dbwiddis, line 369:
```suggestion
        Map<String, WorkflowStepValidator> workflowStepValidators = new HashMap<>();
        Set<String> invalidSteps = new HashSet<>(steps);

        for (WorkflowSteps mapping : WorkflowSteps.values()) {
            String step = mapping.getWorkflowStepName();
            if (steps.contains(step)) {
                workflowStepValidators.put(step, mapping.getWorkflowStepValidator());
                invalidSteps.remove(step);
            }
        }

        if (!invalidSteps.isEmpty()) {
            throw new FlowFrameworkException("Invalid step name: " + invalidSteps, RestStatus.BAD_REQUEST);
        }
```

dbwiddis (commented):
Looks good. A few suggestions to make it better.

src/main/java/org/opensearch/flowframework/rest/RestGetWorkflowStepAction.java
owaiskazi19: Better idea!

src/main/java/org/opensearch/flowframework/transport/GetWorkflowStepTransportAction.java
```diff
@@ -47,11 +53,23 @@ public GetWorkflowStepTransportAction(
...
         try {
-            WorkflowValidator workflowValidator = this.workflowStepFactory.getWorkflowValidator();
+            List<String> steps = request.getParams().size() > 0
+                ? Arrays.asList(Strings.splitStringByCommaToArray(request.getParams().get(WORKFLOW_STEP)))
+                : Collections.emptyList();
+            WorkflowValidator workflowValidator;
+            if (steps.isEmpty()) {
+                workflowValidator = this.workflowStepFactory.getWorkflowValidator();
+            } else {
+                workflowValidator = this.workflowStepFactory.getWorkflowValidatorByStep(steps);
```
dbwiddis, line 65: Can we add a test to cover this code branch?

dbwiddis (approved):

src/main/java/org/opensearch/flowframework/common/CommonValue.java
```diff
@@ -64,6 +64,8 @@ private CommonValue() {}
     public static final String VALIDATION = "validation";
     /** The field name for provision workflow within a use case template*/
     public static final String PROVISION_WORKFLOW = "provision";
+    /** The field name for workflow steps. This field represents the name of the workflow steps to be fetched. */
+    public static final String WORKFLOW_STEP = "workflow_step";
```
amitgalitz, line 68: nit: thoughts on having this `type` or `name`, not a strong suggestion at all, just thinking cause the param is same as the path kinda `_flow_framework/workflow/_steps`

src/main/java/org/opensearch/flowframework/rest/RestGetWorkflowStepAction.java
```diff
@@ -69,13 +71,12 @@ protected RestChannelConsumer prepareRequest(RestRequest restRequest, NodeClient
...
-                public ActionRequestValidationException validate() {
-                    return null;
-                }
-            };
-            return channel -> client.execute(GetWorkflowStepAction.INSTANCE, request, ActionListener.wrap(response -> {
+            Map<String, String> params = request.hasParam(WORKFLOW_STEP)
+                ? Map.of(WORKFLOW_STEP, request.param(WORKFLOW_STEP))
+                : Collections.emptyMap();
+
+            WorkflowRequest workflowRequest = new WorkflowRequest(null, null, params);
```
amitgalitz, line 78: why not create a WorkflowStepRequest class? Not a must since we already have a few different APIs using this, just seeming like its becoming a catch all for many APIs

src/main/java/org/opensearch/flowframework/transport/GetWorkflowStepTransportAction.java
```diff
@@ -47,11 +53,23 @@ public GetWorkflowStepTransportAction(
...
+            }
             listener.onResponse(new GetWorkflowStepResponse(workflowValidator));
         } catch (Exception e) {
+            if (e instanceof FlowFrameworkException) {
+                logger.error("Invalid step name");
```
amitgalitz, line 70: can you also log the exception message here so we record which names were invalid on the request

owaiskazi19: Since we were removing it anyway, that's why I didn't log it: https://github.com/opensearch-project/flow-framework/pull/524

src/main/java/org/opensearch/flowframework/rest/RestGetWorkflowStepAction.java
owaiskazi19: I didn't have to change any code in the `WorkflowRequest` class to use it here. Instead of adding another request class when the same thing can be achieved using the existing one. If we have additional requirements beyond what's already in the current `WorkflowRequest` class, then it definitely makes sense to introduce another request class.

src/main/java/org/opensearch/flowframework/common/CommonValue.java
owaiskazi19: No strong opinion here. @dbwiddis, @joshpalis  thoughts?

src/main/java/org/opensearch/flowframework/rest/RestGetWorkflowStepAction.java
amitgalitz: I am okay with this for now, I personally prefer having a Request class for each API (provision, get, create) instead of the multiple constructors but its already there, agree if we make a change that requires another constructor we can potentially refactor it then.

src/main/java/org/opensearch/flowframework/common/CommonValue.java
dbwiddis: Looking at the docs we kind of make them synonymous.   "Step" seems like an "instance" whereas "Type" seems like a "class".

I don't think we should change this constant, the underlying question is what to put from an API perspective.

Technically we might want to use "type" or "types" as the filtering param, but since the overall list is called "_steps" then an individual step as the param may be better.  Or maybe "step_types"? 

No strong opinion.

src/main/java/org/opensearch/flowframework/transport/GetWorkflowStepTransportAction.java
amitgalitz: We will be reversing changes in #524 to include helpful exception messages and stack traces as well when helpful for debugging purposes

owaiskazi19: Added the exception message.

src/main/java/org/opensearch/flowframework/common/CommonValue.java
owaiskazi19: @amitgalitz let's get the PR in if no concern here. If there are more inputs on the ask, I'll raise a separate PR. Thanks

joshpalis (approved):

src/main/java/org/opensearch/flowframework/transport/GetWorkflowStepTransportAction.java
```diff
@@ -47,11 +53,23 @@ public GetWorkflowStepTransportAction(
...
+            WorkflowValidator workflowValidator;
+            if (steps.isEmpty()) {
+                workflowValidator = this.workflowStepFactory.getWorkflowValidator();
+            } else {
+                workflowValidator = this.workflowStepFactory.getWorkflowValidatorByStep(steps);
+            }
             listener.onResponse(new GetWorkflowStepResponse(workflowValidator));
         } catch (Exception e) {
+            if (e instanceof FlowFrameworkException) {
+                logger.error("Invalid step name", e.getMessage());
```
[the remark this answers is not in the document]
owaiskazi19: Shouldn't be a nit. That's a good point. Pushed the change

Issue #496, opened 2024-02-06
[FEATURE] Pre-defined templates and defaults 


amitgalitz:
### Is your feature request related to a problem?

Flow-Framework currently supports the creation and provisioning of templates in order to executes a sequence of API calls in OpenSearch. Pasted in the bottom of this issue is an example of a more complicated template where we create a connector, set up a model, a few tools and agents. 

As seen this template can be quite long to send to an API and users might not want or need to constantly be altering all the variables in it. 

It is popular in CDK repos or other configuration automation platforms to have a simpler mechanisms to easily change just the variables that we want. Another potential improvement that could be added together with this is having predefined templates that are easily used and also ‘fine-grained’ edited by customers.

### What solution would you like?

Users should be able to make an API request to create a template by choosing a predefined template and editing only the variables they care about.

```
http://_plugins/_flow_framework/workflow?use_case=openAI-deploy-model-1
{
  "create_connector_1.credential.openAIkey": "asdfg-1234",
  "create_connector_1.model": "chatgpt-4.0",
  "register_remote_model_1.name": "chatgpt-1"
}
```
### How we will achieve this?

In order to be able to easily substitute any subset of field in the template that we want we will have to maintain two different set of documents. Giving these temporary names for easier discussion: 

1. **Substitution Ready Template** A set of templates that have all there fields ready to be substituted 
2. **Predefined defaults** A set of files with predefined defaults for each of the templates (potentially some default files can be used for multiple templates

#### Example:

**Substitution ready default template:**

* Lots of things aren’t set in stone (set here the ID to be create_connector_1 and also what we refer to in our API call below incase a template has multiple steps that are of the same type)

```
{
  "name": "{template.name}",
  "description": "{template.description}",
  "use_case": "MODEL_DEPLOYMENT_API_KEY",
  "version": {
    "template": "1.0.0",
    "compatibility": [
      "2.12.0",
      "3.0.0"
    ]
  },
  "workflows": {
    "provision": {
      "nodes": [
        {
          "id": "create_connector_1",
          "type": "create_connector",
          "user_inputs": {
            "name": "${create_connector_1}",
            "description": "${create_connector_1.description}",
            "version": "1",
            "protocol": "${create_connector_1.protocol}",
            "parameters": {
              "endpoint": "${create_connector_1.endpoint}",
              "model": "${create_connector_1.model}"
            },
            "credential": {
              "key": "${create_connector_1.credential.key}"
            },
            "actions": [
              {
[...]

PR #551, opened 2024-03-05
Add created, updated, and provisioned timestamps to saved template

### LINKED ISSUE #548: [FEATURE] Support timestamp fields in workflows
A common field for plugin resources include timestamps indicating when a resource is created / updated / deployed. This is also a requirement based on the latest @opensearch-project/opensearch-ux mocks for the frontend (refer to mock in https://github.com/opensearch-project/dashboards-flow-framework/issues/65). I propose the following added fields:

- `last_update_time` in the workflows system index, and updating using the current timestamp when a POST/PUT request is made.
- `last_launched_time` (no strong opinion on the name) in the workflows state index indicating when a provision/launch was last executed. This could be updated the same time the provisioning progress field is updated, for example.

dbwiddis:
### Description

Adds created time, last updated time, and last provisioned time to the template stored in the global context.
 - Updates the last updated time when updating a workflow
 - Updates the last provisioned time when provisoning a workflow

Also adds BWC tests!

Also fixes Integ tests which were essentially running twice since yamlRestTest was not properly filtered and was running them a second time.

### Issues Resolved

Fixes #548 

### DIFF
--- a/.github/workflows/test_bwc.yml
+++ b/.github/workflows/test_bwc.yml
@@ -1,0 +1,45 @@
+name: BWC
+on:
+  push:
+    branches:
+      - "**"
+  pull_request:
+    branches:
+      - "**"
+
+jobs:
+  Build-ff-linux:
+    strategy:
+      matrix:
+        java: [11,17,21]
+      fail-fast: false
+
+    name: Test Flow Framework BWC
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Setup Java ${{ matrix.java }}
+        uses: actions/setup-java@v4
+        with:
+          distribution: 'temurin'
+          java-version: ${{ matrix.java }}
+
+      - name: Checkout Flow Framework
+        uses: actions/checkout@v4
+
+      - name: Assemble Flow Framework
+        run: |
+          plugin_version=`./gradlew properties -q | grep "opensearch_build:" | awk '{print $2}'`
+          echo plugin_version $plugin_version
+          ./gradlew assemble
+          echo "Creating ./src/test/resources/org/opensearch/flowframework/bwc/flow-framework/$plugin_version ..."
+          mkdir -p ./src/test/resources/org/opensearch/flowframework/bwc/flow-framework/$plugin_version
+          echo "Copying ./build/distributions/*.zip to ./src/test/resources/org/opensearch/flowframework/bwc/flow-framework/$plugin_version ..."
+          ls ./build/distributions/
+          cp ./build/distributions/*.zip ./src/test/resources/org/opensearch/flowframework/bwc/flow-framework/$plugin_version
+          echo "Copied ./build/distributions/*.zip to ./src/test/resources/org/opensearch/flowframework/bwc/flow-framework/$plugin_version ..."
+          ls ./src/test/resources/org/opensearch/flowframework/bwc/flow-framework/$plugin_version
+      - name: Run Flow Framework Backwards Compatibility Tests
+        run: |
+          echo "Running backwards compatibility tests ..."
+          ./gradlew bwcTestSuite -Dtests.security.manager=false
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,8 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 - Added an optional workflow_step param to the get workflow steps API ([#538](https://github.com/opensearch-project/flow-framework/pull/538))
 
 ### Enhancements
+- Add created, updated, and provisioned timestamps to saved template ([#551](https://github.com/opensearch-project/flow-framework/pull/551))
+
 ### Bug Fixes
 ### Infrastructure
 ### Documentation
--- a/build.gradle
+++ b/build.gradle
@@ -1,5 +1,6 @@
 import java.nio.file.Files
 import org.opensearch.gradle.testclusters.OpenSearchCluster
+import org.opensearch.gradle.testclusters.StandaloneRestIntegTestTask
 import org.opensearch.gradle.test.RestIntegTestTask
 import java.util.concurrent.Callable
 import java.nio.file.Paths
@@ -23,6 +24,16 @@ buildscript {
         opensearch_no_snapshot = opensearch_build.replace("-SNAPSHOT","")
         System.setProperty('tests.security.manager', 'false')
         common_utils_version = System.getProperty("common_utils.version", opensearch_build)
+
+        bwcVersionShort = "2.12.0"
+        bwcVersion = bwcVersionShort + ".0"
+        bwcOpenSearchFFDownload = 'https://ci.opensearch.org/ci/dbc/distribution-build-opensearch/' + bwcVersionShort + '/latest/linux/x64/tar/builds/' +
+                'opensearch/plugins/opensearch-flow-framework-' + bwcVersion + '.zip'
+        baseName = "ffBwcCluster"
+        bwcFilePath = "src/test/resources/org/opensearch/flowframework/bwc/"
+        bwcFlowFrameworkPath = bwcFilePath + "flowframework/"
+
+        isSameMajorVersion = opensearch_version.split("\\.")[0] == bwcVersionShort.split("\\.")[0]
     }
 
     repositories {
@@ -78,6 +89,9 @@ dependencyLicenses.enabled = false
 // This requires an additional Jar not published as part of build-tools
 loggerUsageCheck.enabled = false
 thirdPartyAudit.enabled = false
+// Allow test cases to be named Tests without having to be inherited from LuceneTestCase.
+// see https://github.com/elastic/elasticsearch/blob/323f312bbc829a63056a79ebe45adced5099f6e6/buildSrc/src/main/java/org/elasticsearch/gradle/precommit/TestingConventionsTasks.java
+testingConventions.enabled = false
 
 // No need to validate pom, as we do not upload to maven/sonatype
 validateNebulaPom.enabled = false
@@ -192,6 +206,12 @@ jacocoTestReport {
 }
 tasks.named("check").configure { dependsOn(jacocoTestReport) }
 
+tasks.named("yamlRestTest").configure {
+    filter {
+        excludeTestsMatching "org.opensearch.flowframework.rest.*IT"
+        excludeTestsMatching "org.opensearch.flowframework.bwc.*IT"
+    }
+}
 
 // Set up integration tests
 task integTest(type: RestIntegTestTask) {
@@ -228,6 +248,13 @@ integTest {
     if (System.getProperty("tests.rest.cluster") != null) {
         filter {
             includeTestsMatching "org.opensearch.flowframework.rest.*IT"
+        }
+    }
+
+    // Exclude BWC tests, run separately
+    if (System.getProperty("tests.rest.bwcsuite") == null) {
+        filter {
+            excludeTestsMatching "org.opensearch.flowframework.bwc.*IT"
         }
     }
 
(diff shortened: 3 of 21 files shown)


src/main/java/org/opensearch/flowframework/transport/ProvisionWorkflowTransportAction.java
```diff
@@ -145,7 +146,36 @@ protected void doExecute(Task task, WorkflowRequest request, ActionListener<Work
...
-                                listener.onResponse(new WorkflowResponse(workflowId));
+                                // update last provisioned field in template
+                                Template newTemplate = new Template.Builder().name(template.name())
+                                    .description(template.description())
+                                    .useCase(template.useCase())
+                                    .templateVersion(template.templateVersion())
+                                    .compatibilityVersion(template.compatibilityVersion())
+                                    .workflows(template.workflows())
+                                    .uiMetadata(template.getUiMetadata())
+                                    .user(template.getUser()) // Should we care about old user here?
```
joshpalis, line 157: I'd agree here, the `user` in templates should be the one who created it. Curious though if it would be worth to track which user provisioned the template

joshpalis (approved):
Lgtm, just a small comment

dbwiddis:
> pending fix for failing integration tests. 

Yeah, I chose to update the state index before updating the template in the global context which triggers this.  I can easily reverse the order (it made more sense that way) but I wanted to flip the flag to provisioning as soon as possible to avoid a race condition!  Will see if there's another workaround.

amitgalitz:
Start and end time should already be in the state index

dbwiddis:
> Start and end time should already be in the state index

Yes, they are.  

This should match (within milliseconds) the start time of the latest provisioning attempt.  Probably didn't need to add it here, but now that it's here should we leave it or take it back out?

dbwiddis:
Good news: I've added BWC tests and they work well for old cluster and restart.  

The BWC tests are failing on the rolling upgrade, but I have seen identical behavior and results on multiple other plugins (AD, knn).  The only plugin I found with successful rolling upgrade BWC was neural search but they are only testing 3.0.0 vs. 2.13.0 snapshot, not the 2.12.0 that's failing.

So whatever the failure is, is not in this plugin.

src/main/java/org/opensearch/flowframework/transport/CreateWorkflowTransportAction.java
```diff
@@ -223,40 +231,73 @@ protected void doExecute(Task task, WorkflowRequest request, ActionListener<Work
...
+                            .useCase(templateWithUser.useCase())
+                            .templateVersion(templateWithUser.templateVersion())
+                            .compatibilityVersion(templateWithUser.compatibilityVersion())
+                            .workflows(templateWithUser.workflows())
+                            .uiMetadata(templateWithUser.getUiMetadata())
+                            .user(templateWithUser.getUser()) // Should we care about old user here?
+                            .createdTime(existingTemplate.createdTime())
+                            .lastUpdatedTime(Instant.now())
+                            .lastProvisionedTime(existingTemplate.lastProvisionedTime())
+                            .build();
```
owaiskazi19, line 254: Can we just not do `Template.Builder().lastUpdatedTime(Instant.now()))` to only update one field?

src/main/java/org/opensearch/flowframework/indices/FlowFrameworkIndicesHandler.java
```diff
@@ -395,6 +395,17 @@ public void putInitialStateToWorkflowState(String workflowId, User user, ActionL
...
+    }
+
+    /**
+     * Replaces a document in the global context index
+     * @param documentId the document Id
+     * @param template the use-case template
+     * @param listener action listener
+     * @param force if set true, ignores the requirement that the provisioning is not started
+     */
+    public void updateTemplateInGlobalContext(String documentId, Template template, ActionListener<IndexResponse> listener, boolean force) {
```
owaiskazi19, line 408:
```suggestion
    public void updateTemplateInGlobalContext(String documentId, Template template, ActionListener<IndexResponse> listener, boolean isWorkflowStarted) {
```

src/test/java/org/opensearch/flowframework/bwc/FlowFrameworkBackwardsCompatibilityIT.java
```diff
@@ -0,0 +1,189 @@
...
+        // Enable Flow Framework Plugin Rest APIs
+        Response response = TestHelpers.makeRequest(
+            client(),
+            "PUT",
+            "_cluster/settings",
+            null,
+            "{\"transient\":{\"plugins.flow_framework.enabled\":true}}",
+            List.of(new BasicHeader(HttpHeaders.USER_AGENT, ""))
+        );
+        assertEquals(200, response.getStatusLine().getStatusCode());
```
owaiskazi19, line 55: Do we need to enable the settings anymore since we have set the flag as true by default?

owaiskazi19 (commented):
Test classes yet to be reviewed. Added few comments

CHANGELOG.md
```diff
@@ -21,6 +21,8 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 - Added an optional workflow_step param to the get workflow steps API ([#538](https://github.com/opensearch-project/flow-framework/pull/538))
 
 ### Enhancements
+- Add created, updated, and provisioned timestamps to saved template ([#551](https://github.com/opensearch-project/flow-framework/pull/551))
```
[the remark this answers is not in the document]
owaiskazi19: I see you have updated the changes. Ignore this comment.

src/test/java/org/opensearch/flowframework/bwc/FlowFrameworkBackwardsCompatibilityIT.java
dbwiddis: Yes, because it's not enabled on 2.12 and this BWC test runs on 2.12. :)

src/main/java/org/opensearch/flowframework/model/Template.java
```diff
@@ -48,14 +53,17 @@ public class Template implements ToXContentObject {
...
-    private String description;
-    private String useCase; // probably an ENUM actually
-    private Version templateVersion;
-    private List<Version> compatibilityVersion;
-    private Map<String, Workflow> workflows;
-    private Map<String, Object> uiMetadata;
-    private User user;
+    private final String name;
+    private final String description;
+    private final String useCase; // probably an ENUM actually
```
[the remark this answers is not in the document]
dbwiddis: I don't know. Can we make that a different issue discussion, possibly associated with #496?  It's not related to this PR where I just added final to what already existed.

src/main/java/org/opensearch/flowframework/transport/CreateWorkflowTransportAction.java
dbwiddis: `Template.Builder()` initializes with no defaults.

We could add a new constructor to do `Template.Builder(existingTemplate)`.  WDYT?

src/main/java/org/opensearch/flowframework/indices/FlowFrameworkIndicesHandler.java
dbwiddis: I'm ok with changing `force` to something else, but the point is to update it regardless of whether it's started or not.   Maybe `ignoreNotStartedCheck`?

build.gradle
```diff
@@ -425,6 +452,164 @@ task integTestRemote(type: RestIntegTestTask) {
...
+
+// A bwc test suite which runs all the bwc tasks combined.
+task bwcTestSuite(type: StandaloneRestIntegTestTask) {
+    exclude '**/*Test*'
+    exclude '**/*IT*'
+
+    dependsOn tasks.named("${baseName}#mixedClusterTask")
+    dependsOn tasks.named("${baseName}#rollingUpgradeClusterTask")
+    dependsOn tasks.named("${baseName}#fullRestartClusterTask")
+}
```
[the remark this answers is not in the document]
dbwiddis: I'll look into it, but I've spent countless hours getting this working and would prefer to refactor it later.

src/main/java/org/opensearch/flowframework/transport/CreateWorkflowTransportAction.java
dbwiddis: Created a new builder with template arg. Nice suggestion.

build.gradle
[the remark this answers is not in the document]
dbwiddis: https://github.com/opensearch-project/flow-framework/issues/555

src/test/java/org/opensearch/flowframework/bwc/FlowFrameworkBackwardsCompatibilityIT.java
```diff
@@ -0,0 +1,189 @@
...
+                default:
+                    throw new AssertionError("unknown cluster type: " + value);
+            }
+        }
+    }
+
+    @SuppressWarnings("unchecked")
+    public void testBackwardsCompatibility() throws Exception {
+        String uri = getUri();
+        Map<String, Map<String, Object>> responseMap = (Map<String, Map<String, Object>>) getAsMap(uri).get("nodes");
```
owaiskazi19, line 106: I am playing with ObjectMapper for one of my task and thought about the below code to avoid explicit casting
How about

```java
Map<String, Object> resultMap = getAsMap(uri);
ObjectMapper objectMapper = new ObjectMapper();

Map<String, Map<String, Object>> responseMap = objectMapper.convertValue(
    resultMap.get("nodes"),
    objectMapper.getTypeFactory().constructMapType(Map.class, String.class, Object.class)
);
```

src/test/java/org/opensearch/flowframework/bwc/FlowFrameworkBackwardsCompatibilityIT.java
```diff
@@ -0,0 +1,189 @@
...
+        for (Map<String, Object> response : responseMap.values()) {
+            List<Map<String, Object>> plugins = (List<Map<String, Object>>) response.get("plugins");
+            Set<Object> pluginNames = plugins.stream().map(map -> map.get("name")).collect(Collectors.toSet());
+            String workflowId = createNoopTemplate();
+            Template t = getTemplate(workflowId);
+            switch (CLUSTER_TYPE) {
+                case OLD:
+                    assertTrue(pluginNames.contains("opensearch-flow-framework"));
```
owaiskazi19, line 114: This is common for all the cases. Can be moved above switch?

src/test/java/org/opensearch/flowframework/bwc/FlowFrameworkBackwardsCompatibilityIT.java
```diff
@@ -0,0 +1,189 @@
...
+                return "_nodes/" + CLUSTER_NAME + "-0/plugins";
+            case MIXED:
+                String round = System.getProperty("tests.rest.bwcsuite_round");
+                if (round.equals("second")) {
+                    return "_nodes/" + CLUSTER_NAME + "-1/plugins";
+                } else if (round.equals("third")) {
+                    return "_nodes/" + CLUSTER_NAME + "-2/plugins";
+                } else {
+                    return "_nodes/" + CLUSTER_NAME + "-0/plugins";
+                }
```
owaiskazi19, line 150: How about?
```suggestion
      int nodeId = round.equals("second") ? 1 : round.equals("third") ? 2 : 0;
      return "_nodes/" + CLUSTER_NAME + "-" + nodeId + "/plugins";
```

owaiskazi19 (commented):
Overall looks good. Few comments

src/main/java/org/opensearch/flowframework/indices/FlowFrameworkIndicesHandler.java
owaiskazi19: `ignoreNotStartedCheck` sounds better than `force`.

src/test/java/org/opensearch/flowframework/bwc/FlowFrameworkBackwardsCompatibilityIT.java
dbwiddis: I'd rather keep the code the same as other implementations of BWC from which this code was copied.

ObjectMapper is still doing casting under the hood.

src/test/java/org/opensearch/flowframework/bwc/FlowFrameworkBackwardsCompatibilityIT.java
dbwiddis: It can actually be removed since it's the same for all cases.  I figure it didn't hurt to leave as a placeholder if we ever end up changing the plugin name like AD did.

src/test/java/org/opensearch/flowframework/bwc/FlowFrameworkBackwardsCompatibilityIT.java
dbwiddis: Another case of "this code was directly copied from other BWC implementations".  I'd rather leave working code alone than try to make it look different than others, making it harder to track/copy bug fixes from other implementations if they ever occur.

owaiskazi19: No harm in improving it. I'll leave that choice to you.

.github/workflows/test_bwc.yml
```diff
@@ -0,0 +1,45 @@
...
+
+      - name: Assemble Flow Framework
+        run: |
+          plugin_version=`./gradlew properties -q | grep "opensearch_build:" | awk '{print $2}'`
+          echo plugin_version $plugin_version
+          ./gradlew assemble
+          echo "Creating ./src/test/resources/org/opensearch/flowframework/bwc/flow-framework/$plugin_version ..."
+          mkdir -p ./src/test/resources/org/opensearch/flowframework/bwc/flow-framework/$plugin_version
+          echo "Copying ./build/distributions/*.zip to ./src/test/resources/org/opensearch/flowframework/bwc/flow-framework/$plugin_version ..."
+          ls ./build/distributions/
```
owaiskazi19, line 38: We can remove `ls`

.github/workflows/test_bwc.yml
```diff
@@ -0,0 +1,45 @@
...
+          cp ./build/distributions/*.zip ./src/test/resources/org/opensearch/flowframework/bwc/flow-framework/$plugin_version
+          echo "Copied ./build/distributions/*.zip to ./src/test/resources/org/opensearch/flowframework/bwc/flow-framework/$plugin_version ..."
+          ls ./src/test/resources/org/opensearch/flowframework/bwc/flow-framework/$plugin_version
```
owaiskazi19, line 41: And from here as well

.github/workflows/test_bwc.yml
```diff
@@ -0,0 +1,45 @@
...
+
+jobs:
+  Build-ff-linux:
+    strategy:
+      matrix:
+        java: [11,17,21]
+      fail-fast: false
+
+    name: Test Flow Framework BWC
+    runs-on: ubuntu-latest
```
owaiskazi19, line 18: We should at least add windows platform too here.

.github/workflows/test_bwc.yml
```diff
@@ -0,0 +1,45 @@
...
+          distribution: 'temurin'
+          java-version: ${{ matrix.java }}
+
+      - name: Checkout Flow Framework
+        uses: actions/checkout@v4
+
+      - name: Assemble Flow Framework
+        run: |
+          plugin_version=`./gradlew properties -q | grep "opensearch_build:" | awk '{print $2}'`
+          echo plugin_version $plugin_version
```
owaiskazi19, line 33:
```suggestion
          echo "Plugin Version:" $plugin_version
```

owaiskazi19 (commented):
We can improve on the manual copying of distribution to zip in the future

src/test/java/org/opensearch/flowframework/bwc/FlowFrameworkBackwardsCompatibilityIT.java
dbwiddis: Personally I think it’s more readable as is. :)

src/test/java/org/opensearch/flowframework/bwc/FlowFrameworkBackwardsCompatibilityIT.java
dbwiddis: Also we don’t actually need to cast.  We could just leave the strings as object type and the key lookup would still work.

.github/workflows/test_bwc.yml
dbwiddis: This lets us inspect the logs to see that it was successful.

.github/workflows/test_bwc.yml
dbwiddis: Are you going to submit all these changes to the AD bwc yaml?

.github/workflows/test_bwc.yml
dbwiddis: Yes we can, but this lets us inspect the logs if something goes wrong, to see what zip files were already in the directory.

.github/workflows/test_bwc.yml
dbwiddis: Same answer. This is useful information in the GitHub action log to show if something went wrong in the assemble step.  See [this example in AD](https://github.com/opensearch-project/anomaly-detection/actions/runs/8196784809/job/22417700186#step:5:82).

.github/workflows/test_bwc.yml
dbwiddis: I like it matching the variable name as-is.  Here's how it looks in the debug log output:
```
Run plugin_version=`./gradlew properties -q | grep "opensearch_build:" | awk '{print $2}'`
plugin_version 3.0.0.0-SNAPSHOT
```

.github/workflows/test_bwc.yml
dbwiddis: I tried, script doesn't work as is and would need tweaking:
```
Run plugin_version=`./gradlew properties -q | grep "opensearch_build:" | awk '{print $2}'`
plugin_version=./gradlew: D:\a\_temp\a25bdbdb-202a-4aa2-bc52-9ebd85a8cd4a.ps1:2
Line |
   2 |  plugin_version=`./gradlew properties -q | grep "opensearch_build:" |  …
     |  ~~~~~~~~~~~~~~~~~~~~~~~~~
     | The term 'plugin_version=./gradlew' is not recognized as a name of a cmdlet, function, script file, or
     | executable program. Check the spelling of the name, or if a path was included, verify that the path is correct
     | and try again.
```
This PR is not about BWCs, can we please move all requests to change BWC behavior into #555?

owaiskazi19: It's fine to move it to #555. We should have had a separate PR for adding BWC tests in the project.

owaiskazi19 (approved):
LGTM! Thanks for addressing the comments.

PR #558, opened 2024-03-11
Adding create ingest pipeline step

### LINKED ISSUE #24: [FEATURE] Create an ingest pipeline workflow step

amitgalitz:
### Description

Created ingest pipeline step so we can create a pipeline through a template. This supports any number of processors, and we support all processors here due to the limited validation. The pipeline steps also supports taking in a model ID currently from a previous step through substitution.

The two required fields for a pipeline are the pipelineId which is also referred to as pipeline name, and configurations which refer to the configurations of the pipeline, this includes the processors list, the description, tag or other optional lists. In order to support any ingest processor out there without ammending this step each time a new one is added, we convert the configurations given to a string and pass that string without extra validation on our end to the create pipeline API. 

Examples:

```
                {
                    "id": "create_ingest_pipeline",
                    "type": "create_ingest_pipeline",
                    "previous_node_inputs": {
                        "deploy_openai_model": "model_id"
                    },
                    "user_inputs": {
                        "pipeline_id": "text-embedding-3",
                        "configurations": {
                            "description": "A text embedding pipeline",
                            "processors": [
                                {
                                    "text_embedding": {
                                        "model_id": "${{deploy_openai_model.model_id}}",
                                        "field_map": {
                                            "passage_text": "passage_embedding"
                                        }
                                    }
                                }
                            ]
                        }
                    }
  ```
  
  ```
                  {
                    "id": "create_ingest_pipeline_2",
                    "type": "create_ingest_pipeline",
                    "previous_node_inputs": {
                        "deploy_openai_model": "model_id"
                    },
                    "user_inputs": {
                        "pipeline_id": "append-1",
                        "configurations": {
                            "description": "Pipeline that appends event type",
                            "processors": [
                                {
                                    "append": {
                                        "field": "event_types",
                                        "value": [
                                            "page_view"
                                        ]
                                    }
                                },
                                {
                                    "drop": {
                                        "if": "ctx.user_info.contains('password') || ctx.user_info.contains('credit card')"
                                    }
                                }
                            ]
                        }
                    }
                }
 ```      
          

### Issues Resolved
resolves #24 

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -175,6 +175,8 @@ dependencies {
     implementation 'com.amazonaws:aws-encryption-sdk-java:2.4.1'
     implementation 'org.bouncycastle:bcprov-jdk18on:1.77'
     api "org.apache.httpcomponents.core5:httpcore5:5.2.2"
+    implementation("com.fasterxml.jackson.core:jackson-annotations:${versions.jackson}")
+    implementation("com.fasterxml.jackson.core:jackson-databind:${versions.jackson_databind}")
 
     // ZipArchive dependencies used for integration tests
     zipArchive group: 'org.opensearch.plugin', name:'opensearch-ml-plugin', version: "${opensearch_build}"
--- a/src/main/java/org/opensearch/flowframework/util/ParseUtils.java
+++ b/src/main/java/org/opensearch/flowframework/util/ParseUtils.java
@@ -374,7 +374,7 @@ public class ParseUtils {
     }
 
     /**
-     * Generates a string based on an arbitrary String to object map sing Jackson
+     * Generates a string based on an arbitrary String to object map using Jackson
      * @param map content map
      * @return instance of the string
      * @throws JsonProcessingException JsonProcessingException from Jackson for issues processing map
--- a/src/main/java/org/opensearch/flowframework/util/ParseUtils.java
+++ b/src/main/java/org/opensearch/flowframework/util/ParseUtils.java
@@ -8,6 +8,9 @@
  */
 package org.opensearch.flowframework.util;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.opensearch.client.Client;
@@ -355,5 +358,18 @@ public class ParseUtils {
             }
         }
         return value;
+    }
+
+    /**
+     * Generates a string based on an arbitrary String to object map sing Jackson
+     * @param map content map
+     * @return instance of the string
+     * @throws JsonProcessingException JsonProcessingException from Jackson for issues processing map
+     */
+    public static String parseArbitraryStringToObjectMapToString(Map<String, Object> map) throws JsonProcessingException {
+        ObjectMapper mapper = new ObjectMapper();
+        // Convert the map to a JSON string
+        String mappedString = mapper.writeValueAsString(map);
+        return mappedString;
     }
 }
--- a/src/main/java/org/opensearch/flowframework/util/ParseUtils.java
+++ b/src/main/java/org/opensearch/flowframework/util/ParseUtils.java
@@ -56,7 +56,8 @@ public class ParseUtils {
     private static final Logger logger = LogManager.getLogger(ParseUtils.class);
 
     // Matches ${{ foo.bar }} (whitespace optional) with capturing groups 1=foo, 2=bar
-    private static final Pattern SUBSTITUTION_PATTERN = Pattern.compile("\\$\\{\\{\\s*(.+)\\.(.+?)\\s*\\}\\}");
+    // private static final Pattern SUBSTITUTION_PATTERN = Pattern.compile("\\$\\{\\{\\s*(.+)\\.(.+?)\\s*\\}\\}");
+    private static final Pattern SUBSTITUTION_PATTERN = Pattern.compile("\\$\\{\\{\\s*([\\w_]+)\\.([\\w_]+)\\s*\\}\\}");
 
     private ParseUtils() {}
 
@@ -344,13 +345,25 @@ public class ParseUtils {
     private static Object conditionallySubstitute(Object value, Map<String, WorkflowData> outputs, Map<String, String> params) {
         if (value instanceof String) {
             Matcher m = SUBSTITUTION_PATTERN.matcher((String) value);
-            if (m.matches()) {
-                // Try matching a previous step+value pair
-                WorkflowData data = outputs.get(m.group(1));
-                if (data != null && data.getContent().containsKey(m.group(2))) {
-                    return data.getContent().get(m.group(2));
+            StringBuilder result = new StringBuilder();
+            while (m.find()) {
+                // outputs content map contains values for previous node input (e.g: deploy_openai_model.model_id)
+                // Check first if the substitution is looking for the same key, value pair and if yes
+                // then replace it with the key value pair in the inputs map
+                String replacement = m.group(0);
+                if (outputs.containsKey(m.group(1)) && outputs.get(m.group(1)).getContent().containsKey(m.group(2))) {
+                    // Extract the key for the inputs (e.g., "model_id" from ${{deploy_openai_model.model_id}})
+                    String key = m.group(2);
+                    if (outputs.get(m.group(1)).getContent().get(key) instanceof String) {
+                        replacement = (String) outputs.get(m.group(1)).getContent().get(key);
+                        // Replace the whole sequence with the value from the map
+                        m.appendReplacement(result, Matcher.quoteReplacement(replacement));
+                    }
                 }
             }
+            m.appendTail(result);
+            value = result.toString();
+
             // Replace all params if present
             for (Entry<String, String> e : params.entrySet()) {
                 String regex = "\\$\\{\\{\\s*" + Pattern.quote(e.getKey()) + "\\s*\\}\\}";
--- a/src/main/java/org/opensearch/flowframework/workflow/CreateIngestPipelineStep.java
+++ b/src/main/java/org/opensearch/flowframework/workflow/CreateIngestPipelineStep.java
@@ -13,7 +13,6 @@ import org.apache.logging.log4j.Logger;
 import org.opensearch.ExceptionsHelper;
 import org.opensearch.action.ingest.PutPipelineRequest;
 import org.opensearch.action.support.PlainActionFuture;
-import org.opensearch.action.support.master.AcknowledgedResponse;
 import org.opensearch.client.Client;
 import org.opensearch.client.ClusterAdminClient;
 import org.opensearch.common.xcontent.XContentType;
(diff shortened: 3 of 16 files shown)

### CODE REVIEW

File: src/main/java/org/opensearch/flowframework/workflow/CreateIngestPipelineStep.java [hunk 1]
```diff
@@ -57,7 +51,7 @@ public class CreateIngestPipelineStep implements WorkflowStep {
     private final FlowFrameworkIndicesHandler flowFrameworkIndicesHandler;
 
     /**
-     * Instantiates a new CreateIngestPipelineStep
+     * Instantiates a new CreateIngestPipelineStepDraft
```
joshpalis, line 54: Nit : Lets change this to CreateIngestPipelineStep

src/main/java/org/opensearch/flowframework/workflow/CreateIngestPipelineStep.java
```diff
@@ -187,12 +112,75 @@ public PlainActionFuture<WorkflowData> execute(
...
+            Map<String, Object> inputs = ParseUtils.getInputsFromPreviousSteps(
+                requiredKeys,
+                optionalKeys,
+                currentNodeInputs,
+                outputs,
+                previousNodeInputs,
+                params
+            );
+
+            String pipelineId = (String) inputs.get(PIPELINE_ID);
```
joshpalis, line 141: Perhaps we can fail this earlier by checking if the pipeline ID is null here. If it is, then we shouldn't attempt substitution on the configuration since the request will fail either way.

joshpalis (approved):
Overall looks good to me, just a few comments

src/main/java/org/opensearch/flowframework/workflow/CreateIngestPipelineStep.java
dbwiddis: @joshpalis that null check is done in the `getInputsFromPreviousSteps` since that's a required key.

src/main/java/org/opensearch/flowframework/workflow/CreateIngestPipelineStep.java
```diff
@@ -187,12 +112,75 @@ public PlainActionFuture<WorkflowData> execute(
...
+                if (params.containsKey(matcher.group(1)) && params.get(matcher.group(1)).equals(matcher.group(2))) {
+                    // Extract the key for the inputs (e.g., "model_id" from ${{deploy_openai_model.model_id}})
+                    String key = matcher.group(2);
+                    if (inputs.containsKey(key)) {
+                        // Replace the whole sequence with the value from the map
+                        matcher.appendReplacement(result, (String) inputs.get(key));
+                    }
+                }
+            }
+            matcher.appendTail(result);
```
dbwiddis, line 163: The existing parsing that gets you `inputs` probably already handles this format (and should actually work after the bug fix #559 ... this whole section here seems to duplicate that).  The only thing missing is that we may want more optional inputs based on searching the configuration.  You could probably take the `conditionallySubstitute()` method and modify it to find the keys and add them to the optional inputs prior to calling that.

dbwiddis (changes requested):
Not sure how much of this is working around the bug fixed in #559, but can definitely improve the parsing...

src/main/java/org/opensearch/flowframework/workflow/CreateIngestPipelineStep.java
```diff
@@ -187,12 +112,75 @@ public PlainActionFuture<WorkflowData> execute(
...
+
+            if (result == null || pipelineId == null) {
```
jackiehanyang, line 165: This check seems unnecessary. `result` cannot be null because a new `StringBuffer` is always instantiated at line 149, and `pipelineId` is fetched at line 141. If it were null, it would have caused NPE earlier.

dbwiddis: > it would have caused NPE earlier.

Actually it would have thrown this exception:
https://github.com/opensearch-project/flow-framework/blob/4bf6e329097610021b39b3570a5274de5a264946/src/main/java/org/opensearch/flowframework/util/ParseUtils.java#L325-L336

src/main/java/org/opensearch/flowframework/workflow/CreateIngestPipelineStep.java
amitgalitz: removed null checks as they happen earlier

build.gradle
```diff
@@ -175,6 +175,8 @@ dependencies {
     implementation 'com.amazonaws:aws-encryption-sdk-java:2.4.1'
     implementation 'org.bouncycastle:bcprov-jdk18on:1.77'
     api "org.apache.httpcomponents.core5:httpcore5:5.2.2"
+    implementation("com.fasterxml.jackson.core:jackson-annotations:${versions.jackson}")
+    implementation("com.fasterxml.jackson.core:jackson-databind:${versions.jackson_databind}")
```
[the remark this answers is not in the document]
amitgalitz: we force jackson-core version which I see done in other plugins. Here the versions.jackson comes from opensearch core.

src/main/java/org/opensearch/flowframework/workflow/CreateIngestPipelineStep.java
```diff
@@ -77,93 +71,24 @@ public PlainActionFuture<WorkflowData> execute(
...
-                logger.info("Created ingest pipeline : " + putPipelineRequest.getId());
-
-                try {
-                    String resourceName = getResourceByWorkflowStep(getName());
                     flowFrameworkIndicesHandler.updateResourceInStateIndex(
                         currentNodeInputs.getWorkflowId(),
                         currentNodeId,
                         getName(),
-                        putPipelineRequest.getId(),
+                        currentNodeInputs.getContent().get(PIPELINE_ID).toString(),
```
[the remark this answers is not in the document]
amitgalitz: moved

src/main/java/org/opensearch/flowframework/workflow/CreateIngestPipelineStep.java
```diff
@@ -187,12 +112,75 @@ public PlainActionFuture<WorkflowData> execute(
...
+
+            if (result == null || pipelineId == null) {
+                // Required workflow data not found
+                createIngestPipelineFuture.onFailure(
+                    new FlowFrameworkException(
+                        "Failed to create ingest pipeline for " + currentNodeId + ", required inputs not found",
+                        RestStatus.BAD_REQUEST
+                    )
+                );
+            }
```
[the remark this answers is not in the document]
amitgalitz: removed theses

src/main/java/org/opensearch/flowframework/workflow/CreateIngestPipelineStep.java
amitgalitz: Moved the parsing into the ParseUtils. conditionallySubstitute(), lmk if this looks good and if its what you meant. I tested we can still do param substitution through rest calls or previous nodes input incase of pipeline configuration

src/main/java/org/opensearch/flowframework/workflow/CreateIngestPipelineStep.java
amitgalitz: removed check

src/main/java/org/opensearch/flowframework/util/ParseUtils.java
```diff
@@ -356,4 +372,17 @@ private static Object conditionallySubstitute(Object value, Map<String, Workflow
         }
         return value;
     }
+
+    /**
+     * Generates a string based on an arbitrary String to object map sing Jackson
```
dbwiddis, line 377: nit: "sing" -> "using"

src/main/java/org/opensearch/flowframework/util/ParseUtils.java
```diff
@@ -356,4 +372,17 @@ private static Object conditionallySubstitute(Object value, Map<String, Workflow
...
+     * @param map content map
+     * @return instance of the string
+     * @throws JsonProcessingException JsonProcessingException from Jackson for issues processing map
+     */
+    public static String parseArbitraryStringToObjectMapToString(Map<String, Object> map) throws JsonProcessingException {
```
dbwiddis, line 382: This functionality looks like it should probably exist in core, such as in `o.o.core.common.Strings` util class, or in OpenSearch's `o.o.index.mapper.ObjectMapper` class.   Jackson is a regular source of CVEs and we've made efforts to remove parsing like this from core for good reasons.

I'm OK with merging this now but can we open an issue to investigate other alternatives?

src/main/java/org/opensearch/flowframework/workflow/CreateIngestPipelineStep.java
```diff
@@ -77,122 +68,85 @@ public PlainActionFuture<WorkflowData> execute(
...
-                logger.info("Created ingest pipeline : " + putPipelineRequest.getId());
 
-                try {
+            String pipelineId = (String) inputs.get(PIPELINE_ID);
+            String configurations = (String) inputs.get(CONFIGURATIONS);
+
+            byte[] byteArr = configurations.getBytes(StandardCharsets.UTF_8);
+            BytesReference configurationsBytes = new BytesArray(byteArr);
+
+            ActionListener<AcknowledgedResponse> actionListener = new ActionListener<>() {
```
dbwiddis, line 92: You could just define this inline on line 146, where it'd be a bit more logically readable.

dbwiddis (approved):
LGTM! A few suggestions.

src/main/java/org/opensearch/flowframework/workflow/CreateIngestPipelineStep.java
amitgalitz: moved

src/main/java/org/opensearch/flowframework/util/ParseUtils.java
amitgalitz: I see Jackson is used in core in lots of other plugins so I thought this would be fine, but I can open an issue to investigate

owaiskazi19:
@amitgalitz can you raise a manual backport PR?

PR #561, opened 2024-03-11
Removes unnecessary gradle properties

joshpalis:
### Description
Gradle properties seem to be directly pulled from OpenSearch gradle properties, this PR removes them as they are not necessary for this plugin. For example, `systemProp.org.gradle.warning.mode=fail` would fail a build outright when gradle APIs marked for future deprecation are used. This makes sense to set in OpenSearch to force a gradle version upgrade, but in this plugin's case, gradle version is only updated once OpenSearch does. 

Testing locally, the build and integration test pass with these properties removed.


dbwiddis (approved):
I wasted many hours figuring out that my BWCs were failing on this dang switch.

amitgalitz:
I think this is used `options.forkOptions.memoryMaximumSize=2g` for the integTest cluster to determine heap size, not sure how it gets default otherwise

dbwiddis:
> I think this is used `options.forkOptions.memoryMaximumSize=2g` for the integTest cluster to determine heap size, not sure how it gets default otherwise

Default if not specified is 1g.

Tests seem to be passing with this removed, though.

amitgalitz:
> > I think this is used `options.forkOptions.memoryMaximumSize=2g` for the integTest cluster to determine heap size, not sure how it gets default otherwise
> 
> Default if not specified is 512mb.
> 
> Tests seem to be passing with this removed, though.

sounds good, just checked also that :run also works

amitgalitz (approved):

owaiskazi19 (approved):

PR #577, opened 2024-03-14
Adding new exception type for workflow step failures

amitgalitz:
### Description
Adds a new exception type to differentiate between errors that are due to flow-framework's fault during provisioning vs errors that are due to the downstream plugin's faults.

Also fixed error handling on deprovision:
previously if we gave a payload to the deprovision request we got a 500 error of `Channel is already closed`

The two error types: 

```
"error": "org.opensearch.flowframework.exception.WorkflowStepException during step create_ingest_pipeline, restStatus: BAD_REQUEST"
"error": "org.opensearch.flowframework.exception.WorkflowStepException during step create_ingest_pipeline, restStatus: BAD_REQUEST"
```

### DIFF
--- a/src/main/java/org/opensearch/flowframework/rest/RestCreateWorkflowAction.java
+++ b/src/main/java/org/opensearch/flowframework/rest/RestCreateWorkflowAction.java
@@ -20,7 +20,6 @@ import org.opensearch.core.xcontent.XContentParser;
 import org.opensearch.flowframework.common.DefaultUseCases;
 import org.opensearch.flowframework.common.FlowFrameworkSettings;
 import org.opensearch.flowframework.exception.FlowFrameworkException;
-import org.opensearch.flowframework.exception.WorkflowStepException;
 import org.opensearch.flowframework.model.Template;
 import org.opensearch.flowframework.transport.CreateWorkflowAction;
 import org.opensearch.flowframework.transport.WorkflowRequest;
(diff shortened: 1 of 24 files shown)


src/main/java/org/opensearch/flowframework/rest/RestCreateWorkflowAction.java
```diff
@@ -123,9 +124,14 @@ protected RestChannelConsumer prepareRequest(RestRequest request, NodeClient cli
...
-                        ? (FlowFrameworkException) exception
-                        : new FlowFrameworkException("Failed to create workflow.", ExceptionsHelper.status(exception));
+                    FlowFrameworkException ex;
+                    if (exception instanceof WorkflowStepException) {
+                        ex = (WorkflowStepException) exception;
+                    } else if (exception instanceof FlowFrameworkException) {
+                        ex = (FlowFrameworkException) exception;
+                    } else {
+                        ex = new FlowFrameworkException("Failed to create workflow.", ExceptionsHelper.status(exception));
+                    }
```
owaiskazi19, line 134:
```suggestion
                   FlowFrameworkException ex = (exception instanceof WorkflowStepException)
                             ? (WorkflowStepException) exception
                             : (exception instanceof FlowFrameworkException)
                              ? (FlowFrameworkException) exception
                       : new FlowFrameworkException("Failed to create workflow.", ExceptionsHelper.status(exception));
```

owaiskazi19 (commented):
Tests are not failing currently because `WorkflowStepException`  extends`FlowFrameworkException` but we should change the rest test cases to check for `WorkflowStepException`.

amitgalitz:
> Tests are not failing currently because `WorkflowStepException` extends`FlowFrameworkException` but we should change the rest test cases to check for `WorkflowStepException`.

Actually there is another reason for this also, that makes me debate if we even need this here. We actually don't return the WorkflowStepException in the rest here ever because we just write it to the error field in the index if it happens during the step. User gets the workflow ID back

amitgalitz:
> > Tests are not failing currently because `WorkflowStepException` extends`FlowFrameworkException` but we should change the rest test cases to check for `WorkflowStepException`.
> 
> Actually there is another reason for this also, that makes me debate if we even need this here. We actually don't return the WorkflowStepException in the rest here ever because we just write it to the error field in the index if it happens during the step. User gets the workflow ID back

@joshpalis lmk if you have any opposition to just adding the WorkflowStepException and not changing the rest classes due to above reason

joshpalis:
> @joshpalis lmk if you have any opposition to just adding the WorkflowStepException and not changing the rest classes due to above reason

No concerns, looks good to me

joshpalis (approved):

src/main/java/org/opensearch/flowframework/rest/RestCreateWorkflowAction.java
amitgalitz: removed this code completely

owaiskazi19 (approved):

PR #583, opened 2024-03-15
Adding default use cases

amitgalitz:
### Description

The reason I went with a separate defaults file for now is so we don't make breaking changes on our template if customers gives us feedback for how to this differently. 

Some concerns I have:

1. Customers might have a hard time knowing what to substitute, for example if they want to change connector post functions they need to know this key:
`"create_connector.actions.pre_process_function": "<content>"`
or for sparse search: `"register_local_sparse_encoding_model.function_name"`

2. If we want to make these more flexibly its going to be hard to do substitution on things like the index mapping unless user gives a whole string version of there mapping, settings and alias and we make the entire configuration a substitution.

3. Currently you can see I have two versions of create connector - deploy - register. This is due that some connectors have more optional parameters then the others. One solution is having many files for each variation, another is to potentially have the connectors made with empty fields (for example `truncate` is only needed in cohere default so if we have that for an OpenAI connector it will just be there and unused).

I only added 3 fully working defaults here for now, will have lots more to add if this approach seems fine. I want to also add a good amount of local models so that at least for fast POC or demo users can have a one click set up of an entire use case with 0 additional input like:
```
POST /_plugins/_flow_framework/workflow?use_case=local_neural_sparse_search
{}
or
POST /_plugins/_flow_framework/workflow?use_case=local_neural_sparse_search
```

another example with current use cases in PR:

```
POST /_plugins/_flow_framework/workflow?use_case=cohere-embedding_model_deploy
{
    "create_connector.credential.key" : "123"
}
```

### Issues Resolved
_List any issues this PR will resolve, e.g. Closes [...]._ 

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -180,6 +180,7 @@ dependencies {
 
     // ZipArchive dependencies used for integration tests
     zipArchive group: 'org.opensearch.plugin', name:'opensearch-ml-plugin', version: "${opensearch_build}"
+
     secureIntegTestPluginArchive group: 'org.opensearch.plugin', name:'opensearch-security', version: "${opensearch_build}"
 
     configurations.all {
--- a/src/main/java/org/opensearch/flowframework/common/DefaultUseCases.java
+++ b/src/main/java/org/opensearch/flowframework/common/DefaultUseCases.java
@@ -1,0 +1,120 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+package org.opensearch.flowframework.common;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.opensearch.core.rest.RestStatus;
+import org.opensearch.flowframework.exception.FlowFrameworkException;
+
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+/**
+ * Enum encapsulating the different default use cases and templates we have stored
+ */
+public enum DefaultUseCases {
+
+    /** defaults file and substitution ready template for OpenAI embedding model */
+    OPEN_AI_EMBEDDING_MODEL_DEPLOY(
+        "open_ai_embedding_model_deploy",
+        "defaults/open-ai-embedding-defaults.json",
+        "substitutionTemplates/deploy-remote-model-template.json"
+    ),
+    /** defaults file and substitution ready template for cohere embedding model */
+    COHERE_EMBEDDING_MODEL_DEPLOY(
+        "cohere-embedding_model_deploy",
+        "defaults/cohere-embedding-defaults.json",
+        "substitutionTemplates/deploy-remote-model-template-extra-params.json"
+    ),
+    LOCAL_NEURAL_SPARSE_SEARCH(
+        "local_neural_sparse_search",
+        "defaults/local-sparse-search-defaults.json",
+        "substitutionTemplates/neural-sparse-local-template.json"
+    ),
+    /** defaults file and substitution ready template for cohere embedding model */ // TODO: not finalized
+    COHERE_EMBEDDING_MODEL_DEPLOY_SEMANTIC_SEARCH(
+        "cohere-embedding_model_deploy_semantic_search",
+        "defaults/cohere-embedding-defaults.json",
+        "substitutionTemplates/deploy-remote-model-template-v1.json"
+    );
+
+    private final String useCaseName;
+    private final String defaultsFile;
+    private final String substitutionReadyFile;
+    private static final Logger logger = LogManager.getLogger(DefaultUseCases.class);
+    private static final Set<String> allResources = Stream.of(values()).map(DefaultUseCases::getDefaultsFile).collect(Collectors.toSet());
+
+    DefaultUseCases(String useCaseName, String defaultsFile, String substitutionReadyFile) {
+        this.useCaseName = useCaseName;
+        this.defaultsFile = defaultsFile;
+        this.substitutionReadyFile = substitutionReadyFile;
+    }
+
+    /**
+     * Returns the useCaseName for the given enum Constant
+     * @return the useCaseName of this use case.
+     */
+    public String getUseCaseName() {
+        return useCaseName;
+    }
+
+    /**
+     * Returns the defaultsFile for the given enum Constant
+     * @return the defaultsFile of this for the given useCase.
+     */
+    public String getDefaultsFile() {
+        return defaultsFile;
+    }
+
+    /**
+     * Returns the substitutionReadyFile for the given enum Constant
+     * @return the substitutionReadyFile of the given useCase
+     */
+    public String getSubstitutionReadyFile() {
+        return substitutionReadyFile;
+    }
+
+    /**
+     * Gets the defaultsFile based on the given use case.
+     * @param useCaseName name of the given use case
+     * @return the deafultsFile for that usecase
+     * @throws FlowFrameworkException if the use case doesn't exist in enum
+     */
+    public static String getDefaultsFileByUseCaseName(String useCaseName) throws FlowFrameworkException {
+        if (useCaseName != null && !useCaseName.isEmpty()) {
+            for (DefaultUseCases mapping : values()) {
+                if (useCaseName.equals(mapping.getUseCaseName())) {
+                    return mapping.getDefaultsFile();
+                }
+            }
+        }
+        logger.error("Unable to find defaults file for use case: {}", useCaseName);
+        throw new FlowFrameworkException("Unable to find defaults file for use case: " + useCaseName, RestStatus.BAD_REQUEST);
+    }
+
+    /**
+     * Gets the substitutionReadyFile based on the given use case
+     * @param useCaseName name of the given use case
+     * @return the substitutionReadyFile which has the template
+     * @throws FlowFrameworkException if the use case doesn't exist in enum
+     */
+    public static String getSubstitutionReadyFileByUseCaseName(String useCaseName) throws FlowFrameworkException {
+        if (useCaseName != null && !useCaseName.isEmpty()) {
+            for (DefaultUseCases mapping : values()) {
+                if (mapping.getUseCaseName().equals(useCaseName)) {
+                    return mapping.getSubstitutionReadyFile();
+                }
+            }
+        }
+        logger.error("Unable to find substitution ready file for use case: {}", useCaseName);
+        throw new FlowFrameworkException("Unable to find substitution ready file for use case: " + useCaseName, RestStatus.BAD_REQUEST);
+    }
+}
--- a/src/main/java/org/opensearch/flowframework/common/DefaultUseCases.java
+++ b/src/main/java/org/opensearch/flowframework/common/DefaultUseCases.java
@@ -38,12 +38,6 @@ public enum DefaultUseCases {
         "local_neural_sparse_search",
         "defaults/local-sparse-search-defaults.json",
         "substitutionTemplates/neural-sparse-local-template.json"
-    ),
-    /** defaults file and substitution ready template for cohere embedding model */ // TODO: not finalized
-    COHERE_EMBEDDING_MODEL_DEPLOY_SEMANTIC_SEARCH(
-        "cohere-embedding_model_deploy_semantic_search",
-        "defaults/cohere-embedding-defaults.json",
-        "substitutionTemplates/deploy-remote-model-template-v1.json"
     );
 
     private final String useCaseName;
--- a/src/main/java/org/opensearch/flowframework/common/DefaultUseCases.java
+++ b/src/main/java/org/opensearch/flowframework/common/DefaultUseCases.java
@@ -13,10 +13,6 @@ import org.apache.logging.log4j.Logger;
 import org.opensearch.core.rest.RestStatus;
 import org.opensearch.flowframework.exception.FlowFrameworkException;
 
-import java.util.Set;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
-
 /**
  * Enum encapsulating the different default use cases and templates we have stored
  */
@@ -45,7 +41,6 @@ public enum DefaultUseCases {
     private final String defaultsFile;
     private final String substitutionReadyFile;
     private static final Logger logger = LogManager.getLogger(DefaultUseCases.class);
-    private static final Set<String> allResources = Stream.of(values()).map(DefaultUseCases::getDefaultsFile).collect(Collectors.toSet());
 
     DefaultUseCases(String useCaseName, String defaultsFile, String substitutionReadyFile) {
         this.useCaseName = useCaseName;
@@ -85,9 +80,9 @@ public enum DefaultUseCases {
      */
     public static String getDefaultsFileByUseCaseName(String useCaseName) throws FlowFrameworkException {
         if (useCaseName != null && !useCaseName.isEmpty()) {
-            for (DefaultUseCases mapping : values()) {
-                if (useCaseName.equals(mapping.getUseCaseName())) {
-                    return mapping.getDefaultsFile();
+            for (DefaultUseCases usecase : values()) {
+                if (useCaseName.equals(usecase.getUseCaseName())) {
+                    return usecase.getDefaultsFile();
                 }
             }
         }
@@ -103,9 +98,9 @@ public enum DefaultUseCases {
      */
     public static String getSubstitutionReadyFileByUseCaseName(String useCaseName) throws FlowFrameworkException {
         if (useCaseName != null && !useCaseName.isEmpty()) {
-            for (DefaultUseCases mapping : values()) {
-                if (mapping.getUseCaseName().equals(useCaseName)) {
-                    return mapping.getSubstitutionReadyFile();
+            for (DefaultUseCases useCase : values()) {
+                if (useCase.getUseCaseName().equals(useCaseName)) {
+                    return useCase.getSubstitutionReadyFile();
                 }
             }
         }
--- a/src/main/java/org/opensearch/flowframework/common/DefaultUseCases.java
+++ b/src/main/java/org/opensearch/flowframework/common/DefaultUseCases.java
@@ -34,6 +34,7 @@ public enum DefaultUseCases {
         "defaults/cohere-embedding-defaults.json",
         "substitutionTemplates/deploy-remote-model-template-extra-params.json"
     ),
+    /** defaults file and substitution ready template for local neural sparse model and ingest pipeline*/
     LOCAL_NEURAL_SPARSE_SEARCH(
         "local_neural_sparse_search",
         "defaults/local-sparse-search-defaults.json",
@@ -79,7 +80,7 @@ public enum DefaultUseCases {
     /**
      * Gets the defaultsFile based on the given use case.
      * @param useCaseName name of the given use case
-     * @return the deafultsFile for that usecase
+     * @return the defaultsFile for that usecase
      * @throws FlowFrameworkException if the use case doesn't exist in enum
      */
     public static String getDefaultsFileByUseCaseName(String useCaseName) throws FlowFrameworkException {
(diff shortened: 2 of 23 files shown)


src/main/java/org/opensearch/flowframework/util/ParseUtils.java
```diff
@@ -385,4 +401,17 @@ public static String parseArbitraryStringToObjectMapToString(Map<String, Object>
...
     }
+
+    /**
+     * Generates a String to String map based on a Json File
+     * @param path file path
+     * @return instance of the string
+     * @throws JsonProcessingException JsonProcessingException from Jackson for issues processing map
+     */
+    public static Map<String, String> parseJsonFileToStringToStringMap(String path) throws IOException {
+        ObjectMapper mapper = new ObjectMapper();
```
jackiehanyang, line 412: nit: consider reusing a single ObjectMapper instance instead of creating a new one each time the method is called

src/main/java/org/opensearch/flowframework/util/ParseUtils.java
```diff
@@ -385,4 +401,17 @@ public static String parseArbitraryStringToObjectMapToString(Map<String, Object>
...
+        String jsonContent = resourceToString(path);
+        Map<String, String> mappedJsonFile = mapper.readValue(jsonContent, Map.class);
```
jackiehanyang, line 414: curiosity question: is all values in the JSON file are text? Asking because if the JSON contains nested objects or arrays, this `Map<String, String>` might fail

src/main/java/org/opensearch/flowframework/rest/RestCreateWorkflowAction.java
```diff
@@ -112,11 +117,54 @@ protected RestChannelConsumer prepareRequest(RestRequest request, NodeClient cli
...
+            Map<String, String> useCaseDefaultsMap = Collections.emptyMap();
+            if (useCase != null) {
+                String json = ParseUtils.resourceToString("/" + DefaultUseCases.getSubstitutionReadyFileByUseCaseName(useCase));
+                String defaultsFilePath = DefaultUseCases.getDefaultsFileByUseCaseName(useCase);
+                useCaseDefaultsMap = ParseUtils.parseJsonFileToStringToStringMap("/" + defaultsFilePath);
+
+                if (request.hasContent()) {
+                    try {
+                        XContentParser parser = request.contentParser();
+                        ensureExpectedToken(XContentParser.Token.START_OBJECT, parser.nextToken(), parser);
```
jackiehanyang, line 131: nit duplicate code in both if and else branch

src/main/java/org/opensearch/flowframework/common/DefaultUseCases.java
```diff
@@ -0,0 +1,114 @@
...
+     * @return the substitutionReadyFile of the given useCase
+     */
+    public String getSubstitutionReadyFile() {
+        return substitutionReadyFile;
+    }
+
+    /**
+     * Gets the defaultsFile based on the given use case.
+     * @param useCaseName name of the given use case
+     * @return the deafultsFile for that usecase
```
jackiehanyang, line 82: nit: defaultsFile

jackiehanyang:
Some nit comments on this draft pr, will review again once the pr is ready for review.

src/main/java/org/opensearch/flowframework/rest/RestCreateWorkflowAction.java
```diff
@@ -112,11 +117,54 @@ protected RestChannelConsumer prepareRequest(RestRequest request, NodeClient cli
...
+                        // updates the default params with anything user has given that matches
+                        for (Map.Entry<String, String> userDefaultsEntry : userDefaults.entrySet()) {
+                            if (useCaseDefaultsMap.containsKey(userDefaultsEntry.getKey())) {
+                                useCaseDefaultsMap.put(userDefaultsEntry.getKey(), userDefaultsEntry.getValue());
+                            }
+                        }
+                    } catch (Exception ex) {
+                        String errorMessage = "failure parsing request body when a use case is given";
+                        logger.error(errorMessage, ex);
+                        throw new FlowFrameworkException(errorMessage, ExceptionsHelper.status(ex));
```
joshpalis, line 142: This may throw an internal server error if the parseStringToStringMap throws an IOException. The ExceptionsHelper has no case for IOException and defaults the status code to 500. Lets change that

src/main/java/org/opensearch/flowframework/util/ParseUtils.java
```diff
@@ -107,6 +107,8 @@ public static String resourceToString(String path) throws IOException {
     public static void buildStringToStringMap(XContentBuilder xContentBuilder, Map<?, ?> map) throws IOException {
         xContentBuilder.startObject();
         for (Entry<?, ?> e : map.entrySet()) {
+            String key = (String) e.getKey();
```
joshpalis, line 110: key and value are never used here

joshpalis (approved):
Overall looks good to me, thanks @amitgalitz for this, definitely improves ease of use. Just had a few comments

src/main/java/org/opensearch/flowframework/util/ParseUtils.java
amitgalitz: Yeah right now this method is only used for our defaults file which always will be Map<String, String>.In 2.14 we might find ways for customers to add there own, in which we will have to do stricter checking.

src/main/java/org/opensearch/flowframework/rest/RestCreateWorkflowAction.java
amitgalitz: moved

src/main/java/org/opensearch/flowframework/util/ParseUtils.java
amitgalitz: moved as a static final variable

build.gradle
```diff
@@ -180,6 +180,7 @@ dependencies {
 
     // ZipArchive dependencies used for integration tests
     zipArchive group: 'org.opensearch.plugin', name:'opensearch-ml-plugin', version: "${opensearch_build}"
+
```
owaiskazi19, line 183: Extra line

src/main/resources/defaults/cohere-embedding-defaults.json
```diff
@@ -0,0 +1,18 @@
...
+    "create_connector.truncate": "end",
+    "create_connector.endpoint": "api.openai.com",
+    "create_connector.credential.key": "123",
+    "create_connector.actions.url": "https://api.cohere.ai/v1/embed",
+    "create_connector.actions.request_body": "{ \"texts\": ${parameters.texts}, \"truncate\": \"${parameters.truncate}\", \"model\": \"${parameters.model}\", \"input_type\": \"${parameters.input_type}\" }",
+    "create_connector.actions.pre_process_function": "connector.pre_process.cohere.embedding",
+    "create_connector.actions.post_process_function": "connector.post_process.cohere.embedding",
+    "register_remote_model.name": "Cohere english embed model",
+    "register_remote_model.description": "cohere-embedding-model"
+}
```
owaiskazi19, line 18: We should not have multiple default files rather just one and multiple objects for default values

src/main/java/org/opensearch/flowframework/common/DefaultUseCases.java
```diff
@@ -0,0 +1,115 @@
...
+        "cohere-embedding_model_deploy",
+        "defaults/cohere-embedding-defaults.json",
+        "substitutionTemplates/deploy-remote-model-template-extra-params.json"
+    ),
+    /** defaults file and substitution ready template for local neural sparse model and ingest pipeline*/
+    LOCAL_NEURAL_SPARSE_SEARCH(
+        "local_neural_sparse_search",
+        "defaults/local-sparse-search-defaults.json",
+        "substitutionTemplates/neural-sparse-local-template.json"
+    );
```
owaiskazi19, line 42: We should provide an easy way for users to add their own `default` files and `substitutionTemplate`. Probably a json/yaml file which contains the mapping of the default file and substitution template. Having an enum would require them to add here.

src/main/java/org/opensearch/flowframework/common/DefaultUseCases.java
```diff
@@ -0,0 +1,115 @@
...
+
+    private final String useCaseName;
+    private final String defaultsFile;
+    private final String substitutionReadyFile;
+    private static final Logger logger = LogManager.getLogger(DefaultUseCases.class);
+    private static final Set<String> allResources = Stream.of(values()).map(DefaultUseCases::getDefaultsFile).collect(Collectors.toSet());
```
owaiskazi19, line 48: Not using `allResources` anywhere?

If want to use somewhere, use `EnumSet.allOf(DefaultUseCases.class)` instead of `values` to avoid issues with enum order.

```java
EnumSet.allOf(DefaultUseCases.class).stream()
            .map(DefaultUseCases::getDefaultsFile)
            .collect(Collectors.toSet());
```

src/main/java/org/opensearch/flowframework/common/DefaultUseCases.java
```diff
@@ -0,0 +1,115 @@
...
+
+    /**
+     * Gets the defaultsFile based on the given use case.
+     * @param useCaseName name of the given use case
+     * @return the defaultsFile for that usecase
+     * @throws FlowFrameworkException if the use case doesn't exist in enum
+     */
+    public static String getDefaultsFileByUseCaseName(String useCaseName) throws FlowFrameworkException {
+        if (useCaseName != null && !useCaseName.isEmpty()) {
+            for (DefaultUseCases mapping : values()) {
```
owaiskazi19, line 88:
```suggestion
            for (DefaultUseCases useCase : values()) {
```

src/main/java/org/opensearch/flowframework/common/DefaultUseCases.java
```diff
@@ -0,0 +1,115 @@
...
+
+    /**
+     * Gets the substitutionReadyFile based on the given use case
+     * @param useCaseName name of the given use case
+     * @return the substitutionReadyFile which has the template
+     * @throws FlowFrameworkException if the use case doesn't exist in enum
+     */
+    public static String getSubstitutionReadyFileByUseCaseName(String useCaseName) throws FlowFrameworkException {
+        if (useCaseName != null && !useCaseName.isEmpty()) {
+            for (DefaultUseCases mapping : values()) {
```
owaiskazi19, line 106:
```suggestion
            for (DefaultUseCases useCase : values()) {
```

src/main/java/org/opensearch/flowframework/util/ParseUtils.java
amitgalitz: removed these, good catch

src/main/java/org/opensearch/flowframework/rest/RestCreateWorkflowAction.java
amitgalitz: changes so rest status is BAD_REQUEST when IO exception

src/main/java/org/opensearch/flowframework/rest/RestCreateWorkflowAction.java
```diff
@@ -112,11 +117,53 @@ protected RestChannelConsumer prepareRequest(RestRequest request, NodeClient cli
...
         }
         try {
+
+            Template template;
+            Map<String, String> useCaseDefaultsMap = Collections.emptyMap();
             XContentParser parser = request.contentParser();
             ensureExpectedToken(XContentParser.Token.START_OBJECT, parser.nextToken(), parser);
-            Template template = Template.parse(parser);
+            if (useCase != null) {
+                String json = ParseUtils.resourceToString("/" + DefaultUseCases.getSubstitutionReadyFileByUseCaseName(useCase));
```
owaiskazi19, line 126: `String json` looks weird here.

```suggestion
                String useCaseFile = ParseUtils.resourceToString("/" + DefaultUseCases.getSubstitutionReadyFileByUseCaseName(useCase));
```

src/main/java/org/opensearch/flowframework/rest/RestCreateWorkflowAction.java
```diff
@@ -112,11 +117,53 @@ protected RestChannelConsumer prepareRequest(RestRequest request, NodeClient cli
...
+                String defaultsFilePath = DefaultUseCases.getDefaultsFileByUseCaseName(useCase);
+                useCaseDefaultsMap = ParseUtils.parseJsonFileToStringToStringMap("/" + defaultsFilePath);
+
+                if (request.hasContent()) {
+                    try {
+                        Map<String, String> userDefaults = ParseUtils.parseStringToStringMap(parser);
+                        // updates the default params with anything user has given that matches
+                        for (Map.Entry<String, String> userDefaultsEntry : userDefaults.entrySet()) {
+                            if (useCaseDefaultsMap.containsKey(userDefaultsEntry.getKey())) {
```
owaiskazi19, line 135: for better readability 
```suggestion
                            if (useCaseDefaultsMap.containsKey(userDefaultsEntry.getKey())) {
                            String key = userDefaultsEntry.getKey();
                            String value = userDefaultsEntry.getValue();
                            
```

src/main/java/org/opensearch/flowframework/transport/WorkflowRequest.java
```diff
@@ -150,6 +166,22 @@ public Map<String, String> getParams() {
         return Map.copyOf(this.params);
     }
 
+    /**
+     * Gets the params map
```
owaiskazi19, line 170: and on L171
```suggestion
     * Gets the use case
```

src/main/java/org/opensearch/flowframework/util/ParseUtils.java
```diff
@@ -364,10 +373,19 @@ private static Object conditionallySubstitute(Object value, Map<String, Workflow
...
+                    String regex = "\\$\\{\\{\\s*" + Pattern.quote(e.getKey()) + "\\s*\\}\\}";
+                    String replacement = e.getValue();
+
+                    // Special handling for JSON strings that contain placeholders (connectors action)
+                    replacement = Matcher.quoteReplacement(replacement.replace("\"", "\\\""));
+
+                    // Use Pattern.compile().matcher() to avoid issues with replaceAll's direct pattern compilation
+                    Pattern pattern = Pattern.compile(regex);
+                    Matcher matcher = pattern.matcher((String) value);
+                    value = matcher.replaceAll(replacement);
```
owaiskazi19, line 387: Instead of pattern can we use StringBuilder here. Something like

```java
new StringBuilder(modifiedValue.toString().replaceAll(regex, replacement));
```

src/main/java/org/opensearch/flowframework/transport/WorkflowRequest.java
```diff
@@ -150,6 +166,22 @@ public Map<String, String> getParams() {
...
+        return this.useCase;
+    }
+
+    /**
+     * Gets the params map
+     * @return the params map
+     */
+    public Map<String, String> getDefaultParams() {
+        return Map.copyOf(this.defaultParams);
+    }
```
owaiskazi19, line 183: Let's add test for this getters in `WorkflowRequestResponseTests`

owaiskazi19 (commented):
Have few design questions and few code/style comments

src/main/resources/defaults/cohere-embedding-defaults.json
amitgalitz: I was thinking its clearer to parse the needed params only. Otherwise we would need to parse a potentially extremely long json all to Xcontent and then find the one we need

src/main/java/org/opensearch/flowframework/common/DefaultUseCases.java
amitgalitz: Are you suggesting not using the enum? Enum at least allows in compile time guarantee that users have both a use case name, a default file and a template file all registered together for that use case. relaying on a json to ensure we have both means more runtime errors.

I do agree with you though that it means they have to add things in multiple places, this however isn't final version of this and I was hoping in 2.14/15 to have an API for user to add more usecases

src/main/java/org/opensearch/flowframework/util/ParseUtils.java
amitgalitz: I was getting a bug with replaceAll that it didn't diffrentiate a more complicated substiutitin, for example when the string to be substituted in is: "{ \"texts\": ${parameters.texts}, \"truncate\": \"${parameters.truncate}\", \"model\": \"${parameters.model}\", \"input_type\": \"${parameters.input_type}\" }"

it would get confused with the other ${}, I can try this again though after I made changes after

owaiskazi19: Declare a StringBuilder like

```
StringBuilder modifiedValue = new StringBuilder(value);
```

Replace it like

```
modifiedValue = new StringBuilder(modifiedValue.toString().replaceAll(regex, replacement));
```

then your final value would be `value=modifiedValue.toString()` because you will convert StringBuilder to String here

amitgalitz: I see, i actually tested it out now and it works like this even: 
`value = ((String) value).replaceAll(regex, replacement);`

src/main/java/org/opensearch/flowframework/transport/WorkflowRequest.java
amitgalitz: added

src/main/java/org/opensearch/flowframework/util/ParseUtils.java
owaiskazi19: Didn't work with `StringBuilder`?

build.gradle
owaiskazi19: Is this added by spotless?

src/main/resources/defaults/cohere-embedding-defaults.json
owaiskazi19: We can have a key with useCase name and value as the default params for that use case in the json. That way we can just retrieve one object.

src/main/java/org/opensearch/flowframework/common/DefaultUseCases.java
owaiskazi19: Can you create an issue to address this? This can be make more cleaner. Having enum would ask user to add code here.

src/main/java/org/opensearch/flowframework/util/ParseUtils.java
amitgalitz: Works same without it too though cause anyways I want to convert in the end to a string

src/main/java/org/opensearch/flowframework/common/DefaultUseCases.java
amitgalitz: yeah this is a matter of what is easier for user to add more too and what can be more safe in terms of runtime error. I am still not against having some system index users can just hit API to add more at least on there cluster so don't need to touch code

src/main/resources/defaults/cohere-embedding-defaults.json
amitgalitz: then we need to parse that entire object into some map to get(key) op no?

build.gradle
amitgalitz: I guess so, I run spotlessApply everytime and its passing spotless check

owaiskazi19:
@dbwiddis can you take a look and we can merge this in?

src/main/resources/defaults/cohere-embedding-defaults.json
owaiskazi19: Can be addressed in a follow up PR. Can you create an issue for this?

owaiskazi19 (approved):
Please create those 2 issues to either continue the discussion over the design or have a follow up PR. Rest looks good. Thanks for addressing the comments

src/main/java/org/opensearch/flowframework/common/DefaultUseCases.java
amitgalitz: created issue https://github.com/opensearch-project/flow-framework/issues/584

src/main/resources/defaults/cohere-embedding-defaults.json
amitgalitz: https://github.com/opensearch-project/flow-framework/issues/585

dbwiddis (approved):
LGTM!  This is an awesome addition.

src/main/java/org/opensearch/flowframework/common/DefaultUseCases.java
dbwiddis: > We should provide an easy way for users to add their own default files and substitutionTemplate

Definitely for the next version.  I think we need discussion on where to store those (all in same template or separately) as either way is a one-way decision with change to system indices.

Issue #585, opened 2024-03-16
[FEATURE] Invistigate combining all default files into one file


amitgalitz:
### Is your feature request related to a problem?
After adding additional 20-30 use cases, it would be good to re-evaluate if we want all the use case defaults to be in one file that has them all in a nested object or to continue with the separate file for each default use case approach.

PR #588, opened 2024-03-18
Adding new use cases

amitgalitz:
### Description
This PR mainly adds a few new use cases, including there defaults and templates.
Additionally:
1. Changed neural sparse step required fields based on new upgrades made in 2.12 so only name, version and format is required
2. Added some extra escaping logic in substitution so i can handle substituting in complicated painless script
3. Added logic in create pipeline step to escape " of an array.

Some open questions:
1. I have quite a few templates that are very similar cause of slight changes in connectors, should I keep it that way for 2.13 or potentially have connectors get made with fields that are never used and that look like this: 
`"post_process_function": "${{create_connector.actions.post_process_function}}"`
I was thinking it might confuse customers and not be the cleanest look.

2. For some of the more nested param I want users to have control over like the mapping and processors the keys were getting pretty nested so I tried to limit them but then it might not be as consistent. Let me know if you think this can be improved. Another potential solution is to just have the key be ${{node_id.last-field-name}} or two fields back incase they are the same like in hybrid-template. 

for example now I have: 
`"${{normalization-processor.combination.technique}}"` and `"${{normalization-processor.normalization.technique}}"`.

this could be changed to `"${{create_search_pipeline.combination.technique}}"` and `"${{create_search_pipeline.normalization.technique}}"`

other fields like: `"text_embedding.field_map.input"` will change to `"create_ingest_pipeline.field_map.input"`

other option is keeping as is or having the full path in dot form

### Issues Resolved


dbwiddis (approved):

src/main/resources/defaults/multi-modal-search-defaults.json
```diff
@@ -0,0 +1,15 @@
...
+    "create_ingest_pipeline.model_id": "123",
+    "create_ingest_pipeline.embedding": "vector_embedding",
+    "text_image_embedding.field_map.text": "image_description",
+    "text_image_embedding.field_map.image": "image_binary",
+    "create_index.name": "my-multimodal-nlp-index",
+    "create_index.settings.number_of_shards": 2,
+    "text_image_embedding.field_map.output.dimension": 1024,
+    "create_index.mappings.method.engine": "lucene",
+    "create_index.mappings.method.name": "hnsw"
+}
```
owaiskazi19, line 15: This is the reason we should address https://github.com/opensearch-project/flow-framework/issues/585. Multiple files for just defaults.

build.gradle
```diff
@@ -180,7 +180,6 @@ dependencies {
 
     // ZipArchive dependencies used for integration tests
     zipArchive group: 'org.opensearch.plugin', name:'opensearch-ml-plugin', version: "${opensearch_build}"
-
```
owaiskazi19, line 183: So spotless removed it again? Weird

owaiskazi19 (approved):
Great work @amitgalitz! This feature would widen our plugin's use cases.

build.gradle
amitgalitz: this time while testing I actually added:
```
    zipArchive group: 'org.opensearch.plugin', name:'opensearch-knn', version: "${opensearch_build}"
        zipArchive group: 'org.opensearch.plugin', name:'neural-search', version: "${opensearch_build}"
```
a bunch of times, maybe in the end after I removed it just kept line deleted

src/main/resources/defaults/bedrock-titan-embedding-defaults.json
```diff
@@ -0,0 +1,17 @@
+{
+    "template.name": "deploy-bedrock-titan-embedding-model",
+    "template.description": "Deploying Amazon Bedrock Titan embedding model ",
+    "create_connector.name": "Amazon Bedrock Connector: embedding",
+    "create_connector.description": "The connector to bedrock Titan embedding model",
+    "create_connector.region": "us-east-1",
+    "create_connector.endpoint": "api.openai.com",
+    "create_connector.credential.access_key": "123",
```
joshpalis, line 8: Should we be including dummy credential values in these defaults?

joshpalis (commented):
A few questions

jackiehanyang (approved):

src/main/resources/defaults/bedrock-titan-embedding-defaults.json
amitgalitz: its actually just there cause its either going to be the "123" here or `${{create_connector.credential.key}}`

I could potentially fail for all templates with create_connector in them if no key is passed. I was actually thinking of doing this but didnt get to it. 

I would also do something similar for like semantic search template (the one without model creation), I could fail if no model ID is provided

build.gradle
owaiskazi19: @amitgalitz can you add a note on top of the template which requires the associated plugins in the OS cluster?

amitgalitz: do you mean in the use case name?

src/main/resources/defaults/cohere-chat-defaults.json
```diff
@@ -0,0 +1,14 @@
+{
+    "template.name": "deploy-cohere-chat-model",
+    "template.description": "deploying cohere chat model",
+    "create_connector.name": "Cohere Chat Model",
+    "create_connector.description": "The connector to Cohere's public chat API",
+    "create_connector.protocol": "http",
+    "create_connector.model": "command",
+    "create_connector.endpoint": "api.cohere.ai",
+    "create_connector.credential.key": "123",
```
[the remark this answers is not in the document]
amitgalitz: same reply as above

build.gradle
amitgalitz: Will add in comments and documentation

PR #600, opened 2024-03-20
Fixing create index and use case input parsing bugs 

amitgalitz:
### Description
Multiple bug fixes:

1. Address the inconsistency created by this issue: https://github.com/opensearch-project/OpenSearch/issues/12775, we want to have the same behavior as users today expect through the regular Create Index REST API. This means that we shouldn't expect _doc in the mapping and we should add it ourselves, this is exactly what core does on the rest layer but have missed doing on transport layer. https://github.com/opensearch-project/OpenSearch/blob/2.x/server/src/main/java/org/opensearch/rest/action/admin/indices/RestCreateIndexAction.java#L106

2. We know correctly expect arrays in the input fields for the default use cases, I make sure to correctly escape the strings weather we are expecting a numerical array or string array. For example:
 ```
 "[\"text\", \"hello\"]" to "["text", "hello"]"), this is needed for processors that take in string arrays,
 This also removes the quotations around the array making the array valid to consume
 (e.g. "weights": "[0.7, 0.3]" -> "weights": [0.7, 0.3]) 
```
this is done through parsing these arrays into strings in RestCreateWorkflowAction (we currently support inputs of string, array, map of strings of strings) and handleing the backlashes in createpipelinestep to correctly send the body to the transport action

3. Added conversational search template without creating a conversation after discussing missing conversation API here: https://github.com/opensearch-project/ml-commons/pull/2211 

### CODE REVIEW

File: src/main/java/org/opensearch/flowframework/util/ParseUtils.java [hunk 1]
```diff
@@ -182,6 +184,15 @@ public static Map<String, Object> parseStringToObjectMap(XContentParser parser)
             if (parser.currentToken() == XContentParser.Token.START_OBJECT) {
                 // If the current token is a START_OBJECT, parse it as Map<String, String>
                 map.put(fieldName, parseStringToStringMap(parser));
+            } else if (parser.currentToken() == XContentParser.Token.START_ARRAY) {
+                // If an array, parse it to a string
+                // Handle array: convert it to a string representation
+                List<String> elements = new ArrayList<>();
+                while (parser.nextToken() != XContentParser.Token.END_ARRAY) {
+                    elements.add("\"" + parser.text() + "\"");  // Adding escaped quotes around each element
```
jackiehanyang, line 192: how the parser handles special characters? could there be a situation we need additional escaping for some characters?

src/main/java/org/opensearch/flowframework/util/ParseUtils.java
```diff
@@ -413,4 +424,21 @@ public static Map<String, String> parseJsonFileToStringToStringMap(String path)
...
+    /**
+     * Takes an input string, then checks if there is an array in the string with backslashes around strings
+     * (e.g.  "[\"text\", \"hello\"]" to "["text", "hello"]"), this is needed for processors that take in string arrays,
+     * This also removes the quotations around the array making the array valid to consume
+     * (e.g. "weights": "[0.7, 0.3]" to "weights": [0.7, 0.3])
+     * @param input The inputString given to be transformed
+     * @return the transformed string
+     */
+    public static String removingBackslashesAndQuotesInArrayInJsonString(String input) {
+        return Pattern.compile("\"\\[(.*?)]\"").matcher(input).replaceAll(matchResult -> {
```
jackiehanyang, line 437: using regex to manipulate JSON strings can be risky due to the complexity and nested nature of JSON. Regex might not handle all cases correctly, especially with nested arrays or objects. Consider whether a JSON parsing library could more reliably achieve the goal.

src/main/java/org/opensearch/flowframework/util/ParseUtils.java
```diff
@@ -413,4 +424,21 @@ public static Map<String, String> parseJsonFileToStringToStringMap(String path)
...
+
+    /**
+     * Takes an input string, then checks if there is an array in the string with backslashes around strings
+     * (e.g.  "[\"text\", \"hello\"]" to "["text", "hello"]"), this is needed for processors that take in string arrays,
+     * This also removes the quotations around the array making the array valid to consume
+     * (e.g. "weights": "[0.7, 0.3]" to "weights": [0.7, 0.3])
+     * @param input The inputString given to be transformed
+     * @return the transformed string
+     */
+    public static String removingBackslashesAndQuotesInArrayInJsonString(String input) {
```
jackiehanyang, line 436: What will happen if the input string includes escaped characters other than quotes, such as `\\`, `\b`, `\n`, etc.

src/main/java/org/opensearch/flowframework/util/ParseUtils.java
```diff
@@ -182,6 +184,15 @@ public static Map<String, Object> parseStringToObjectMap(XContentParser parser)
...
                 // If the current token is a START_OBJECT, parse it as Map<String, String>
                 map.put(fieldName, parseStringToStringMap(parser));
+            } else if (parser.currentToken() == XContentParser.Token.START_ARRAY) {
+                // If an array, parse it to a string
+                // Handle array: convert it to a string representation
+                List<String> elements = new ArrayList<>();
+                while (parser.nextToken() != XContentParser.Token.END_ARRAY) {
+                    elements.add("\"" + parser.text() + "\"");  // Adding escaped quotes around each element
+                }
+                String arrayString = "[" + String.join(", ", elements) + "]";
```
dbwiddis, line 194: Doesn't `List.toString()` already do this?

owaiskazi19 (changes requested):
Let's wait for https://github.com/opensearch-project/ml-commons/pull/2211 to be merged before merging this

src/main/java/org/opensearch/flowframework/util/ParseUtils.java
dbwiddis, line 437: For performance, we should pre-compile the pattern into a constant.

src/main/java/org/opensearch/flowframework/rest/RestCreateWorkflowAction.java
```diff
@@ -131,18 +131,19 @@ protected RestChannelConsumer prepareRequest(RestRequest request, NodeClient cli
...
+                            String value = userDefaultsEntry.getValue().toString();
                             if (useCaseDefaultsMap.containsKey(key)) {
                                 useCaseDefaultsMap.put(key, value);
                             }
                         }
                     } catch (Exception ex) {
                         RestStatus status = ex instanceof IOException ? RestStatus.BAD_REQUEST : ExceptionsHelper.status(ex);
-                        String errorMessage = "failure parsing request body when a use case is given";
+                        String errorMessage =
+                            "failure parsing request body when a use case is given, make sure to provide a map with values that are either strings, arrays, or map of strings to strings";
```
owaiskazi19, line 146:
```suggestion
                            "failure parsing request body when a use case is given, make sure to provide a map with values that are either Strings, Arrays, or Map of Strings to Strings";
```

src/main/java/org/opensearch/flowframework/workflow/CreateIndexStep.java
```diff
@@ -85,8 +92,22 @@ public PlainActionFuture<WorkflowData> execute(
...
+            try {
+                if (!configurations.isEmpty()) {
+                    Map<String, Object> sourceAsMap = XContentHelper.convertToMap(configurationsBytes, false, MediaTypeRegistry.JSON).v2();
+                    sourceAsMap = prepareMappings(sourceAsMap);
+                    createIndexRequest.source(sourceAsMap, LoggingDeprecationHandler.INSTANCE);
+                }
+            } catch (Exception ex) {
+                String errorMessage = "Failed to create the index"
+                    + indexName
+                    + " due to incorrect mapping, make sure you aren't providing _doc in mapping";
```
dbwiddis, line 106:
```suggestion
                    + ", _doc is not permitted in mapping";
```

dbwiddis (approved):
LGTM with a few suggestions.

src/main/java/org/opensearch/flowframework/util/ParseUtils.java
dbwiddis: we probably want a more narrow regex to define that these are an array of numerics

amitgalitz:
> Let's wait for [opensearch-project/ml-commons#2211](https://github.com/opensearch-project/ml-commons/pull/2211) to be merged before merging this

Yes had a conversation with Yaliang on this, for now users will have to create there own conversation and put into the rag processor, conversation as user created isn't too bad as these are basically per user conversation and should be created for each new one. If we created a conversation here and user got it from state index it's fairly similar amount of calls for the users, Yaliang suggested we auto create this conversation idea in RAG processor if processor doesn't get one in the search request. I overall agree with that, it is probably too tight for 2.13 though.

src/main/java/org/opensearch/flowframework/util/ParseUtils.java
owaiskazi19: This compilation might fail for the cases where there might be multiple occurrences of quoted JSON arrays in the input string.
Can you try the below?
```java
Pattern pattern = Pattern.compile("\"\\[(.*?)]\"");
Matcher matcher = pattern.matcher(input);

 StringBuffer result = new StringBuffer();
 while (matcher.find()) {
               // Extract matched content and remove backslashes before quotes
            String withoutEscapes = matcher.group(1).replaceAll("\\\\\"", "\"");
            // Return the transformed string with the brackets but without the outer quotes
            matcher.appendReplacement(result, "[" + withoutEscapes + "]");
}
         // Append remaining input after the last match
    matcher.appendTail(result);
    
    return result.toString();

```

Or else use a JSON parsing library as Jackie called out.

src/main/java/org/opensearch/flowframework/workflow/CreateIndexStep.java
```diff
@@ -129,6 +150,26 @@ public PlainActionFuture<WorkflowData> execute(
         return createIndexFuture;
     }
 
+    // This method to check if the mapping contains a type `_doc` and if yes we fail the request
+    // is to duplicate the behavior we have today through create index rest API, we want users
+    // to encounter the same behavior and not suddenly have to add `_doc` while using our create_index step
```
owaiskazi19, line 155: You should link the bug issue as well here.

owaiskazi19 (commented):
We are doing so much of regex. I am worried it might fail if user has some complex mappings or payload. Is there any parsing library we can use to achieve the same?

src/main/java/org/opensearch/flowframework/util/ParseUtils.java
amitgalitz: yes this does have limitations, for example this fails on the parsing:
```
{
    "create_ingest_pipeline.model_id": "UyE4XY4BDyI7YxgR_A1v",
    "normalization-processor.combination.parameters.weights": "[0.3, 0.7, #$2_)]"
}
```

for example tho if we try to create a hybrid search processor with nested array it actaully gets cast exception on the processor code itself:
```
{
    "create_ingest_pipeline.model_id": "UyE4XY4BDyI7YxgR_A1v",
    "normalization-processor.combination.parameters.weights": "[0.3, 0.7, [0.3]]"
}
```
fails with:
```
java.lang.ClassCastException: class java.util.ArrayList cannot be cast to class java.lang.Double (java.util.ArrayList and java.lang.Double are in module java.base of loader 'bootstrap')
        at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197) ~[?:?]
        at java.base/java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1625) ~[?:?]
        at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:509) ~[?:?]
        at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499) ~[?:?]
        at java.base/java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:921) ~[?:?]
        at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234) ~[?:?]
        at java.base/java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:682) ~[?:?]
        at org.opensearch.neuralsearch.processor.combination.ScoreCombinationUtil.getWeights(ScoreCombinationUtil.java:42) ~[?:?]
        ```

**Comment by owaiskazi19** [comment]
> I overall agree with that, it is probably too tight for 2.13 though.

Can we have a follow up issue to handle it later?

File: src/main/java/org/opensearch/flowframework/util/ParseUtils.java (hunk 2 above)
Comment (amitgalitz) [reply src/main/java/org/opensearch/flowframework/util/ParseUtils.java:437]: I wasn't sure if other library methods would have other unexpected results since I basically tested it with the two main use cases of having a string array and a numerical array and while in the file its all in quotes, when we give to the create pipeline transport actions we don't want those. Do you suggest finding some backslash remover in jackson or somewhere else?

>we probably want a more narrow regex to define that these are an array of numerics

there's also the case where its not just numeric like here:     "create_search_pipeline.retrieval_augmented_generation.context_field_list": "[\"text\"]"

File: src/main/java/org/opensearch/flowframework/workflow/CreateIndexStep.java [hunk 8]
```diff
@@ -129,6 +150,26 @@ public PlainActionFuture<WorkflowData> execute(
...
+        if (source.containsKey("mappings") == false || (source.get("mappings") instanceof Map) == false) {
+            return source;
+        }
+
+        Map<String, Object> newSource = new HashMap<>(source);
+
+        @SuppressWarnings("unchecked")
+        Map<String, Object> mappings = (Map<String, Object>) source.get("mappings");
+        if (MapperService.isMappingSourceTyped(MapperService.SINGLE_MAPPING_NAME, mappings)) {
+            throw new WorkflowStepException("The mapping definition cannot be nested under a type", RestStatus.BAD_REQUEST);
```
owaiskazi19, line 166:
```suggestion
            throw new WorkflowStepException("The mapping definition cannot be nested under _doc", RestStatus.BAD_REQUEST);
```

amitgalitz: want to keep same error message as core and rest API today

src/main/java/org/opensearch/flowframework/rest/RestCreateWorkflowAction.java
amitgalitz: fixed

src/main/java/org/opensearch/flowframework/workflow/CreateIndexStep.java
amitgalitz: fixed

src/main/java/org/opensearch/flowframework/util/ParseUtils.java
amitgalitz: Additionally this is only being used in createPipeline steps where we are just limiting to what we can expect, I can also try and find a library that will solve all cases more generically

src/main/java/org/opensearch/flowframework/common/DefaultUseCases.java
```diff
@@ -93,7 +93,13 @@ public enum DefaultUseCases {
         "substitutionTemplates/semantic-search-with-model-and-query-enricher-template.json"
     ),
     /** defaults file and substitution ready template for hybrid search, no model creation*/
-    HYBRID_SEARCH("hybrid_search", "defaults/hybrid-search-defaults.json", "substitutionTemplates/hybrid-search-template.json");
+    HYBRID_SEARCH("hybrid_search", "defaults/hybrid-search-defaults.json", "substitutionTemplates/hybrid-search-template.json"),
```
ohltyler, line 96: why the name discrepancy on line 31? It doesn't end in `-template` like the others.

src/main/java/org/opensearch/flowframework/util/ParseUtils.java
amitgalitz: The parser can deal with strings like, haven't tried this within an array but also that isn't an allowed input in downstream apis : 
```
 "\n    StringBuilder parametersBuilder = new StringBuilder(\"{\");\n    if (params.text_docs.length > 0 && params.text_docs[0] != null) {\n      parametersBuilder.append(\"\\\"inputText\\\":\");\n      parametersBuilder.append(\"\\\"\");\n      parametersBuilder.append(params.text_docs[0]);\n      parametersBuilder.append(\"\\\"\");\n      \n      if (params.text_docs.length > 1 && params.text_docs[1] != null) {\n        parametersBuilder.append(\",\");\n      }\n    }\n    \n    \n    if (params.text_docs.length > 1 && params.text_docs[1] != null) {\n      parametersBuilder.append(\"\\\"inputImage\\\":\");\n      parametersBuilder.append(\"\\\"\");\n      parametersBuilder.append(params.text_docs[1]);\n      parametersBuilder.append(\"\\\"\");\n    }\n    parametersBuilder.append(\"}\");\n    \n    return  \"{\" +\"\\\"parameters\\\":\" + parametersBuilder + \"}\";",
 ```

File: src/main/resources/defaults/conversational-search-defaults.json [hunk 10]
```diff
@@ -0,0 +1,20 @@
...
+    "create_connector.model": "command",
+    "create_connector.endpoint": "api.cohere.ai",
+    "create_connector.credential.key": "123",
+    "create_connector.actions.url": "https://api.cohere.ai/v1/chat",
+    "create_connector.actions.request_body": "{ \"message\": \"${parameters.message}\", \"model\": \"${parameters.model}\" }",
+    "register_remote_model.name": "Cohere chat model",
+    "register_remote_model.description": "cohere-chat-model",
+    "create_search_pipeline.pipeline_id": "rag-pipeline",
+    "create_search_pipeline.retrieval_augmented_generation.tag": "openai_pipeline_demo",
+    "create_search_pipeline.retrieval_augmented_generation.description": "Demo pipeline Using cohere Connector",
```
ohltyler, line 16: nit - `Demo pipeline using a Cohere chat model`

src/main/resources/defaults/conversational-search-defaults.json
```diff
@@ -0,0 +1,20 @@
+{
+    "template.name": "deploy-cohere-chat-model",
+    "template.description": "deploying cohere chat model",
```
ohltyler, line 3: nit - `A template to deploy a Cohere chat model`. I think we should try to be clean with metadata / descriptions, as we will probably use this to bubble up on the frontend for future OOTB use cases.

src/main/resources/substitutionTemplates/conversational-search-with-cohere-model-template.json
```diff
@@ -0,0 +1,83 @@
...
+  "name": "${{template.name}}",
+  "description": "${{template.description}}",
+  "use_case": "SEMANTIC_SEARCH",
+  "version": {
+    "template": "1.0.0",
+    "compatibility": [
+      "2.12.0",
+      "3.0.0"
+    ]
+  },
```
ohltyler, line 11: We should make clear in the documentation these template schemas are subject to change. For example, once workflow versioning is available, all of these templates will need to be updated, and these existing ones won't be compatible.

src/main/java/org/opensearch/flowframework/util/ParseUtils.java
amitgalitz: yes a little braindead moment right there

src/main/java/org/opensearch/flowframework/rest/RestCreateWorkflowAction.java
```diff
@@ -131,18 +131,19 @@ protected RestChannelConsumer prepareRequest(RestRequest request, NodeClient cli
...
+                            String value = userDefaultsEntry.getValue().toString();
                             if (useCaseDefaultsMap.containsKey(key)) {
                                 useCaseDefaultsMap.put(key, value);
                             }
                         }
                     } catch (Exception ex) {
                         RestStatus status = ex instanceof IOException ? RestStatus.BAD_REQUEST : ExceptionsHelper.status(ex);
-                        String errorMessage = "failure parsing request body when a use case is given";
+                        String errorMessage =
+                            "failure parsing request body when a use case is given, make sure to provide a map with values that are either Strings, Arrays, or Map of Strings to Strings";
```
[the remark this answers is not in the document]
amitgalitz: we also can do arrays of numbers

src/main/resources/defaults/conversational-search-defaults.json
amitgalitz: makes sense, do you think if we mention defaults might change in other releases its okay to change these later without worrying on bwc between templates? At least for simple things like if we want to change descriptions later

ohltyler: Sure, but I think we should at least clean up the metadata/description stuff now. Quick minor changes but looks much more presentable. Adding details etc. we can change later, as we'll for sure already need to make schema changes.

src/main/java/org/opensearch/flowframework/util/ParseUtils.java
amitgalitz: tbh this was a good call out cause, I am not sure if my fix somewhere else broke taking arrays of numbers but now code will be: 
```else if (parser.currentToken() == XContentParser.Token.START_ARRAY) {
                // If an array, parse it to a string
                // Handle array: convert it to a string representation
                List<String> elements = new ArrayList<>();
                while (parser.nextToken() != XContentParser.Token.END_ARRAY) {
                    if (parser.currentToken().equals(XContentParser.Token.VALUE_NUMBER)) {
                        elements.add(String.valueOf(parser.numberValue()));  // If number value don't add escaping quotes
                    } else {
                        elements.add("\"" + parser.text() + "\"");  // Adding escaped quotes around each element
                    }
                }
                map.put(fieldName, elements.toString());
            }
            ```

File: src/main/resources/defaults/conversational-search-defaults.json (hunk 11 above)
Comment (amitgalitz) [reply src/main/resources/defaults/conversational-search-defaults.json]: Yup, didn't mean instead of now, just made me though of if we need to do more in future, did some more clean up now

File: src/main/java/org/opensearch/flowframework/util/ParseUtils.java (hunk 1 above)
Comment (amitgalitz) [reply src/main/java/org/opensearch/flowframework/util/ParseUtils.java]: @jackiehanyang @owaiskazi19 lmk what you think of above tested it with different inputs and all templates

File: src/main/resources/substitutionTemplates/conversational-search-with-cohere-model-template.json (hunk 12 above)
Comment (amitgalitz) [reply src/main/resources/substitutionTemplates/conversational-search-with-cohere-model-template.json:11]: will add this to documentation

**Review by joshpalis — APPROVED** [review]

**Comment by amitgalitz** [comment]
@owaiskazi19  I changed method to what you suggested for ParseUtils.removingBackslashesAndQuotesInArrayInJsonString, I also made slight change to ParseUtils.parseStringToObjectMap() in latest commit

File: src/test/java/org/opensearch/flowframework/rest/FlowFrameworkRestApiIT.java [hunk 14]
```diff
@@ -465,7 +467,21 @@ public void testDefaultSemanticSearchUseCaseWithFailureExpected() throws Excepti
...
+            .collect(Collectors.toSet());
+
+        for (String useCaseName : allUseCaseNames) {
+            Response response = createWorkflowWithUseCase(client(), useCaseName);
+            assertEquals(RestStatus.CREATED, TestHelpers.restStatus(response));
+
+            Map<String, Object> responseMap = entityAsMap(response);
+            String workflowId = (String) responseMap.get(WORKFLOW_ID);
+            getAndAssertWorkflowStatus(client(), workflowId, State.NOT_STARTED, ProvisioningProgress.NOT_STARTED);
+        }
```
ohltyler, line 485: Nice! Can you also create a tracking issue for looking into e2e integ / provisioning tests for some of these templates, if there's not one already?

ohltyler (approved):

src/main/java/org/opensearch/flowframework/util/ParseUtils.java
jackiehanyang: I'm still not very confident in manually wrapping `"\"` around elements for escape purpose. For example, it will cause incorrect escape result for strings start with quotes and control-chars like backslash. But it could a rare edge case, I'm fine with revising this in a separate pr/issue.

jackiehanyang (approved):

src/test/java/org/opensearch/flowframework/rest/FlowFrameworkRestApiIT.java
amitgalitz: https://github.com/opensearch-project/flow-framework/issues/601

src/main/java/org/opensearch/flowframework/util/ParseUtils.java
amitgalitz: Will create an issue for this

src/main/java/org/opensearch/flowframework/util/ParseUtils.java
amitgalitz: changed to suggestion Owais gave

owaiskazi19:
@amitgalitz we need to forward port this to main

PR #604, opened 2024-03-22
Fixing default semantic use case integration test

joshpalis:
### Description
Addresses failed distribution build test : https://github.com/opensearch-project/flow-framework/issues/599#issuecomment-2014020214

### DIFF
--- a/src/test/java/org/opensearch/flowframework/FlowFrameworkRestTestCase.java
+++ b/src/test/java/org/opensearch/flowframework/FlowFrameworkRestTestCase.java
@@ -33,6 +33,8 @@ import org.opensearch.client.RestClientBuilder;
 import org.opensearch.common.settings.Settings;
 import org.opensearch.common.unit.TimeValue;
 import org.opensearch.common.util.concurrent.ThreadContext;
+import org.opensearch.common.xcontent.LoggingDeprecationHandler;
+import org.opensearch.common.xcontent.json.JsonXContent;
 import org.opensearch.commons.rest.SecureRestClientBuilder;
 import org.opensearch.core.rest.RestStatus;
 import org.opensearch.core.xcontent.DeprecationHandler;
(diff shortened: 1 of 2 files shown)


src/test/java/org/opensearch/flowframework/FlowFrameworkRestTestCase.java
```diff
@@ -706,4 +708,22 @@ protected GetPipelineResponse getPipelines() throws IOException {
...
+            "",
+            ImmutableList.of(new BasicHeader(HttpHeaders.USER_AGENT, ""))
+        );
+        List<Object> pluginsList = JsonXContent.jsonXContent.createParser(
+            NamedXContentRegistry.EMPTY,
+            LoggingDeprecationHandler.INSTANCE,
+            response.getEntity().getContent()
+        ).list();
+        return pluginsList.stream().map(o -> ((Map<String, Object>) o).get("component").toString()).collect(Collectors.toList());
+    }
```
owaiskazi19, line 728: We can use the same method [here](https://github.com/opensearch-project/flow-framework/blob/main/src/test/java/org/opensearch/flowframework/FlowFrameworkPluginIT.java#L23) as well

owaiskazi19:
One question. How the test is failing for distribution build though? Since it contains all the plugins already.

joshpalis:
> One question. How the test is failing for distribution build though? Since it contains all the plugins already.

So the distribution build is failing since all plugin integration tests run on the same cluster, this particular integration test of ours assumes the neural search plugin is not in our cluster since our we dont have that plugin dependency in our build.gradle.

owaiskazi19:
> So the distribution build is failing since all plugin integration tests run on the same cluster, this particular integration test of ours assumes the neural search plugin is not in our cluster since our we dont have that plugin dependency in our build.gradle.

How did this test pass on GHA then?

joshpalis:
> How did this test pass on GHA then?

When we run integration tests on our CI, the cluster only has ML Commons and Flow Framework

owaiskazi19:
> When we run integration tests on our CI, the cluster only has ML Commons and Flow Framework

I see it was a failure test case that's why no issue on our GHA

owaiskazi19 (approved):

amitgalitz (approved):

joshpalis:
will raise manual PRs for the backports since this modifies the FlowFrameworkRestTestCase file

PR #619, opened 2024-03-26
[POC] Adding Orchestrate API and Search Response Processor Step

joshpalis:
### Description
Adds a new `/_plugins/_flow_framework/<workflow_id>/_orchestrate` API which accepts a map of string to string as the request body content. 

Example payload looks like : 
```
POST _plugins/_flow_framework/<workflow_id>/_orchestrate 
{
    "model_id" : "uuihhqw0f9i90932r1h0a9w0f"
}
```

The transport action retrieves this along with the workflow ID to fetch a previously created `search` use case template from the global context index. The process nodes are then created, passing in the request body content map for substitution and then executes the graph. 

Upon attempting to complete each plain action future, we track the final returned search response to return to the user

### DIFF
--- a/src/main/java/org/opensearch/flowframework/workflow/SearchResponseProcessorStep.java
+++ b/src/main/java/org/opensearch/flowframework/workflow/SearchResponseProcessorStep.java
@@ -39,6 +39,9 @@ public class SearchResponseProcessorStep implements WorkflowStep {
 
     private static final Logger logger = LogManager.getLogger(SearchResponseProcessorStep.class);
     private final SearchPipelineService searchPipelineService;
+    /**
+     * The name of the SearchResponseProcessor step
+     */
     public static final String NAME = "search_response_processor";
 
     /**
--- a/src/main/java/org/opensearch/flowframework/workflow/SearchResponseProcessorStep.java
+++ b/src/main/java/org/opensearch/flowframework/workflow/SearchResponseProcessorStep.java
@@ -1,0 +1,108 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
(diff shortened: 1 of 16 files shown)


joshpalis:
Tests will fail since since the getters for the request and response processor factories are package private : 
```
> Task :compileJava
/__w/flow-framework/flow-framework/src/main/java/org/opensearch/flowframework/workflow/SearchResponseProcessorStep.java:85: error: getResponseProcessorFactories() is not public in SearchPipelineService; cannot be accessed from outside package
                .getResponseProcessorFactories();
                ^
```

Modifying this [line](https://github.com/opensearch-project/OpenSearch/blob/52b27f47bca5b3ab52cab237542f32c307d203b4/server/src/main/java/org/opensearch/search/pipeline/SearchPipelineService.java#L419) in the `SearchPipelineService` so that the getter is public and publishing to maven local will allow the tests to pass locally

src/main/java/org/opensearch/flowframework/workflow/SearchResponseProcessorStep.java
```diff
@@ -0,0 +1,115 @@
...
+import static org.opensearch.flowframework.common.CommonValue.PROCESSOR_CONFIG;
+import static org.opensearch.flowframework.common.CommonValue.SEARCH_REQUEST;
+import static org.opensearch.flowframework.common.CommonValue.SEARCH_RESPONSE;
+import static org.opensearch.flowframework.common.CommonValue.TAG;
+import static org.opensearch.flowframework.common.CommonValue.TYPE;
+
+/**
+ * Step to create a search response processor
+ */
+public class SearchResponseProcessorStep implements WorkflowStep {
```
owaiskazi19, line 38: Can you name it as ReRankProcessorStep for better clarity?

joshpalis: Its a general step to create any search response processor, the type of processor created depends on the processor config passed to it

owaiskazi19 (approved):
LGTM

PR #635, opened 2024-04-02
Reset workflow state to initial state after successful deprovision

### LINKED ISSUE #578: [BUG] Deprovision doesn't remove the error field if it exists
### What is the bug?
If provisioning has a failure on one of the steps we write to the error field in the state index. However when we deprovision that same workflow we don't remove the error message from the document in the workflow state index.
1. Provision with an expected failure (easy way is to provision a pipeline with a processor that isn't in your cluster)
2.Deprovision the same workflow:

resulting state index:

```
            {
                "_index": ".plugins-flow-framework-state",
                "_id": "QljrPY4BJDcxHyOJPUNd",
                "_score": 1.0,
                "_source": {
                    "workflow_id": "QljrPY4BJDcxHyOJPUNd",
                    "state": "NOT_STARTED",
                    "provisioning_progress": "NOT_STARTED",
                    "provision_start_time": 1710435961112,
                    "resources_created": [],
                    "provision_end_time": 1710435828966,
                    "error": "org.opensearch.flowframework.exception.WorkflowStepException during step create_ingest_pipeline, restStatus: BAD_REQUEST"
                }
            }
            
```

error field shouldn't be there anymore
### Potential solution

Some potential solution
1. We can either simply place the error field as an empty string, however this isn't prefered because if the error field wasn't there to begin with now we have the field with just `"error": ""`. 
2. You can't directly remove the field through update, this has to be done through a script as far as I know. You also can't provide both a contentMap(doc) and a script in the same update request. This means we either have to just update the logic here https://github.com/opensearch-project/flow-framework/blob/main/src/main/java/org/opensearch/flowframework/transport/DeprovisionWorkflowTransportAction.java#L230 to update with a script that removes the error field if it exists (e.g `"if (ctx._source.contains(error)) { ctx._source.remove('error)}"`) and updates the other fields accordingly or we have to make two update operations, one with the current way we do it with a new doc and another with the painless script

dbwiddis:
### Description

On successful deprovision, resets the workflow state to its initial state, rather than simply updating the fields.

### Issues Resolved

Fixes #578 

### DIFF
--- a/src/main/java/org/opensearch/flowframework/transport/DeprovisionWorkflowTransportAction.java
+++ b/src/main/java/org/opensearch/flowframework/transport/DeprovisionWorkflowTransportAction.java
@@ -44,11 +44,12 @@ import java.util.stream.Collectors;
 
 import static org.opensearch.flowframework.common.CommonValue.DEPROVISION_WORKFLOW_THREAD_POOL;
 import static org.opensearch.flowframework.common.CommonValue.PROVISIONING_PROGRESS_FIELD;
-import static org.opensearch.flowframework.common.CommonValue.PROVISION_START_TIME_FIELD;
+import static org.opensearch.flowframework.common.CommonValue.PROVISION_END_TIME_FIELD;
 import static org.opensearch.flowframework.common.CommonValue.RESOURCES_CREATED_FIELD;
 import static org.opensearch.flowframework.common.CommonValue.STATE_FIELD;
 import static org.opensearch.flowframework.common.WorkflowResources.getDeprovisionStepByWorkflowStep;
 import static org.opensearch.flowframework.common.WorkflowResources.getResourceByWorkflowStep;
+import static org.opensearch.flowframework.util.ParseUtils.getUserContext;
 
 /**
  * Transport Action to deprovision a workflow from a stored use case template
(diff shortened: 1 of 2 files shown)


joshpalis (approved):

ohltyler (approved):

src/main/java/org/opensearch/flowframework/transport/DeprovisionWorkflowTransportAction.java
```diff
@@ -246,7 +242,7 @@ private void updateWorkflowState(
                 Map.ofEntries(
                     Map.entry(STATE_FIELD, State.COMPLETED),
                     Map.entry(PROVISIONING_PROGRESS_FIELD, ProvisioningProgress.DONE),
-                    Map.entry(PROVISION_START_TIME_FIELD, Instant.now().toEpochMilli()),
+                    Map.entry(PROVISION_END_TIME_FIELD, Instant.now().toEpochMilli()),
```
amitgalitz, line 245: is there a new provision end time when we fail deprovisioning? Is the end time here not when the provisioning itself ends (last step is executed), unrelated to deprovisioning?

Issue #140, opened 2023-11-01
[FEATURE] Move UI Metadata from State Index to Global Context Index


owaiskazi19:
### Is your feature request related to a problem?
We moved `UI_METADATA_FIELD` to state index recently [here](https://github.com/opensearch-project/opensearch-ai-flow-framework/pull/110/files#diff-d49816ce76f874e242a545dd3de0515c2435ec7a6a315faf83ae28e4069f08e0R321). The Search APIs as a part of https://github.com/opensearch-project/opensearch-ai-flow-framework/issues/124 searches on GC index. For frontend to utilize the search APIs, the `UI_METADATA_FIELD` should be moved to GC index back.

### What solution would you like?
1. Move `UI_METADATA_FIELD` field to GC index.
2. See if the request came from dashboard based on user-agents [here](https://github.com/opensearch-project/opensearch-ai-flow-framework/pull/139/files#diff-60d4924f5a70806d00988d81fdcd27d050a1d09ecb925cfa9b4449bb7bd224eeR35) and include `UI_METADATA_FIELD`.
3. Else, exclude `UI_METADATA_FIELD`.

### What alternatives have you considered?
_A clear and concise description of any alternative solutions or features you've considered._

### Do you have any additional context?
_Add any other context or screenshots about the feature request here._

joshpalis:
There's already an issue for this : https://github.com/opensearch-project/opensearch-ai-flow-framework/issues/135

owaiskazi19:
> There's already an issue for this : #135

Ah! I'll close #135

PR #636, opened 2024-04-02
Moves ui-metadata field from state to template index

joshpalis:
### Description
Moves the ui-metadata field from the state to template index mapping

### Issues Resolved
https://github.com/opensearch-project/flow-framework/issues/140

### DIFF
--- a/src/main/resources/mappings/global-context.json
+++ b/src/main/resources/mappings/global-context.json
@@ -1,7 +1,7 @@
 {
   "dynamic": false,
   "_meta": {
-    "schema_version": 2
+    "schema_version": 3
   },
   "properties": {
     "workflow_id": {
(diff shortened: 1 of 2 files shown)


owaiskazi19 (approved):

joshpalis:
@owaiskazi19 should I also increment the schema versions of both mappings? This field is not searchable, only used by the front end, so not sure if we would need to

owaiskazi19:
> @owaiskazi19 should I also increment the schema versions of both mappings? This field is not searchable, only used by the front end, so not sure if we would need to

This change will be a part of 2.14 after GA in 2.13, so yes we should increment it.

ohltyler (approved):

dbwiddis:
Should we add this to the BWC tests? What happens if we read a template from a 2.13 index (that has this field) in 2.14?

PR #639, opened 2024-04-02
Silently ignore content on APIs that don't require it

### LINKED ISSUE #638: [BUG] The RestController always consumes content when handling real HTTP requests
### What is the bug?

The `BaseRestHandler` includes a check whether content (the REST body) has been consumed by an API, throwing an exception [here](https://github.com/opensearch-project/OpenSearch/blob/c25b00c6b62fd6766999c09fd06dd1f7b281b3eb/server/src/main/java/org/opensearch/rest/BaseRestHandler.java#L125-L127) if it does not do so.   This exception is tested in OpenSearch [in a few places](https://github.com/search?q=repo%3Aopensearch-project%2FOpenSearch%20%22does%20not%20support%20having%20a%20body%22&type=code) and we relied on it [here](https://github.com/search?q=repo%3Aopensearch-project%2Fflow-framework+%22having+a+body%22&type=code) for multiple APIs:
 - Get Workflow
 - Delete Workflow
 - Get Workflow State
 - Deprovision Workflow
 - Get Workflow Step

The reason for this check was that including a body would still execute the requested action but then fail to provide a proper REST result to the user.  For the Get APIs this was mostly harmless, but deleting or deprovisioning a workflow would have side effects not reported to the user.

Unfortunately, it appears that these unit tests which rely on the behavior of the `BaseRestHandler` do not take into account the fact that the `RestController` always consumes content [here](https://github.com/opensearch-project/OpenSearch/blob/2dc071ff2546b0c6483b9e0a382badf9ac353cdc/server/src/main/java/org/opensearch/rest/RestController.java#L296) when a REST request arrives via the HTTP handler.

Query any of the above APIs with a body in the request, for example
```json
localhost:9200/_plugins/_flow_framework/workflow/any_id_here
{}
```
The result is:
```json
{
    "error": {
        "root_cause": [
            {
                "type": "illegal_state_exception",
                "reason": "Channel is already closed"
            }
        ],
        "type": "illegal_state_exception",
        "reason": "Channel is already closed"
    },
    "status": 500
}
```

As intended: a 400 response with the message "request [GET /_plugins/_flow_framework/workflow/{workflow_id}] does not support having a body"

As implemented elsewhere in OpenSearch:  silently ignored

This behavior has existed for at least 5 years but is undocumented/confusing. I'll be submitting a bug on the OpenSearch side reporting this, but the ideal fix (not consuming content) would potentially be seriously backwards incompatble with many scripts developed in the past 5 years.   

In the meantime, we should not be sending `null` back to the Base Rest Handler.

dbwiddis:
### Description

Despite code paths and unit tests showing otherwise, OpenSearch silently ignores the content (body) of a REST request that doesn't require it.  We should do the same.

### Issues Resolved

Fixes #638 (or at least doesn't give users a poor experience).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,9 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 ### Features
 ### Enhancements
 ### Bug Fixes
+### Bug Fixes
+- Silently ignore content on APIs that don't require it ([#639](https://github.com/opensearch-project/flow-framework/pull/639))
+
 ### Infrastructure
 ### Documentation
 ### Maintenance
--- a/src/main/java/org/opensearch/flowframework/rest/RestDeleteWorkflowAction.java
+++ b/src/main/java/org/opensearch/flowframework/rest/RestDeleteWorkflowAction.java
@@ -69,11 +69,11 @@ public class RestDeleteWorkflowAction extends BaseRestHandler {
                     RestStatus.FORBIDDEN
                 );
             }
-            // Validate content
-            if (request.hasContent()) {
-                // BaseRestHandler will give appropriate error message
-                return channel -> channel.sendResponse(null);
-            }
+
+            // Always consume content to silently ignore it
+            // https://github.com/opensearch-project/flow-framework/issues/578
+            request.content();
+
             // Validate params
             if (workflowId == null) {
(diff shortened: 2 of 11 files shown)


joshpalis (approved):

owaiskazi19:
We should not backport it 2.13 as the release date is today and there will be no new RC generation

owaiskazi19 (approved):

Issue #655, opened 2024-04-09
[FEATURE] Add validation for required parameters for default use cases


amitgalitz:
### Is your feature request related to a problem?

Currently all default use cases can run as is without giving any arguments. However by the nature of how some of the use cases work, while the resources can be created they wont be useable without giving a credential or for example a model ID if we are relaying on one in an ingest processor

### What solution would you like?
We should add another field in this [enum](https://github.com/opensearch-project/flow-framework/blob/main/src/main/java/org/opensearch/flowframework/common/DefaultUseCases.java) to specify which field is a must have, currently most of the workflows required either an API key, AWS credentials or a model ID. We can then check if the required param was overwritten by the user [here](https://github.com/opensearch-project/flow-framework/blob/main/src/main/java/org/opensearch/flowframework/rest/RestCreateWorkflowAction.java#L130)

### What alternatives have you considered?
Not having this validation as is right now because technically provision doesn't fail, however I think for a better user experience we also want to push users to have a resources that will be fully working 

### Do you have any additional context?
_Add any other context or screenshots about the feature request here._

PR #658, opened 2024-04-10
Adding guardrails to default use case params

joshpalis:
### Description
Adds a requiredParams field to the DefaultUseCases enum which is used validate the given request body of default use case request

Examples :
```
curl -i -XPOST "http://localhost:9200/_plugins/_flow_framework/workflow?use_case=semantic_search_with_cohere_embedding_query_enricher&provision=true" -H "Content-Type:application/json" -d '{"test_value" : "QFaXLdVUFfk7XYxqSOtweCvVxXS2l3Jsw1DeKhuR"}'
HTTP/1.1 400 Bad Request
content-length: 157

{"error":"Missing the following required parameters for use case [semantic_search_with_cohere_embedding_query_enricher] : [create_connector.credential.key]"}

curl -i -XPOST "http://localhost:9200/_plugins/_flow_framework/workflow?use_case=multimodal_search_with_bedrock_titan&provision=true&?pretty" -H "Content-Type:application/json" -d '{"test_value" : "QFaXLdVUFfk7XYxqSOtweCvVxXS2l3Jsw1DeKhuR"}'
HTTP/1.1 400 Bad Request
content-length: 231

{"error":"Missing the following required parameters for use case [multimodal_search_with_bedrock_titan] : [create_connector.credential.access_key, create_connector.credential.secret_key, create_connector.credential.session_token]"}
```

### Issues Resolved
https://github.com/opensearch-project/flow-framework/issues/655

### DIFF
--- a/src/main/java/org/opensearch/flowframework/common/DefaultUseCases.java
+++ b/src/main/java/org/opensearch/flowframework/common/DefaultUseCases.java
@@ -219,9 +219,8 @@ public enum DefaultUseCases {
      * Gets the required parameters based on the given use case
      * @param useCaseName name of the given use case
      * @return the list of required params
-     * @throws FlowFrameworkException if the use case doesn't exist in enum
      */
-    public static List<String> getRequiredParamsByUseCaseName(String useCaseName) throws FlowFrameworkException {
+    public static List<String> getRequiredParamsByUseCaseName(String useCaseName) {
         if (useCaseName != null && !useCaseName.isEmpty()) {
             for (DefaultUseCases useCase : values()) {
                 if (useCase.getUseCaseName().equals(useCaseName)) {
@@ -229,7 +228,7 @@ public enum DefaultUseCases {
                 }
             }
         }
-        logger.error("Unable to find required parameters for use case: {}", useCaseName);
-        throw new FlowFrameworkException("Unable to find required parameters for use case: " + useCaseName, RestStatus.BAD_REQUEST);
+        logger.error("Default use case [" + useCaseName + "] does not exist");
+        throw new FlowFrameworkException("Default use case [" + useCaseName + "] does not exist", RestStatus.BAD_REQUEST);
     }
 }
--- a/src/main/java/org/opensearch/flowframework/common/DefaultUseCases.java
+++ b/src/main/java/org/opensearch/flowframework/common/DefaultUseCases.java
@@ -215,6 +215,12 @@ public enum DefaultUseCases {
         throw new FlowFrameworkException("Unable to find substitution ready file for use case: " + useCaseName, RestStatus.BAD_REQUEST);
     }
 
+    /**
+     * Gets the required parameters based on the given use case
+     * @param useCaseName name of the given use case
+     * @return the list of required params
+     * @throws FlowFrameworkException if the use case doesn't exist in enum
+     */
     public static List<String> getRequiredParamsByUseCaseName(String useCaseName) throws FlowFrameworkException {
         if (useCaseName != null && !useCaseName.isEmpty()) {
             for (DefaultUseCases useCase : values()) {
--- a/src/main/java/org/opensearch/flowframework/common/DefaultUseCases.java
+++ b/src/main/java/org/opensearch/flowframework/common/DefaultUseCases.java
@@ -13,6 +13,16 @@ import org.apache.logging.log4j.Logger;
 import org.opensearch.core.rest.RestStatus;
 import org.opensearch.flowframework.exception.FlowFrameworkException;
 
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+import static org.opensearch.flowframework.common.CommonValue.CREATE_CONNECTOR_CREDENTIAL_ACCESS_KEY;
+import static org.opensearch.flowframework.common.CommonValue.CREATE_CONNECTOR_CREDENTIAL_KEY;
+import static org.opensearch.flowframework.common.CommonValue.CREATE_CONNECTOR_CREDENTIAL_SECRET_KEY;
+import static org.opensearch.flowframework.common.CommonValue.CREATE_CONNECTOR_CREDENTIAL_SESSION_TOKEN;
+import static org.opensearch.flowframework.common.CommonValue.CREATE_INGEST_PIPELINE_MODEL_ID;
+
 /**
  * Enum encapsulating the different default use cases and templates we have stored
  */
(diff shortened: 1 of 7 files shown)


src/main/java/org/opensearch/flowframework/rest/RestCreateWorkflowAction.java
```diff
@@ -132,6 +133,18 @@ protected RestChannelConsumer prepareRequest(RestRequest request, NodeClient cli
                         XContentParser parser = request.contentParser();
                         ensureExpectedToken(XContentParser.Token.START_OBJECT, parser.nextToken(), parser);
                         Map<String, Object> userDefaults = ParseUtils.parseStringToObjectMap(parser);
+
```
amitgalitz, line 136: might be my fault here since I pointed to wrong point in the code, but what if request has no content at all, we should still send this error message back right?

joshpalis: Good point, I can move the validation logic around to cover both cases. There are some default use cases that have no required parameters

src/main/java/org/opensearch/flowframework/common/DefaultUseCases.java
```diff
@@ -22,94 +32,119 @@ public enum DefaultUseCases {
...
+        "substitutionTemplates/neural-sparse-local-biencoder-template.json",
+        Collections.emptyList()
     ),
     /** defaults file and substitution ready template for semantic search, no model creation*/
-    SEMANTIC_SEARCH("semantic_search", "defaults/semantic-search-defaults.json", "substitutionTemplates/semantic-search-template.json"),
+    SEMANTIC_SEARCH(
+        "semantic_search",
+        "defaults/semantic-search-defaults.json",
+        "substitutionTemplates/semantic-search-template.json",
+        List.of(CREATE_INGEST_PIPELINE_MODEL_ID)
```
owaiskazi19, line 85: Not a part of this PR but can you create an issue to improve `semantic-search-template` to include creating model as well?

src/main/java/org/opensearch/flowframework/common/DefaultUseCases.java
```diff
@@ -22,94 +32,119 @@ public enum DefaultUseCases {
...
+    ),
     /** defaults file and substitution ready template for multimodal search, no model creation*/
     MULTI_MODAL_SEARCH(
         "multimodal_search",
         "defaults/multi-modal-search-defaults.json",
-        "substitutionTemplates/multi-modal-search-template.json"
+        "substitutionTemplates/multi-modal-search-template.json",
```
owaiskazi19, line 91: And this too

src/main/java/org/opensearch/flowframework/rest/RestCreateWorkflowAction.java
```diff
@@ -126,12 +127,29 @@ protected RestChannelConsumer prepareRequest(RestRequest request, NodeClient cli
                 );
                 String defaultsFilePath = DefaultUseCases.getDefaultsFileByUseCaseName(useCase);
                 useCaseDefaultsMap = ParseUtils.parseJsonFileToStringToStringMap("/" + defaultsFilePath);
-
-                if (request.hasContent()) {
+                List<String> requiredParams = DefaultUseCases.getRequiredParamsByUseCaseName(useCase);
+
+                if (request.hasContent() == false && requiredParams.size() != 0) {
```
owaiskazi19, line 132: This would fail for the templates which doesn't have required params.

src/main/java/org/opensearch/flowframework/rest/RestCreateWorkflowAction.java
```diff
@@ -126,12 +127,29 @@ protected RestChannelConsumer prepareRequest(RestRequest request, NodeClient cli
...
+                        // Validate user defaults key set
+                        Set<String> userDefaultKeys = userDefaults.keySet();
+                        if (!userDefaultKeys.containsAll(requiredParams)) {
+                            requiredParams.removeAll(userDefaultKeys);
+                            throw new FlowFrameworkException(
+                                "Missing the following required parameters for use case [" + useCase + "] : " + requiredParams.toString(),
+                                RestStatus.BAD_REQUEST
+                            );
+                        }
+
```
owaiskazi19, line 152: Rather removing from `requiredParams` List. We can do

```java
Set<String> userDefaultKeys = userDefaults.keySet();
Set<String> missingParams = new HashSet<>(requiredParams);
missingParams.removeAll(userDefaultKeys);

if (!missingParams.isEmpty()) {
    throw new FlowFrameworkException(
        "Missing the following required parameters for use case [" + useCase + "] : " + missingParams,
        RestStatus.BAD_REQUEST
    );
}

```

owaiskazi19 (commented):
LG overall with few comments

src/main/java/org/opensearch/flowframework/rest/RestCreateWorkflowAction.java
joshpalis: Made a modification, tested with local bi-encoder which requires no parameters : 
```
curl -i -XPOST "http://localhost:9200/_plugins/_flow_framework/workflow?use_case=local_neural_sparse_search_bi_encoder&provision=true"
HTTP/1.1 201 Created
content-length: 38

{"workflow_id":"CIbTzo4BDyZpM8DAIpBz"}%                                                                                                                                              

```

src/main/java/org/opensearch/flowframework/rest/RestCreateWorkflowAction.java
joshpalis: `DefaultUseCases.getRequiredParamsByUseCaseName()` just returns a copy of the required param list, so we're not modifying the original list.

src/main/java/org/opensearch/flowframework/common/DefaultUseCases.java
```diff
@@ -171,4 +214,22 @@ public static String getSubstitutionReadyFileByUseCaseName(String useCaseName) t
...
+    public static List<String> getRequiredParamsByUseCaseName(String useCaseName) throws FlowFrameworkException {
+        if (useCaseName != null && !useCaseName.isEmpty()) {
+            for (DefaultUseCases useCase : values()) {
+                if (useCase.getUseCaseName().equals(useCaseName)) {
+                    return new ArrayList<String>(useCase.getRequiredParams());
+                }
+            }
+        }
+        logger.error("Unable to find required parameters for use case: {}", useCaseName);
+        throw new FlowFrameworkException("Unable to find required parameters for use case: " + useCaseName, RestStatus.BAD_REQUEST);
```
dbwiddis, line 233: This may be accurate but doesn't match the root cause of the error which is a bad use case, which is what the user needs to correct with the BAD_REQUEST.

src/main/java/org/opensearch/flowframework/rest/RestCreateWorkflowAction.java
```diff
@@ -126,12 +127,31 @@ protected RestChannelConsumer prepareRequest(RestRequest request, NodeClient cli
                 );
                 String defaultsFilePath = DefaultUseCases.getDefaultsFileByUseCaseName(useCase);
                 useCaseDefaultsMap = ParseUtils.parseJsonFileToStringToStringMap("/" + defaultsFilePath);
-
-                if (request.hasContent()) {
+                List<String> requiredParams = DefaultUseCases.getRequiredParamsByUseCaseName(useCase);
+
+                if (request.hasContent() == false) {
+                    if (requiredParams.size() != 0) {
```
dbwiddis, line 133: Combine these with an &&.  Also `size != 0` is better `!requiredParams.isEmpty()`.

dbwiddis, line 133: Combine these with an &&.  Also:
- don't check `== false`.  Just do `!request.hasContent()`.
- `size != 0` is better as `!requiredParams.isEmpty()`.

src/main/java/org/opensearch/flowframework/rest/RestCreateWorkflowAction.java
```diff
@@ -126,12 +127,31 @@ protected RestChannelConsumer prepareRequest(RestRequest request, NodeClient cli
...
+                } else {
                     try {
                         XContentParser parser = request.contentParser();
                         ensureExpectedToken(XContentParser.Token.START_OBJECT, parser.nextToken(), parser);
                         Map<String, Object> userDefaults = ParseUtils.parseStringToObjectMap(parser);
+
+                        // Validate user defaults key set
+                        Set<String> userDefaultKeys = userDefaults.keySet();
+                        if (!userDefaultKeys.containsAll(requiredParams)) {
+                            requiredParams.removeAll(userDefaultKeys);
```
dbwiddis, line 148: I see @owaiskazi19 has made a comment about removing.  I'm generally in favor of this approach but I haven't dug in to check whether the list you're returning is mutable.   Just looking at this and the other branch of the if/else it seems the whole thing can be simplified a lot.

dbwiddis, line 148: I see @owaiskazi19 has made a comment about removing here.  I'm generally in favor of this approach but I haven't dug in to check whether the list you're returning is mutable.   But  Just looking at this and the other branch of the if/else it seems the whole thing can be simplified.  Or not.  If it works we can simplify later.

src/main/java/org/opensearch/flowframework/rest/RestCreateWorkflowAction.java
```diff
@@ -141,13 +161,16 @@ protected RestChannelConsumer prepareRequest(RestRequest request, NodeClient cli
                             }
                         }
                     } catch (Exception ex) {
-                        RestStatus status = ex instanceof IOException ? RestStatus.BAD_REQUEST : ExceptionsHelper.status(ex);
-                        String errorMessage =
-                            "failure parsing request body when a use case is given, make sure to provide a map with values that are either Strings, Arrays, or Map of Strings to Strings";
-                        logger.error(errorMessage, ex);
-                        throw new FlowFrameworkException(errorMessage, status);
+                        if (ex instanceof FlowFrameworkException) {
+                            throw (FlowFrameworkException) ex;
```
dbwiddis, line 165: I'm curious if the casting actually makes any difference here?

src/test/java/org/opensearch/flowframework/FlowFrameworkRestTestCase.java
```diff
@@ -345,16 +345,24 @@ protected Response createWorkflow(RestClient client, Template template) throws E
...
+     * @param the required params
      * @throws Exception if the request fails
      * @return a rest response
      */
-    protected Response createWorkflowWithUseCase(RestClient client, String useCase) throws Exception {
+    protected Response createWorkflowWithUseCase(RestClient client, String useCase, List<String> params) throws Exception {
+
+        StringBuilder sb = new StringBuilder();
+        for (String param : params) {
+            sb.append("\"" + param + "\" : \"\"").append(",");
```
dbwiddis, line 356: Mixing StringBuilder append and string concatenation seems wrong somehow.   And appending characters is cleaner than appending escaped strings.   Can we do `.append('"').append(param)` and so on?

src/test/java/org/opensearch/flowframework/FlowFrameworkRestTestCase.java
```diff
@@ -345,16 +345,24 @@ protected Response createWorkflow(RestClient client, Template template) throws E
...
+        }
+        if (params.size() != 0) sb.deleteCharAt(sb.length() - 1);
```
dbwiddis, line 358: Prefer `isEmpty()` here, and I'm surprised spotless doesn't complain about lack of curly braces here!

dbwiddis (commented):
LGTM generally.  Lots of nits for your consideration. I'm traveling and unlikely to review again promptly so please don't wait on me to review again.

src/main/java/org/opensearch/flowframework/common/DefaultUseCases.java
joshpalis: Sure I'll improve the error message

src/main/java/org/opensearch/flowframework/rest/RestCreateWorkflowAction.java
joshpalis: The returned list is mutable. `DefaultUseCases.getRequiredParamsByUseCaseName()` returns a copy of the list

src/main/java/org/opensearch/flowframework/rest/RestCreateWorkflowAction.java
joshpalis: If we combine these, then we will error out if a user attempts to use a use-case that requires no parameters, such as the `local_neural_sparse_search_bi_encoder`

joshpalis: answered above

src/main/java/org/opensearch/flowframework/common/DefaultUseCases.java
owaiskazi19: @joshpalis is this being tracked anywhere? I see the comment is resolved

joshpalis: we have a `semantic-search-with-model` template which includes model registration

owaiskazi19 (approved):
LGTM

dbwiddis (approved):

Issue #617, opened 2024-03-26
[FEATURE] Enable easy migration from BM25 to Neural Index with Reindex Step


dbwiddis:
### Is your feature request related to a problem?

One of the most common use cases for ML offerings is enabling Neural Search by adding embeddings to an existing index.  The steps to set this up are simple, [documented here](https://opensearch.org/docs/latest/im-plugin/reindex-data/#transform-documents-during-reindexing), and [here](https://opensearch.org/docs/latest/ingest-pipelines/processors/text-embedding/) and @owaiskazi19 demonstrated them early on in our exploratory development with [a scrappy demo](https://github.com/owaiskazi19/Low-Code-Platform/blob/1078222d119c5c3f95a88b25cd0319dcc4d559d8/backend/main.py#L73-L151)
 - Configure an Ingest Pipeline to add the text embeddings to an index
 - Create a new index using the pipeline prepared in the previous step
 - Reindex an existing text (BM25) index onto the index created in the previous step.  

### What solution would you like?

The creation of an ingest pipeline and new index configuration have already been completed and will be part of the 2.13 release:
 - https://github.com/opensearch-project/flow-framework/pull/558
 - https://github.com/opensearch-project/flow-framework/pull/574

To complete this solution we need to add a `ReindexStep` that [calls the Reindex API](https://opensearch.org/docs/latest/api-reference/document-apis/reindex/).

Reindexing does have some cautions that a user should be aware of, and hiding these cautions behind an automated workflow risks surprising users with some behavior.

It will only reindex documents which were in the original index at the start of the operation, so if an index is still being written to, the reindex won't capture new documents. Also, it's an expensive operation on large indices, with this note in the linked docs:
> Reindexing can be an expensive operation depending on the size of your source index. We recommend you disable replicas in your destination index by setting number_of_replicas to 0 and re-enable them once the reindex process is complete.

Using a model for embeddings adds even more "expense" to this process.

Accordingly, there should be at least some sort of "confirmation prompt" when this workflow step is used.  From a backend/template perspective, a path parameter expressing the user's acknowledgement of these cautions (e.g., `allow_expensive=true` or similar) should be required, with the provisioning step failed-fast with a helpful/verbose error message if there is a reindex step present and the parameter is not set true.

### What alternatives have you considered?

[...]

PR #660, opened 2024-04-12
Added Workflow Step for ReIndex

owaiskazi19:
### Description
Added Workflow Step for ReIndex. This workflow step currently support just parameters source and destination index. 
Reindex does support other lots of params like [here](https://opensearch.org/docs/1.2/opensearch/rest-api/document-apis/reindex/#request-body).

More details about the indices, template and response can be found [here](https://github.com/opensearch-project/flow-framework/issues/617#issuecomment-2052670258).

Keeping this in draft mode.

### Issues Resolved
https://github.com/opensearch-project/flow-framework/issues/617

### DIFF
--- a/src/main/java/org/opensearch/flowframework/common/CommonValue.java
+++ b/src/main/java/org/opensearch/flowframework/common/CommonValue.java
@@ -94,28 +94,14 @@ public class CommonValue {
     public static final String SUCCESS = "success";
     /** Type field */
     public static final String TYPE = "type";
-    /** default_mapping_option filed */
-    public static final String DEFAULT_MAPPING_OPTION = "default_mapping_option";
     /** ID Field */
     public static final String ID = "id";
-    /** Processors field */
-    public static final String PROCESSORS = "processors";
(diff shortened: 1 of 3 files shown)


dbwiddis:
@YANG-DB what parameters do you need for your observability use case?

YANG-DB:
We need the ability to parameterize for the create index example:
- index name 
- index alias name 
- index settings 
-----------------------------------------
In the case of creating a mapping template example:
- template name 
- components list

navneet1v:
@owaiskazi19 , @minalsha One suggestion I have here is we should add the parameters from re-index api that controls the speed of re-indexing. As per my understanding the parameter that controls is `requests_per_second`. Reason we should add this in because this step will be used to re-index BM25 data to vectors via ML model. Most of ML models have a limitation on how many request they can handle per sec. So to give better user experience we should have this field atleast.

owaiskazi19:
Closing in favor of https://github.com/opensearch-project/flow-framework/pull/718.

PR #671, opened 2024-04-19
Allow strings for boolean workflow step parameters

### LINKED ISSUE #669: [BUG] Handling of Booleans should be robust to strings
### What is the bug?

Using `"true"` in some workflow steps when a boolean `true` is expected causes a 500 error (ClassCastException).

Add a template to register a local model with `"deploy": "true"`. 

The template parsing interprets the String `"true"` as Boolean.TRUE.

These four instance of casting `(Boolean) inputs.get(...)` should be changed to use `Boolean.parseBoolean(inputs.get(...).toString())`:

https://github.com/search?q=repo%3Aopensearch-project%2Fflow-framework%20%22(Boolean)%20inputs.get%22&type=code

Alternately consider opensearch common `Booleans` class which throws an error for anything other than "true" or "false".

Unit tests should be added for both types of params.

dbwiddis:
### Description
Allows `"true"` or `"false"` strings when boolean parameters are expected.

### Issues Resolved

Fixes #669 

### DIFF
--- a/src/main/java/org/opensearch/flowframework/workflow/RegisterRemoteModelStep.java
+++ b/src/main/java/org/opensearch/flowframework/workflow/RegisterRemoteModelStep.java
@@ -13,7 +13,9 @@ import org.apache.logging.log4j.Logger;
 import org.opensearch.ExceptionsHelper;
 import org.opensearch.action.support.PlainActionFuture;
 import org.opensearch.action.update.UpdateResponse;
+import org.opensearch.common.Booleans;
 import org.opensearch.core.action.ActionListener;
+import org.opensearch.core.rest.RestStatus;
 import org.opensearch.flowframework.exception.FlowFrameworkException;
 import org.opensearch.flowframework.exception.WorkflowStepException;
 import org.opensearch.flowframework.indices.FlowFrameworkIndicesHandler;
@@ -90,7 +92,7 @@ public class RegisterRemoteModelStep implements WorkflowStep {
             String description = (String) inputs.get(DESCRIPTION_FIELD);
             String connectorId = (String) inputs.get(CONNECTOR_ID);
             Guardrails guardRails = (Guardrails) inputs.get(GUARDRAILS_FIELD);
-            final Boolean deploy = (Boolean) inputs.get(DEPLOY_FIELD);
+            final Boolean deploy = inputs.containsKey(DEPLOY_FIELD) ? Booleans.parseBoolean(inputs.get(DEPLOY_FIELD).toString()) : null;
 
             MLRegisterModelInputBuilder builder = MLRegisterModelInput.builder()
                 .functionName(FunctionName.REMOTE)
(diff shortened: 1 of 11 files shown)


src/main/java/org/opensearch/flowframework/workflow/RegisterRemoteModelStep.java
```diff
@@ -90,7 +92,7 @@ public PlainActionFuture<WorkflowData> execute(
             String description = (String) inputs.get(DESCRIPTION_FIELD);
             String connectorId = (String) inputs.get(CONNECTOR_ID);
             Guardrails guardRails = (Guardrails) inputs.get(GUARDRAILS_FIELD);
-            final Boolean deploy = (Boolean) inputs.get(DEPLOY_FIELD);
+            final Boolean deploy = inputs.containsKey(DEPLOY_FIELD) ? Booleans.parseBoolean(inputs.get(DEPLOY_FIELD).toString()) : null;
```
owaiskazi19, line 95: Why a special check just for the boolean type?

Shouldn't it be false?
```suggestion
            final Boolean deploy = inputs.containsKey(DEPLOY_FIELD) ? Booleans.parseBoolean(inputs.get(DEPLOY_FIELD).toString()) : false;
```

or simply
```suggestion
            final Boolean deploy = Booleans.parseBoolean(inputs.get(DEPLOY_FIELD).toString());
```

dbwiddis: > Shouldn't it be false?

Existing behavior is to cast the result of `Map.get()`.  If the value isn't present, the result is `(Boolean) null` not `Boolean.FALSE`.  In theory we could hardcode the "default" but then we'd have to make sure that whichever way we are paring the boolean, we match the default.  False here, true for some of the other fixes in this PR.

dbwiddis: `Booleans.parseBoolean(inputs.get(DEPLOY_FIELD).toString())` will fail with an NPE if the key is not present.  We'd need to use `getOrDefault()`.  I did consider that, but that does require us to know what the default is (see above comment) and from a DRY principle I'd prefer to keep the result the same.... null.

src/test/java/org/opensearch/flowframework/workflow/RegisterModelGroupStepTests.java
```diff
@@ -41,9 +42,14 @@
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.verify;
 
-public class ModelGroupStepTests extends OpenSearchTestCase {
+public class RegisterModelGroupStepTests extends OpenSearchTestCase {
```
owaiskazi19, line 45: :+1:

owaiskazi19 (approved):
LGTM

amitgalitz (approved):

Issue #670, opened 2024-04-18
[BUG] Some error messages are being hidden during provisioning 


amitgalitz:
### What is the bug?
Our error message that we write to the state index while provisioning don't have enough info.

For example if the user has a typo in the processor name this is the error we write
```
"org.opensearch.flowframework.exception.WorkflowStepException during step create_ingest_pipeline, restStatus: BAD_REQUEST" 
```
we use WorkflowStepException to let us know the step itself is the reason for the error, but we don't propagate outside the logs the fact that the issue is that we don't have text_embedding processor available in the cluster.

Today if a user hits the create ingest pipeline API directly they get more information without having to look at the logs:
```
{
    "error": {
        "root_cause": [
            {
                "type": "parse_exception",
                "reason": "No processor type exists with name [text_embeddinssg]",
                "processor_type": "text_embeddinssg"
            }
        ],
        "type": "parse_exception",
        "reason": "No processor type exists with name [text_embeddinssg]",
        "processor_type": "text_embeddinssg"
    },
    "status": 400
}
```

We should potentially write the exception itself too to the state index, however there can be times where the exception exposes too much information

dbwiddis:
> however there can be times where the exception exposes too much information

This is the reason we restricted it so much.  How can we programmatically decide what's safe to pass on?

dbwiddis:
The above exception is an `OpenSearchException`. These are designed to be returned to users. As a first step we should probably allow these exception messages to be passed through (along with their RestStatus)

owaiskazi19:
One of the hacky ways to handle this before writing exceptions is to check which exceptions can not be safely exposed (such as those from RemoteTransportException) and, while adding the rest of the exceptions separately and include them in the error message.

dbwiddis:
Looking at our implemented steps, we can probably just include:
 - `OpenSearchStatusException` (common in ML Commons)
 - `OpenSearchParseException` (a few in ML Commons)
 - `OpenSearchException` with a `getCause()` of `OpenSearchParseException` (Search and Ingest Pipeline configurations)

Haven't looked at any exceptions thrown in Create Index or Reindex, but that would likely complete it.

PR #676, opened 2024-04-24
Throw the correct error message in status API for WorkflowSteps

owaiskazi19:
### Description
Throw the correct error message in status API for WorkflowSteps. As called out in the issue, exceptions added in `getException` method are safe exceptions to be shown in status API.

Few examples:
```
{
    "workflow_id": "y_KgDY8BnN13dc10B7Lw",
    "error": "Missing credential during step create_claude_connector, restStatus: BAD_REQUEST",
    "state": "FAILED"
}
```

```
{
    "workflow_id": "MrGqDY8BlF2RpBCvbloe",
    "error": "No processor type exists with name [text_embeddinsg] during step create_ingest_pipeline, restStatus: BAD_REQUEST",
    "state": "FAILED"
}
```

### Issues Resolved
Resolves https://github.com/opensearch-project/flow-framework/issues/670

### DIFF
--- a/src/main/java/org/opensearch/flowframework/transport/ProvisionWorkflowTransportAction.java
+++ b/src/main/java/org/opensearch/flowframework/transport/ProvisionWorkflowTransportAction.java
@@ -266,7 +266,7 @@ public class ProvisionWorkflowTransportAction extends HandledTransportAction<Wor
                 status = ExceptionsHelper.status(ex);
             }
             logger.error("Provisioning failed for workflow {} during step {}.", workflowId, currentStepId, ex);
-            String errorMessage = (ex.getCause() == null ? ex.getClass().getName() : ex.getCause().getClass().getName())
+            String errorMessage = (ex.getCause() == null ? ex.getMessage() : ex.getCause().getClass().getName())
                 + " during step "
                 + currentStepId
                 + ", restStatus: "
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 - Reset workflow state to initial state after successful deprovision ([#635](https://github.com/opensearch-project/flow-framework/pull/635))
 - Silently ignore content on APIs that don't require it ([#639](https://github.com/opensearch-project/flow-framework/pull/639))
 - Hide user and credential field from search response ([#680](https://github.com/opensearch-project/flow-framework/pull/680))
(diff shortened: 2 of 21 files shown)


src/main/java/org/opensearch/flowframework/transport/ProvisionWorkflowTransportAction.java
```diff
@@ -266,7 +266,7 @@ private void executeWorkflow(List<ProcessNode> workflowSequence, String workflow
                 status = ExceptionsHelper.status(ex);
             }
             logger.error("Provisioning failed for workflow {} during step {}.", workflowId, currentStepId, ex);
-            String errorMessage = (ex.getCause() == null ? ex.getClass().getName() : ex.getCause().getClass().getName())
+            String errorMessage = (ex.getCause() == null ? ex.getMessage() : ex.getCause().getClass().getName())
```
dbwiddis, line 269: I don't see how we're testing the "safe" exception here?

amitgalitz:
Can you test case of how exception looks like when there is unsafe exception like remotetransportexception, when you deploy multiple models at the same time. 

Also can we also add this to create index step, along with changing CreateIndexStep line here to workflowstepexception https://github.com/opensearch-project/flow-framework/blob/main/src/main/java/org/opensearch/flowframework/workflow/CreateIndexStep.java#L142

owaiskazi19:
> Can you test case of how exception looks like when there is unsafe exception like remotetransportexception, when you deploy multiple models at the same time.

Added 2 tests for CreateIndex and RegisterRemoteStep

> Also can we also add this to create index step, along with changing CreateIndexStep line here to workflowstepexception https://github.com/opensearch-project/flow-framework/blob/main/src/main/java/org/opensearch/flowframework/workflow/CreateIndexStep.java#L142

Done

src/main/java/org/opensearch/flowframework/transport/ProvisionWorkflowTransportAction.java
owaiskazi19: For any step we are checking if the exception is safe if not then have a custom message for that step. Either way both the messages should be safe to expose.

```java
                   Exception e = getSafeException(ex);
                    String errorMessage = (e == null ? "Failed to register remote model" : e.getMessage());
                    registerRemoteModelFuture.onFailure(new WorkflowStepException(errorMessage, ExceptionsHelper.status(e)));
```

amitgalitz: Would it be safer to have safe exception check again here in case a step was missed?

amitgalitz (approved):

dbwiddis (approved):
LGTM!

PR #695, opened 2024-05-01
Fixing model group parsing and restoring context

amitgalitz:
### Description
Adding restore.context() in multiple place it isn't restored after stashing.
Fixing model group access_mode parsing, this was always failing but never tested with private on security enabled domain

### Issues Resolved

### DIFF
--- a/src/main/java/org/opensearch/flowframework/transport/CreateWorkflowTransportAction.java
+++ b/src/main/java/org/opensearch/flowframework/transport/CreateWorkflowTransportAction.java
@@ -311,6 +311,7 @@ public class CreateWorkflowTransportAction extends HandledTransportAction<Workfl
             try (ThreadContext.StoredContext context = client.threadPool().getThreadContext().stashContext()) {
                 logger.info("Querying existing workflows to count the max");
                 client.search(searchRequest, ActionListener.wrap(searchResponse -> {
+                    context.restore();
                     internalListener.onResponse(searchResponse.getHits().getTotalHits().value < maxWorkflow);
                 }, exception -> {
                     String errorMessage = "Unable to fetch the workflows";
--- a/src/main/java/org/opensearch/flowframework/util/EncryptorUtils.java
+++ b/src/main/java/org/opensearch/flowframework/util/EncryptorUtils.java
@@ -250,7 +250,6 @@ public class EncryptorUtils {
 
             GetRequest getRequest = new GetRequest(CONFIG_INDEX).id(MASTER_KEY);
             client.get(getRequest, ActionListener.wrap(getResponse -> {
-
                 if (!getResponse.isExists()) {
 
                     // Generate new key and index
@@ -260,6 +259,7 @@ public class EncryptorUtils {
                         .setRefreshPolicy(WriteRequest.RefreshPolicy.IMMEDIATE);
 
                     client.index(masterKeyIndexRequest, ActionListener.wrap(indexResponse -> {
+                        context.restore();
                         // Set generated key to master
                         logger.info("Config has been initialized successfully");
                         this.masterKey = generatedKey;
@@ -270,6 +270,7 @@ public class EncryptorUtils {
                     }));
 
                 } else {
+                    context.restore();
                     // Set existing key to master
                     logger.info("Config has already been initialized");
                     this.masterKey = (String) getResponse.getSourceAsMap().get(MASTER_KEY);
@@ -300,6 +301,7 @@ public class EncryptorUtils {
             try (ThreadContext.StoredContext context = client.threadPool().getThreadContext().stashContext()) {
                 GetRequest getRequest = new GetRequest(CONFIG_INDEX).id(MASTER_KEY);
                 client.get(getRequest, ActionListener.wrap(response -> {
+                    context.restore();
                     if (response.isExists()) {
                         this.masterKey = (String) response.getSourceAsMap().get(MASTER_KEY);
                     } else {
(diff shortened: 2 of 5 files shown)


amitgalitz:
looking at failing tests

src/main/java/org/opensearch/flowframework/workflow/RegisterModelGroupStep.java
```diff
@@ -142,7 +143,7 @@ public void onFailure(Exception ex) {
             String modelGroupName = (String) inputs.get(NAME_FIELD);
             String description = (String) inputs.get(DESCRIPTION_FIELD);
             List<String> backendRoles = getBackendRoles(inputs);
-            AccessMode modelAccessMode = (AccessMode) inputs.get(MODEL_ACCESS_MODE);
+            AccessMode modelAccessMode = AccessMode.from(((String) inputs.get(MODEL_ACCESS_MODE)).toLowerCase(Locale.ROOT));
```
dbwiddis, line 146: Since `MODEL_ACCESS_MODE` is optional, there's a chance the `get()` result is `null` here.  Casting a null Object to a String, it's still null.

AccessMode.from(null) appears to return null, which would throw an NPE on trying to lowercase it.

src/main/java/org/opensearch/flowframework/transport/CreateWorkflowTransportAction.java
```diff
@@ -311,6 +311,7 @@ void checkMaxWorkflows(TimeValue requestTimeOut, Integer maxWorkflow, ActionList
             try (ThreadContext.StoredContext context = client.threadPool().getThreadContext().stashContext()) {
                 logger.info("Querying existing workflows to count the max");
                 client.search(searchRequest, ActionListener.wrap(searchResponse -> {
```
dbwiddis, line 313: (Personal preference) While we're mucking around in this bit of code, it'd be nice to change the name of the `searchRequest` to `countRequest` to more clearly indicate we've set `size(0)`.

dbwiddis (commented):
The context restores LGTM.

Potential NPE you need to address, though.

dbwiddis:
> looking at failing tests

What is the type of the (`Object`) `inputs.get(MODEL_ACCESS_MODE)` ?  If that's already an enum then you can't cast it to a string.

src/main/java/org/opensearch/flowframework/util/EncryptorUtils.java
```diff
@@ -270,6 +270,7 @@ public void initializeMasterKey(ActionListener<Boolean> listener) {
                     }));
 
                 } else {
+                    context.restore();
```
owaiskazi19, line 273: Can have this after if/else block.

src/main/java/org/opensearch/flowframework/transport/CreateWorkflowTransportAction.java
amitgalitz: is `searchRequestHitCount` okay since its not a countRequest which is something different so also can add confusion https://opensearch.org/docs/latest/api-reference/count/, https://github.com/opensearch-project/OpenSearch/blob/main/client/rest-high-level/src/main/java/org/opensearch/client/core/CountRequest.java

src/main/java/org/opensearch/flowframework/util/EncryptorUtils.java
amitgalitz: this doesn't work as we want to restore context before listener.onResponse happens here. Tested with this and we got some exceptions

src/main/java/org/opensearch/flowframework/workflow/RegisterModelGroupStep.java
owaiskazi19: Wouldn't ml-commons would treat `AccessMode` as public when passed null? @amitgalitz

amitgalitz: its public if security isn't enabled and private when not enabled so we should pass null when its null and when given a value we should pass correct lower case string

owaiskazi19 (approved):

src/main/java/org/opensearch/flowframework/transport/CreateWorkflowTransportAction.java
dbwiddis: Let's leave it for now. I've seen `countRequest` elsewhere in the repo. RHLC actually has a CountRequest which implements it this way under the hood.  Just a future thought.

src/main/java/org/opensearch/flowframework/workflow/RegisterModelGroupStep.java
```diff
@@ -142,7 +143,10 @@ public void onFailure(Exception ex) {
             String modelGroupName = (String) inputs.get(NAME_FIELD);
             String description = (String) inputs.get(DESCRIPTION_FIELD);
             List<String> backendRoles = getBackendRoles(inputs);
-            AccessMode modelAccessMode = (AccessMode) inputs.get(MODEL_ACCESS_MODE);
+            AccessMode modelAccessMode = null;
+            if (inputs.get(MODEL_ACCESS_MODE) != null) {
```
dbwiddis, line 147: Can we make this `inputs.containsKey()`?

amitgalitz: changed

PR #705, opened 2024-05-04
Add user mapping to Workflow State index

### LINKED ISSUE #700: [BUG] The Workflow State mapping doesn't include the user field
### What is the bug?

The [Workflow State index mapping](https://github.com/opensearch-project/flow-framework/blob/main/src/main/resources/mappings/workflow-state.json) does not include a `user` field, just `user_outputs`.

Attempt to query the existence of the "user" field using any of the clients, e.g., with the Go client:
```go
userExistsQuery := NewNestedQuery("user", NewExistsQuery("user.name"))
```
Result:
```
Error 400 (Bad Request): all shards failed [type=search_phase_execution_exception]
```

Index mapping matches Java object XContent mapping.

This is not user-facing but is relevant to queries that an admin may make.

dbwiddis:
### Description

Adds the missing mapping for the `user` field.  (Copy/paste from same field in the template mapping.)

### Issues Resolved

Fixes #700 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,27 +12,13 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 ### Maintenance
 ### Refactoring
 
-## [Unreleased 2.x](https://github.com/opensearch-project/flow-framework/compare/2.13...2.x)
+## [Unreleased 2.x](https://github.com/opensearch-project/flow-framework/compare/2.14...2.x)
 ### Features
 ### Enhancements
-- Add guardrails to default use case params ([#658](https://github.com/opensearch-project/flow-framework/pull/658))
-- Allow strings for boolean workflow step parameters ([#671](https://github.com/opensearch-project/flow-framework/pull/671))
-- Add optional delay parameter to no-op step ([#674](https://github.com/opensearch-project/flow-framework/pull/674))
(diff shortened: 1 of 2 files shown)


amitgalitz:
So for this case the bwc could be a little different. Since we are only changing the mapping here I think where this would have an effect is in some sort of rollover, where we are creating the index again based of the new mapping. Would that be the only case here also in your opinion?

dbwiddis:
> So for this case the bwc could be a little different. Since we are only changing the mapping here

I don't think we're changing the mapping. The `source` stored in the index will be identical. We're just documenting the mapping in the index metadata to permit clients to directly act on the source fields rather than having to parse it first.

> I think where this would have an effect is in some sort of rollover

This is a system index, so we're not rolling it over.  The concern here would be an index that already exists with the previous mapping; we would update the mapping with the new one which our code does if the versions differ (which reminds me, I should have incremented the mapping version, so that's a needed change here).

> Would that be the only case here also in your opinion?

The only case I see this being relevant is an upgrade from 2.12-2.14 to 2.15 or later, when the index metadata would be updated.

amitgalitz (approved):
LGTM, lets add a TODO or issue to have bwc with security enabled added and general increase of coverage with security

dbwiddis:
> LGTM, lets add a TODO or issue to have bwc with security enabled added and general increase of coverage with security

https://github.com/opensearch-project/flow-framework/issues/710

minalsha (approved):

PR #718, opened 2024-05-20
Added workflow step for ReIndex STep

owaiskazi19:
### Description
Added Workflow Step for ReIndex. This workflow step currently support source and destination index. And also params like `refresh`, `requests_per_second`, `require_alias`, `slices`, `max_docs`..

More details about the indices, template and response can be found https://github.com/opensearch-project/flow-framework/issues/617#issuecomment-2052670258.

Keeping this in draft mode.

### Issues Resolved
https://github.com/opensearch-project/flow-framework/issues/617


src/main/java/org/opensearch/flowframework/workflow/ReIndexStep.java
```diff
@@ -0,0 +1,175 @@
...
+                                currentNodeInputs.getWorkflowId(),
+                                currentNodeId,
+                                getName(),
+                                destinationIndex,
+                                ActionListener.wrap(response -> {
+                                    logger.info("successfully updated resource created in state index: {}", response.getIndex());
+
+                                    reIndexFuture.onResponse(
+                                        new WorkflowData(
+                                            Map.of(RE_INDEX_FIELD, Map.of(sourceIndices, destinationIndex)),
```
joshpalis, line 130: Why do we need to return the source and destination indices as workflow data?

src/main/java/org/opensearch/flowframework/workflow/ReIndexStep.java
```diff
@@ -0,0 +1,175 @@
...
+                reindexRequest.setSlices(slices);
+            }
+
+            ActionListener<BulkByScrollResponse> actionListener = new ActionListener<>() {
+
+                @Override
+                public void onResponse(BulkByScrollResponse bulkByScrollResponse) {
+                    logger.info("Reindex from source: {} to destination {}", sourceIndices, destinationIndex);
+                    try {
+                        if (bulkByScrollResponse.getBulkFailures().isEmpty() && bulkByScrollResponse.getSearchFailures().isEmpty()) {
```
joshpalis, line 119: In the case in which bulk/search failures are not empty, should we treat that as a failure of the ReindexWorkflowStep?

src/main/java/org/opensearch/flowframework/workflow/ReIndexStep.java
```diff
@@ -0,0 +1,175 @@
...
+                            flowFrameworkIndicesHandler.updateResourceInStateIndex(
```
joshpalis, line 120: Not sure if updating the state index is necessary here since it doesn't create a resource. Re-indexing happens on an existing source and destination index

joshpalis (commented):
Initial pass

src/main/java/org/opensearch/flowframework/workflow/ReIndexStep.java
```diff
@@ -0,0 +1,175 @@
...
+                optionalKeys,
+                currentNodeInputs,
+                outputs,
+                previousNodeInputs,
+                params
+            );
+
+            String sourceIndices = (String) inputs.get(SOURCE_INDEX);
+            String destinationIndex = (String) inputs.get(DESTINATION_INDEX);
+            Boolean refresh = inputs.containsKey(REFRESH) ? Booleans.parseBoolean(inputs.get(REFRESH).toString()) : null;
```
dbwiddis, line 88: (Suggestion) We have the pattern `inputs.containsKey(foo) ? Booleans.paresBoolean(inputs.get(foo).toString()) : null` in quite a few steps.  It probably belongs in a util method somewhere.  Writing this should be a new `good first issue` issue.

src/main/java/org/opensearch/flowframework/workflow/ReIndexStep.java
```diff
@@ -0,0 +1,175 @@
...
+            Integer requestsPerSecond = (Integer) inputs.get(REQUESTS_PER_SECOND);
```
dbwiddis, line 89: Should this be a `Float`?  I think `Float.POSITIVE_INFINITY` is a valid value.

src/main/java/org/opensearch/flowframework/workflow/ReIndexStep.java
```diff
@@ -0,0 +1,175 @@
...
+            Boolean requireAlias = inputs.containsKey(REQUIRE_ALIAS) ? Booleans.parseBoolean(inputs.get(REQUIRE_ALIAS).toString()) : null;
+            Integer slices = (Integer) inputs.get(SLICES);
+            Integer maxDocs = (Integer) inputs.get(MAX_DOCS);
+
+            ReindexRequest reindexRequest = new ReindexRequest();
+            reindexRequest.setSourceIndices(sourceIndices);
+            reindexRequest.setDestIndex(destinationIndex);
```
dbwiddis, line 96:
```suggestion
            ReindexRequest reindexRequest = new ReindexRequest()
                .setSourceIndices(sourceIndices)
                .setDestIndex(destinationIndex);
```

src/main/java/org/opensearch/flowframework/workflow/ReIndexStep.java
```diff
@@ -0,0 +1,175 @@
...
+        String currentNodeId,
+        WorkflowData currentNodeInputs,
+        Map<String, WorkflowData> outputs,
+        Map<String, String> previousNodeInputs,
+        Map<String, String> params
+    ) {
+
+        PlainActionFuture<WorkflowData> reIndexFuture = PlainActionFuture.newFuture();
+
+        Set<String> requiredKeys = Set.of(SOURCE_INDEX, DESTINATION_INDEX);
```
dbwiddis, line 72: Should this be `SOURCE_INDICES`?  The `ReindexRequest` supports this and you use `sourceIndices` as the variable to store this value.

src/main/java/org/opensearch/flowframework/workflow/ReIndexStep.java
```diff
@@ -0,0 +1,175 @@
...
+            Map<String, Object> inputs = ParseUtils.getInputsFromPreviousSteps(
+                requiredKeys,
+                optionalKeys,
+                currentNodeInputs,
+                outputs,
+                previousNodeInputs,
+                params
+            );
+
+            String sourceIndices = (String) inputs.get(SOURCE_INDEX);
```
dbwiddis, line 86: Should this be a `String[]`?

src/main/java/org/opensearch/flowframework/workflow/ReIndexStep.java
dbwiddis: I expect we do (probably) need the destination index here, but not sure about source. Don't think it hurts to include them both, though.

src/main/java/org/opensearch/flowframework/workflow/ReIndexStep.java
```diff
@@ -0,0 +1,175 @@
...
+
+                                    reIndexFuture.onResponse(
+                                        new WorkflowData(
+                                            Map.of(RE_INDEX_FIELD, Map.of(sourceIndices, destinationIndex)),
+                                            currentNodeInputs.getWorkflowId(),
+                                            currentNodeInputs.getNodeId()
+                                        )
+                                    );
+                                }, exception -> {
+                                    String errorMessage = "Failed to update new reindexed"
```
dbwiddis, line 136: need white space before concatenating the next string.  Consider using square brackets to delimit the inserted strings.

src/main/java/org/opensearch/flowframework/workflow/ReIndexStep.java
owaiskazi19: The list of indices is under the same string.

src/main/java/org/opensearch/flowframework/workflow/ReIndexStep.java
owaiskazi19: Good point. Pushed the change

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 ### Enhancements
 ### Bug Fixes
 - Add user mapping to Workflow State index ([#705](https://github.com/opensearch-project/flow-framework/pull/705))
+- Add Workflow Step for Reindex from source index to destination ([#718](https://github.com/opensearch-project/flow-framework/pull/718))
```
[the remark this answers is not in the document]
owaiskazi19: Eh! Enhancement, missed this

src/main/java/org/opensearch/flowframework/common/WorkflowResources.java
```diff
@@ -58,6 +59,8 @@ public enum WorkflowResources {
     CREATE_SEARCH_PIPELINE(CreateSearchPipelineStep.NAME, WorkflowResources.PIPELINE_ID, null), // TODO delete step
     /** Workflow steps for creating an index and associated created resource */
     CREATE_INDEX(CreateIndexStep.NAME, WorkflowResources.INDEX_NAME, NoOpStep.NAME),
+    /** Workflow steps for reindex a source index to destination index and associated created resource */
+    RE_INDEX(ReIndexStep.NAME, CommonValue.DESTINATION_INDEX, NoOpStep.NAME),
```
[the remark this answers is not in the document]
owaiskazi19: Since we support just one field for now, kept destination index as the resource created

src/main/java/org/opensearch/flowframework/workflow/ReIndexStep.java
owaiskazi19: I tried to change it as 
```java
  Float requestsPerSecond = (Float) inputs.get(REQUESTS_PER_SECOND);
```

but it's failing with the ClassCastException. Do you have any idea for the same?

src/main/java/org/opensearch/flowframework/workflow/ReIndexStep.java
owaiskazi19: https://github.com/opensearch-project/flow-framework/issues/720

src/main/java/org/opensearch/flowframework/workflow/ReindexStep.java
```diff
@@ -0,0 +1,182 @@
...
+                currentNodeInputs,
+                outputs,
+                previousNodeInputs,
+                params
+            );
+
+            String sourceIndices = (String) inputs.get(SOURCE_INDICES);
+            String destinationIndex = (String) inputs.get(DESTINATION_INDEX);
+            Boolean refresh = inputs.containsKey(REFRESH) ? Booleans.parseBoolean(inputs.get(REFRESH).toString()) : null;
+            Integer requestsPerSecond = (Integer) inputs.get(REQUESTS_PER_SECOND);
```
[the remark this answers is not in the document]
owaiskazi19: Pushed the change

dbwiddis (approved):

src/main/java/org/opensearch/flowframework/workflow/ReIndexStep.java
dbwiddis: `String ...` varargs is syntactic sugar for `String[]`.  And the internal parameter that gets set is a `String[]`.  Passing a single String in this line as you do here would turn it into a 1-element String array.  

I already approved if we're only allowing reindexing one index, but this seems something we should handle.

src/main/java/org/opensearch/flowframework/common/WorkflowResources.java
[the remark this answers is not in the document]
dbwiddis: If it's a resource created, shouldn't it be the `WorkflowResources.INDEX_NAME` constant here?

src/main/java/org/opensearch/flowframework/workflow/ReIndexStep.java
dbwiddis: Looks pretty easy to support a comma-delimited list of indices with [`Strings.splitStringByCommaToArray((String) value)`](https://github.com/opensearch-project/OpenSearch/blob/2e497436cf9dd78aa3d06f86285aef897c962cac/server/src/main/java/org/opensearch/index/reindex/ReindexRequest.java#L395-L409)

src/main/java/org/opensearch/flowframework/workflow/ReIndexStep.java
joshpalis: The resource created list is used in deprovisioning and will be integral for fine-grained provisioning, since this step doesnt create a resource and it doesn't have a deprovision/update equivalent I think we should avoid interacting with the state index entry here. I would consider this similar to the `SearchRequestStep` which just performs an action rather than creating a resource.

src/main/java/org/opensearch/flowframework/workflow/ReindexStep.java
```diff
@@ -0,0 +1,184 @@
...
+                                        + " resource "
+                                        + getName()
+                                        + " id "
+                                        + destinationIndex;
+                                    logger.error(errorMessage, exception);
+                                    reIndexFuture.onFailure(new FlowFrameworkException(errorMessage, ExceptionsHelper.status(exception)));
+                                })
+                            );
+                        } else {
+                            String errorMessage = "Failed to get bulk response";
```
joshpalis, line 153: Im curious if we should output the bulk/search failures as part of the error message to give users the information to fix their request

src/main/java/org/opensearch/flowframework/workflow/ReIndexStep.java
owaiskazi19: Pushed the change

src/main/java/org/opensearch/flowframework/workflow/ReindexStep.java
owaiskazi19: Good point. Included the error message

joshpalis (approved):
LGTM, thanks @owaiskazi19 for addressing the comments

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 ## [Unreleased 2.x](https://github.com/opensearch-project/flow-framework/compare/2.14...2.x)
 ### Features
 ### Enhancements
+- Add Workflow Step for Reindex from source index to destination ([#718](https://github.com/opensearch-project/flow-framework/pull/718))
```
dbwiddis, line 18: Lack of whitespace below this will break the markdown. :)

src/main/java/org/opensearch/flowframework/workflow/ReindexStep.java
```diff
@@ -0,0 +1,175 @@
...
+            Integer slices = (Integer) inputs.get(SLICES);
+            Integer maxDocs = (Integer) inputs.get(MAX_DOCS);
+
+            ReindexRequest reindexRequest = new ReindexRequest().setSourceIndices(Strings.splitStringByCommaToArray(sourceIndices))
+                .setDestIndex(destinationIndex);
+
+            if (refresh != null) {
+                reindexRequest.setRefresh(refresh);
+            }
+            if (requestsPerSecond != null) {
```
dbwiddis, line 108: Since you used a default `Float.POSITIVE_INFINITY` above, this null check is useless.  However I think the better fix here is to have the "else" part of the line 95-97 calculation be `null` rather than assuming the default we only know because of implementation details.

src/main/java/org/opensearch/flowframework/workflow/ReindexStep.java
```diff
@@ -0,0 +1,175 @@
...
+                    String errorMessage = "Failed to reindex from source " + sourceIndices + " to " + destinationIndex;
+                    logger.error(errorMessage, e);
+                    reIndexFuture.onFailure(new WorkflowStepException(errorMessage, ExceptionsHelper.status(e)));
+                }
+            };
+
+            client.execute(ReindexAction.INSTANCE, reindexRequest, actionListener);
+
+        } catch (IllegalArgumentException iae) {
+            reIndexFuture.onFailure(new WorkflowStepException(iae.getMessage(), RestStatus.BAD_REQUEST));
```
dbwiddis, line 163: I like this, possibly add some text before the parsing exception message identifying which step or step id this failure is occurring in?

dbwiddis (approved):
LGTM with a few suggestions.

PR #719, opened 2024-05-29
Add param to delete workflow API to clear status even if resources exist

### LINKED ISSUE #692: [FEATURE] Add a parameter to force deletion of workflow state document when a template is deleted
Coming from https://github.com/opensearch-project/flow-framework/pull/689#issuecomment-2086247313

Presently when a template is deleted, the workflow state is not deleted if resources exist.

However, a user may wish to delete the state document anyway, and track the resources separately.

Add a REST path parameter such as "force" or "delete_state" or similar (defaulting to false) to delete the state document when the template is deleted.

A separate API to delete the state document.

A combined API to delete the template and state together.

dbwiddis:
### Description

Adds an optional `clear_status` param to the delete workflow API (defaulting `false`).

If present, overrides the "do resources exist" check and permits deleting the workflow state as long as provisioning is not `IN_PROGRESS`

Please also review Documentation PR: https://github.com/opensearch-project/documentation-website/pull/7268

### Issues Resolved

Resolves #692

### DIFF
--- a/src/main/java/org/opensearch/flowframework/rest/RestDeleteWorkflowAction.java
+++ b/src/main/java/org/opensearch/flowframework/rest/RestDeleteWorkflowAction.java
@@ -28,6 +28,7 @@ import java.io.IOException;
 import java.util.List;
 import java.util.Locale;
 
+import static org.opensearch.flowframework.common.CommonValue.CLEAR_STATUS;
 import static org.opensearch.flowframework.common.CommonValue.WORKFLOW_ID;
 import static org.opensearch.flowframework.common.CommonValue.WORKFLOW_URI;
 import static org.opensearch.flowframework.common.FlowFrameworkSettings.FLOW_FRAMEWORK_ENABLED;
@@ -62,6 +63,7 @@ public class RestDeleteWorkflowAction extends BaseRestHandler {
     @Override
     protected BaseRestHandler.RestChannelConsumer prepareRequest(RestRequest request, NodeClient client) throws IOException {
         String workflowId = request.param(WORKFLOW_ID);
+        request.param(CLEAR_STATUS); // consume and ignore, we will pass params to workflow
         try {
             if (!flowFrameworkFeatureEnabledSetting.isFlowFrameworkEnabled()) {
                 throw new FlowFrameworkException(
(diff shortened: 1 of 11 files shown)


src/main/java/org/opensearch/flowframework/rest/RestDeleteWorkflowAction.java
```diff
@@ -62,6 +63,7 @@ public List<Route> routes() {
     @Override
     protected BaseRestHandler.RestChannelConsumer prepareRequest(RestRequest request, NodeClient client) throws IOException {
         String workflowId = request.param(WORKFLOW_ID);
+        request.param(CLEAR_STATUS); // consume and ignore, we will pass params to workflow
```
joshpalis, line 66: Just to refresh my memory, why do we need to invoke `param()` for each of these? We consume all the params either way when we send a transport request

joshpalis (approved):
Overall LGTM, just had a question

src/main/java/org/opensearch/flowframework/rest/RestDeleteWorkflowAction.java
dbwiddis: First, I had neglected to push a commit I'd saved locally that changes this to actually pull just this param and send just that map over transport.  About to do that now.

But the answer to the larger question is that if we throw an exception _before_ getting to the `params()` call later, such as if the plugin is disabled, we won't consume it and will get a misleading error message.

When I first tried to locally run integ tests before putting this here, I got this:

> > Task :integTest
> 
> REPRODUCE WITH: ./gradlew ':integTest' --tests "org.opensearch.flowframework.rest.FlowFrameworkRestApiIT.testCreateAndProvisionRemoteModelWorkflow" -Dtests.seed=FBBABD239888A9F8 -Dtests.security.manager=false -Dtests.locale=en-AU -Dtests.timezone=Antarctica/Macquarie -Druntime.java=17
> 
> org.opensearch.flowframework.rest.FlowFrameworkRestApiIT > testCreateAndProvisionRemoteModelWorkflow FAILED
>     org.opensearch.client.ResponseException: method [DELETE], host [http://127.0.0.1:56350], URI [/_plugins/_flow_framework/workflow/qai-xY8B6INOlEYKiSbi?clear_status=true], status line [HTTP/1.1 400 Bad Request]
>     {"error":{"root_cause":[{"type":"illegal_argument_exception","reason":"request [/_plugins/_flow_framework/workflow/qai-xY8B6INOlEYKiSbi] contains unrecognized parameter: [clear_status]"}],"type":"illegal_argument_exception","reason":"request [/_plugins/_flow_framework/workflow/qai-xY8B6INOlEYKiSbi] contains unrecognized parameter: [clear_status]"},"status":400}

owaiskazi19: Similar to https://github.com/opensearch-project/flow-framework/pull/295?

owaiskazi19 (approved):
LGTM! Thanks for addressing this issue

src/main/java/org/opensearch/flowframework/rest/RestDeleteWorkflowAction.java
dbwiddis: > Similar to #295?

Exactly.

Issue #710, opened 2024-05-07
[CI] Improve BWC Tests and coverage


dbwiddis:
### Is your feature request related to a problem?

Our BWC tests are limited and essentially were the minimum required for the feature they were added with.  We need to increase coverage and capability of the tests.

### What solution would you like?

 - [ ] Currently the old version is fixed to 2.12 (for the single feature we test).  We need to test a range of previous versions.
 - [ ] Currently rolling upgrades are disabled on major version changes. We need to test "latest 2.x" vs. 3.x rolling upgrade.
 - [ ] Currently tests run without security plugin. This makes it impossible to use that plugin API to register a `User` to accurately test BWC on indices related to the User field (I'm looking at you, #705).  Tests should run with and without security.

### What alternatives have you considered?

Manually performing one-time BWC checks when introducing a new feature and hoping you covered all the situations (which you probably didn't.)

### Do you have any additional context?

See [this commit](https://github.com/opensearch-project/flow-framework/commit/d071e3162e181703d50db9c5742610a11c18fa0e) for a failed attempt to fix some of the above issues existing tests.

PR #730, opened 2024-06-06
Set Java target compatibility to JDK 21

dbwiddis:
### Description

Sets source and target compatibility to JDK21.

### Issues Resolved

Resoves #729 

### DIFF
--- a/.github/workflows/test_bwc.yml
+++ b/.github/workflows/test_bwc.yml
@@ -11,7 +11,7 @@ jobs:
   Build-ff-linux:
     strategy:
       matrix:
-        java: [11,17,21]
+        java: [21]
       fail-fast: false
 
     name: Test Flow Framework BWC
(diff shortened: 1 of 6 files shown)


.github/workflows/test_bwc.yml
```diff
@@ -11,7 +11,7 @@ jobs:
   Build-ff-linux:
     strategy:
       matrix:
-        java: [11,17,21]
```
owaiskazi19, line 14: Shouldn't we still run bwc tests on 11 and 17?

dbwiddis: Not on 3.x (note this PR will not be backported):
 - we don't currently have any configuration to use different JDK versions per OpenSearch version, and this would cause combinatorial explosion in the matrix of different versions and different JDKs to test against even if we did, without very much real value anyway. 
 - currently we don't even do major version update BWCs, see discussion on #710

joshpalis (approved):

amitgalitz (approved):

owaiskazi19 (approved):

Issue #666, opened 2024-04-18
[META] Improvements for default use cases 


amitgalitz:
### Is your feature request related to a problem?

In 2.13 we released a new feature that enables customers to utilize predefined templates to configure their workflows by selecting a given use case. Currently the feature is documented here: https://opensearch.org/docs/latest/automating-configurations/workflow-templates/

For the first release of the feature in order not to cause any later breaking changes we made a few decisions that can be revisited. 

1. We currently host our substitution ready templates in a folder [here](https://github.com/opensearch-project/flow-framework/tree/main/src/main/resources/substitutionTemplates), and the different defaults for each use case [here](https://github.com/opensearch-project/flow-framework/tree/main/src/main/resources/defaults). The reason these aren’t in the sample is because we didn’t want to make a breaking change to our template and because we felt that in the future their can be more optimal ways to utilize one substitution template with multiple different default files so it wont always be 1:1 match between substitution ready templates and defaults.

2. The defaults and predefined templates are currently in our repo as part of the resources folder to allow for us to have full ownership and to couple them with the official release of the feature in 2.13. However as we expand the number of use cases we support we might need a better path forward on where to host these templates, whether it be in a different structure within the repo, uploaded to a system index, or in an entirely new location. (related: https://github.com/opensearch-project/flow-framework/issues/584)

### Additional Action Items and suggest improvements

- [ ] We need to define criteria on what meets the bar for a new predefined template
- [x] `id` field need to be removed from some of the create index steps in various templates
- [ ] Template that use pretrained model can have their dimensions automatically fetched from get model so user doesn’t have to over-write that parameter if they change model
- [ ] Increased variation in defined template https://github.com/opensearch-project/flow-framework/issues/667
- [ ] https://github.com/opensearch-project/flow-framework/issues/584 
- [ ] Change substiution file names to align with use case names, with `_` instead `-`
- [x] Remove the default weights field from hybrid search on template and defaults file
- [x] Allow capability to change default text and image type for multimodal templates
- [x] Add more pretrained model templates
- [x] Add template with reindex step

owaiskazi19:
Thanks for creating this issue. We should also change the substitution templates files names with `_` instead of `-` since that's what we support in the API calls

```
POST /_plugins/_flow_framework/workflow?use_case=semantic_search_with_cohere_embedding_query_enricher&provision=true
{
    "create_connector.credential.key" : "<YOUR API KEY>"
}
```

amitgalitz:
[...]

PR #731, opened 2024-06-06
Adding additional default use cases

amitgalitz:
### Description
- Adding two additional default use cases for pretrained model for hybrid and semantic search
- removing `id` field from some of the use case templates
- removing weights from hybrid search template
- making the text and image type dynamic for multimodal search as most customers don't actually just use binary for images. 
- Adding e2e integration test for local semantic search template all the way to neural search

### Issues Resolved
part of #666 


joshpalis:
Interesting that the windows single node integration tests are failing due to : 
```
EPRODUCE WITH: gradlew ':integTest' --tests "org.opensearch.flowframework.rest.FlowFrameworkRestApiIT.testSemanticSearchWithLocalModelEndToEnd" -Dtests.seed=F9C1F84B83F9D1C6 -Dtests.security.manager=false -Dtests.locale=mk -Dtests.timezone=PLT -Druntime.java=11

org.opensearch.flowframework.rest.FlowFrameworkRestApiIT > testSemanticSearchWithLocalModelEndToEnd FAILED
    java.lang.AssertionError: expected:<0> but was:<1>
Suite: Test class org.opensearch.flowframework.rest.FlowFrameworkRestApiIT
        at __randomizedtesting.SeedInfo.seed([F9C1F84B83F9D1C6:9469A614D9123AA7]:0)
  2> SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".  2> 
        at org.junit.Assert.fail(Assert.java:89)
```
Looking at the test, its failing during this line : `assertEquals(neuralSearchResponse.getHits().getHits().length, 1);`
Given that the other integration tests passed, perhaps this failure is due to the GHA runner memory limitation. I see that the model used in the semantic search local model template is `msmarco-distilbert-base-tas-b`, which is about 254 mbs. 

Can we use a smaller model here for this test?

build.gradle
```diff
@@ -181,6 +181,8 @@ dependencies {
 
     // ZipArchive dependencies used for integration tests
     zipArchive group: 'org.opensearch.plugin', name:'opensearch-ml-plugin', version: "${opensearch_build}"
+    zipArchive group: 'org.opensearch.plugin', name:'opensearch-knn', version: "${opensearch_build}"
+    zipArchive group: 'org.opensearch.plugin', name:'neural-search', version: "${opensearch_build}"
```
owaiskazi19, line 185: Do we want to commit this?

amitgalitz: Yes it was added for IT testing, if I can't get BWC to work though I will remove this and the IT I added and add that in a different PR.

src/main/java/org/opensearch/flowframework/workflow/ReindexStep.java
```diff
@@ -95,10 +95,20 @@ public PlainActionFuture<WorkflowData> execute(
...
-            Integer slices = (Integer) inputs.get(SLICES);
-            Integer maxDocs = (Integer) inputs.get(MAX_DOCS);
-
+            Integer slices;
+            Integer maxDocs;
+            if (inputs.get(SLICES) != null) {
+                slices = Integer.parseInt(String.valueOf(inputs.get(SLICES)));
+            } else {
+                slices = (Integer) inputs.get(SLICES);
+            }
```
owaiskazi19, line 106:
```java
slices = inputs.get(SLICES) instanceof String
         ? Integer.parseInt(String.valueOf(inputs.get(SLICES)))
         : (Integer) inputs.get(SLICES);
```

amitgalitz: I need to a != null check and with ternary operator you can't do that cause you need both sides of : to not be null unlike if/else

ohltyler (approved):
At a high level LGTM.

joshpalis (approved):

PR #757, opened 2024-07-04
Support editing of certain workflow fields on a provisioned workflow

### LINKED ISSUE #749: [FEATURE] Support editing of certain workflow fields on a provisioned workflow
Currently, the update API will not execute on a workflow that is not in `NOT_STARTED` state (has been provisioned, failed during provision, is currently provisioning, etc.). This prevents losing track of created resources, and enforcing deprovision will work as expected.

The current logic is simplified, and blanket denies any proposed update, even for trivial fields like `name` or `ui_metadata`. This prevents users from updating these things on a provisioned workflow, even if their goal is to not change the critical parts of the workflow template.

Additionally, on the frontend, a desired capability is updating a frontend configuration (which may be incomplete or in-progress) on some provisioned workflow, as a "lightweight" save. _All of the frontend configuration is stored under `ui_metadata`_. This may be done manually by the users, and/or in a scheduled way, such as autosaving. These save scenarios we want to be able to do at any point, without disrupting what may be currently deployed. Only when users want to run actual functionalities against the workflow (ingest/search/export) do we want to update the underlying template and perform the deeper deprovision/reprovision.

One proposal is allowing specific fields in the workflow schema to be allowed to be updated, at the REST level of the update API. I could see a few options for how the interface may look:

Option 1. Using the existing update API and passing in the entire workflow configuration, and letting the backend perform a diff and determine that based on the changed fields and the state of the workflow, if the update is allowed or not
```
PUT /_plugins/_flow_framework/workflow/my-workflow-id
{
    "name": "my-new-name",
    "description": "my existing description",
    "workflows": {
        "provision": { <some-unchanged-provision-flows> }
    },
    ...,
}
```
Option 2. Allowing users to only specify the fields they want to update in the API body, something like:
```
PUT /_plugins/_flow_framework/workflow/my-workflow-id
{
    "name": "my-new-name"
}
```

dbwiddis:
### Description

Allows passing the parameter `update_fields=true` on the REST path when updating a template (PUT).  This will update the existing template with only the included fields.

### Issues Resolved

Resolves #749 

### Documentation

Please review:
 - https://github.com/opensearch-project/documentation-website/pull/7632

### DIFF
--- a/src/main/java/org/opensearch/flowframework/model/Template.java
+++ b/src/main/java/org/opensearch/flowframework/model/Template.java
@@ -295,9 +295,9 @@ public class Template implements ToXContentObject {
     @Override
     public XContentBuilder toXContent(XContentBuilder builder, Params params) throws IOException {
         XContentBuilder xContentBuilder = builder.startObject();
-        xContentBuilder.field(NAME_FIELD, this.name);
-        xContentBuilder.field(DESCRIPTION_FIELD, this.description == null ? "" : this.description);
-        xContentBuilder.field(USE_CASE_FIELD, this.useCase == null ? "" : this.useCase);
+        xContentBuilder.field(NAME_FIELD, this.name.trim());
+        xContentBuilder.field(DESCRIPTION_FIELD, this.description == null ? "" : this.description.trim());
+        xContentBuilder.field(USE_CASE_FIELD, this.useCase == null ? "" : this.useCase.trim());
 
         if (this.templateVersion != null || !this.compatibilityVersion.isEmpty()) {
             xContentBuilder.startObject(VERSION_FIELD);
--- a/src/main/java/org/opensearch/flowframework/model/Template.java
+++ b/src/main/java/org/opensearch/flowframework/model/Template.java
@@ -29,6 +29,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Set;
 
 import static org.opensearch.core.xcontent.XContentParserUtils.ensureExpectedToken;
 import static org.opensearch.flowframework.common.CommonValue.CREATED_TIME;
@@ -53,6 +54,14 @@ public class Template implements ToXContentObject {
     public static final String TEMPLATE_FIELD = "template";
     /** The template field name for template use case */
     public static final String USE_CASE_FIELD = "use_case";
+    /** Fields which may be updated in the template even if provisioned */
+    private static final Set<String> UPDATE_FIELD_ALLOWLIST = Set.of(
+        NAME_FIELD,
+        DESCRIPTION_FIELD,
+        USE_CASE_FIELD,
+        VERSION_FIELD,
+        UI_METADATA_FIELD
+    );
 
     private final String name;
     private final String description;
@@ -287,8 +296,8 @@ public class Template implements ToXContentObject {
     public XContentBuilder toXContent(XContentBuilder builder, Params params) throws IOException {
         XContentBuilder xContentBuilder = builder.startObject();
         xContentBuilder.field(NAME_FIELD, this.name);
-        xContentBuilder.field(DESCRIPTION_FIELD, this.description);
-        xContentBuilder.field(USE_CASE_FIELD, this.useCase);
+        xContentBuilder.field(DESCRIPTION_FIELD, this.description == null ? "" : this.description);
+        xContentBuilder.field(USE_CASE_FIELD, this.useCase == null ? "" : this.useCase);
 
         if (this.templateVersion != null || !this.compatibilityVersion.isEmpty()) {
             xContentBuilder.startObject(VERSION_FIELD);
@@ -342,9 +351,21 @@ public class Template implements ToXContentObject {
      * @throws IOException if content can't be parsed correctly
      */
     public static Template parse(XContentParser parser) throws IOException {
+        return parse(parser, false);
+    }
+
+    /**
+     * Parse raw xContent into a Template instance.
+     *
+     * @param parser xContent based content parser
+     * @param fieldUpdate if set true, will be used for updating an existing template
+     * @return an instance of the template
+     * @throws IOException if content can't be parsed correctly
+     */
+    public static Template parse(XContentParser parser, boolean fieldUpdate) throws IOException {
         String name = null;
-        String description = "";
-        String useCase = "";
+        String description = null;
+        String useCase = null;
         Version templateVersion = null;
         List<Version> compatibilityVersion = new ArrayList<>();
         Map<String, Workflow> workflows = new HashMap<>();
@@ -357,6 +378,12 @@ public class Template implements ToXContentObject {
         ensureExpectedToken(XContentParser.Token.START_OBJECT, parser.currentToken(), parser);
         while (parser.nextToken() != XContentParser.Token.END_OBJECT) {
             String fieldName = parser.currentName();
+            if (fieldUpdate && !UPDATE_FIELD_ALLOWLIST.contains(fieldName)) {
+                throw new FlowFrameworkException(
+                    "You can not update the field [" + fieldName + "] without updating the whole template.",
+                    RestStatus.BAD_REQUEST
+                );
+            }
             parser.nextToken();
             switch (fieldName) {
                 case NAME_FIELD:
@@ -421,8 +448,16 @@ public class Template implements ToXContentObject {
                     );
             }
         }
-        if (name == null) {
-            throw new FlowFrameworkException("A template object requires a name.", RestStatus.BAD_REQUEST);
+        if (!fieldUpdate) {
+            if (name == null) {
+                throw new FlowFrameworkException("A template object requires a name.", RestStatus.BAD_REQUEST);
+            }
+            if (description == null) {
+                description = "";
+            }
+            if (useCase == null) {
+                useCase = "";
+            }
         }
 
         return new Builder().name(name)
--- a/src/main/java/org/opensearch/flowframework/model/Template.java
+++ b/src/main/java/org/opensearch/flowframework/model/Template.java
@@ -8,6 +8,7 @@
  */
 package org.opensearch.flowframework.model;
 
+import org.apache.logging.log4j.util.Strings;
 import org.opensearch.Version;
 import org.opensearch.common.xcontent.LoggingDeprecationHandler;
 import org.opensearch.common.xcontent.json.JsonXContent;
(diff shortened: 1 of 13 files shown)


dbwiddis:
```json
POST localhost:9200/_plugins/_flow_framework/workflow?provision=true
{
  "name": "test",
  "description": "Flow template",
  "use_case": "TESTING",
  "version": {
    "template": "1.0.0",
    "compatibility": ["2.12.0", "3.0.0"]
  },
  "workflows": {
    "provision": {
      "user_params": {},
      "nodes": [
        {
          "id": "no op",
          "type": "noop"
        }
      ]
    }
  }
}
```

```json
{
    "workflow_id": "yl5IfJABgzSGdpsYUKJE"
}
```

```json
GET localhost:9200/_plugins/_flow_framework/workflow/yl5IfJABgzSGdpsYUKJE
```

```json
{
    "name": "test",
    "description": "Flow template",
    "use_case": "TESTING",
    "version": {
        "template": "1.0.0",
        "compatibility": [
            "2.12.0",
            "3.0.0"
        ]
    },
    "workflows": {
        "provision": {
            "user_params": {},
            "nodes": [
                {
                    "id": "no op",
                    "type": "noop",
                    "previous_node_inputs": {},
                    "user_inputs": {}
                }
            ],
            "edges": []
        }
    },
    "created_time": 1720072032206,
    "last_updated_time": 1720072032206,
    "last_provisioned_time": 1720072032727
}
```

```json
GET localhost:9200/_plugins/_flow_framework/workflow/yl5IfJABgzSGdpsYUKJE/_status
```

```json
{
    "workflow_id": "yl5IfJABgzSGdpsYUKJE",
    "state": "COMPLETED"
}
```

```json
PUT localhost:9200/_plugins/_flow_framework/workflow/yl5IfJABgzSGdpsYUKJE
{
  "name": "a test with no name",
  "description": "Updated flow template",
  "use_case": "TESTING",
  "version": {
    "template": "1.0.0",
    "compatibility": ["2.12.0", "3.0.0"]
  },
  "workflows": {
    "provision": {
      "user_params": {},
      "nodes": [
        {
          "id": "no op",
          "type": "noop"
        }
      ]
    }
  }
}
```

```json
400 Bad Request
{
    "error": "The template can not be updated unless its provisioning state is NOT_STARTED: yl5IfJABgzSGdpsYUKJE. Deprovision the workflow to reset the state."
}
```

```json
PUT localhost:9200/_plugins/_flow_framework/workflow/yl5IfJABgzSGdpsYUKJE?update_fields=true
{
  "workflows": {
    "provision": {
      "user_params": {},
      "nodes": [
        {
          "id": "no op",
          "type": "noop"
        }
      ]
    }
  }
}
```

```json
400 Bad Request
{
    "error": "You can not update the field [workflows] without updating the whole template."
}
```

```json
PUT localhost:9200/_plugins/_flow_framework/workflow/yl5IfJABgzSGdpsYUKJE?update_fields=true
{
  "name": "a test with no name",
  "description": "Updated flow template"
}
```

```json
{
    "workflow_id": "yl5IfJABgzSGdpsYUKJE"
}
```

```json
GET localhost:9200/_plugins/_flow_framework/workflow/yl5IfJABgzSGdpsYUKJE
```
```json
{
    "name": "a test with no name",
    "description": "Updated flow template",
    "use_case": "TESTING",
    "version": {
        "template": "1.0.0",
        "compatibility": [
            "2.12.0",
            "3.0.0"
        ]
    },
    "workflows": {
        "provision": {
            "user_params": {},
            "nodes": [
                {
                    "id": "no op",
                    "type": "noop",
                    "previous_node_inputs": {},
                    "user_inputs": {}
                }
            ],
            "edges": []
        }
    },
    "created_time": 1720072032206,
    "last_updated_time": 1720077009607,
    "last_provisioned_time": 1720072032727
}
```

dbwiddis:
Failing CI tracked in https://github.com/opensearch-project/flow-framework/issues/759

ohltyler (approved):
LGTM! Thanks for adding this. The allowlisted fields should be sufficient for the frontend.

src/main/java/org/opensearch/flowframework/transport/CreateWorkflowTransportAction.java
```diff
@@ -233,43 +235,80 @@ protected void doExecute(Task task, WorkflowRequest request, ActionListener<Work
...
+                            if (templateWithUser.templateVersion() != null) {
+                                builder.templateVersion(templateWithUser.templateVersion());
+                            }
+                            if (!templateWithUser.compatibilityVersion().isEmpty()) {
+                                builder.compatibilityVersion(templateWithUser.compatibilityVersion());
+                            }
+                            if (templateWithUser.getUiMetadata() != null) {
+                                builder.uiMetadata(templateWithUser.getUiMetadata());
+                            }
+                            template = builder.build();
```
owaiskazi19, line 268: We should create a separate method for the builder here

src/main/java/org/opensearch/flowframework/model/Template.java
```diff
@@ -287,8 +296,8 @@ public Template build() {
     public XContentBuilder toXContent(XContentBuilder builder, Params params) throws IOException {
         XContentBuilder xContentBuilder = builder.startObject();
         xContentBuilder.field(NAME_FIELD, this.name);
-        xContentBuilder.field(DESCRIPTION_FIELD, this.description);
-        xContentBuilder.field(USE_CASE_FIELD, this.useCase);
+        xContentBuilder.field(DESCRIPTION_FIELD, this.description == null ? "" : this.description);
```
owaiskazi19, line 299:
```suggestion
        xContentBuilder.field(DESCRIPTION_FIELD, this.description == null ? "" : this.description.trim());
```

src/main/java/org/opensearch/flowframework/model/Template.java
```diff
@@ -287,8 +296,8 @@ public Template build() {
...
+        xContentBuilder.field(USE_CASE_FIELD, this.useCase == null ? "" : this.useCase);
```
owaiskazi19, line 300:
```suggestion
        xContentBuilder.field(USE_CASE_FIELD, this.useCase == null ? "" : this.useCase.trim());
```

src/main/java/org/opensearch/flowframework/model/Template.java
```diff
@@ -53,6 +54,14 @@ public class Template implements ToXContentObject {
...
     /** The template field name for template use case */
     public static final String USE_CASE_FIELD = "use_case";
+    /** Fields which may be updated in the template even if provisioned */
+    private static final Set<String> UPDATE_FIELD_ALLOWLIST = Set.of(
+        NAME_FIELD,
+        DESCRIPTION_FIELD,
+        USE_CASE_FIELD,
+        VERSION_FIELD,
+        UI_METADATA_FIELD
+    );
```
owaiskazi19, line 64: Can we copy/paste this allowlist on the issue to keep the track on?

owaiskazi19 (commented):
Looks good overall with minor suggestions

src/main/java/org/opensearch/flowframework/transport/CreateWorkflowTransportAction.java
dbwiddis: Not sure what you mean here.

src/main/java/org/opensearch/flowframework/model/Template.java
dbwiddis: [Done](https://github.com/opensearch-project/flow-framework/issues/749#issuecomment-2214960669)

src/main/java/org/opensearch/flowframework/transport/CreateWorkflowTransportAction.java
owaiskazi19: For[ L249-268](https://github.com/opensearch-project/flow-framework/pull/757/files#diff-167ff0aad0354b79b0a51b5713f3e951d9af59ded66c73f35cbece5f76ac90fcR249-R268), we should move them to a new method rather the current method bulky.

dbwiddis: I really don't see a way to do that generally. They are all conditional updates of fields in the builder.  At best I could just chain the builder setters like usually and internally handle null, but it's entirely possible that I would want to set some fields to null.  I could maybe iterate over a list of String field names and move those into a map in the builder that maintains the fields to be set instead of object fields but that just moves this relatively simple list of conditionals into ugly code implementation.

If you have an example of how you'd do this simpler I'd love to see it.

owaiskazi19: Something like the below

```java
Template existingTemplate = Template.parse(getResponse.getSourceAsString());
Template template = buildTemplate(existingTemplate, templateWithUser, isFieldUpdate);
            
private Template buildTemplate(Template existingTemplate, Template templateWithUser, boolean isFieldUpdate) {
    if (isFieldUpdate) {
        // Update only specified fields in existing
        Builder builder = new Template.Builder(existingTemplate).lastUpdatedTime(Instant.now());
        updateBuilder(builder, templateWithUser);
        return builder.build();
    } else {
        // Update existing entry, full document replacement
        return new Template.Builder(templateWithUser)
                .createdTime(existingTemplate.createdTime())
                .lastUpdatedTime(Instant.now())
                .lastProvisionedTime(existingTemplate.lastProvisionedTime())
                .build();
    }
}

private void updateBuilder(Builder builder, Template templateWithUser) {
    if (templateWithUser.name() != null) {
        builder.name(templateWithUser.name());
    }
    if (!Strings.isBlank(templateWithUser.description())) {
        builder.description(templateWithUser.description());
    }
    if (!Strings.isBlank(templateWithUser.useCase())) {
        builder.useCase(templateWithUser.useCase());
    }
    if (templateWithUser.templateVersion() != null) {
        builder.templateVersion(templateWithUser.templateVersion());
    }
    if (!templateWithUser.compatibilityVersion().isEmpty()) {
        builder.compatibilityVersion(templateWithUser.compatibilityVersion());
    }
    if (templateWithUser.getUiMetadata() != null) {
        builder.uiMetadata(templateWithUser.getUiMetadata());
    }
}
```

dbwiddis: @owaiskazi19 d0672728054ba702ca4ddd67e386f9b889707113

owaiskazi19 (approved):
Thanks for addressing the comments. LGTM!

dbwiddis:
Tests failing on WIndows because clock resolution isn't sufficient to guarantee "instant" increments. :|

Issue #579, opened 2024-03-14
[FEATURE] Implement Notification for Manual Deletion of Certain Resources


owaiskazi19:
### Is your feature request related to a problem?
Referencing https://github.com/opensearch-project/flow-framework/pull/574#discussion_r1525111295, certain workflow steps cannot include a deletion action due to the potential risk of removing critical resources or data. An example of this is the DeleteIndex workflow step. In such scenarios, upon completion of deprovisioning, users should be informed to manually delete such resources.

### What solution would you like?
For workflow steps that include a NoopStep for deletion (such as [this example](https://github.com/opensearch-project/flow-framework/blob/main/src/main/java/org/opensearch/flowframework/common/WorkflowResources.java#L52)), users should be notified to manually delete the resource by printing a message along with the deprovision API call.

Example: `Delete RegisterGroupModel manually`

### What alternatives have you considered?
Just have a NoopStep for such cases and don't notify the user.

### Do you have any additional context?
_Add any other context or screenshots about the feature request here._

PR #574, opened 2024-03-13
Added Create index workflow step

owaiskazi19:
### Description
- Added Create index and Delete index workflow steps
- Tests

Request
```
{
                    "id": "create_index",
                    "type": "create_index",
                    "user_inputs": {
                        "index_name": "append-4",
                        "configurations": {
                            "settings": {
                                "index": {
                                    "number_of_shards": 2,
                                    "number_of_replicas": 1
                                }
                            },
                            "mappings": {
                                "_doc": {
                                    "properties": {
                                        "age": {
                                            "type": "integer"
                                        }
                                    }
                                }
                            },
                            "aliases": {
                                "sample-alias1": {}
                            }
                        }
                    }
                }
```

Response for Create Index Provisioning
```
[2024-03-13T01:20:08,706][INFO ][o.o.f.w.ProcessNode      ] [ip-172-31-56-214] Starting create_index.
[2024-03-13T01:20:08,709][INFO ][o.o.p.PluginsService     ] [ip-172-31-56-214] PluginService:onIndexModule index:[append-4/stmxO9k-SqyUqqf1RC6dRQ]
[2024-03-13T01:20:08,712][INFO ][o.o.c.m.MetadataCreateIndexService] [ip-172-31-56-214] [append-4] creating index, cause [api], templates [], shards [2]/[1]
[2024-03-13T01:20:08,725][INFO ][o.o.p.PluginsService     ] [ip-172-31-56-214] PluginService:onIndexModule index:[append-4/stmxO9k-SqyUqqf1RC6dRQ]
[2024-03-13T01:20:08,773][INFO ][o.o.f.w.CreateIndexStep  ] [ip-172-31-56-214] Created index: append-4
[2024-03-13T01:20:08,785][INFO ][o.o.f.i.FlowFrameworkIndicesHandler] [ip-172-31-56-214] updated resources created of sN5lNY4B10dFQcqyDKnd
[2024-03-13T01:20:08,785][INFO ][o.o.f.w.CreateIndexStep  ] [ip-172-31-56-214] successfully updated resource created in state index: .plugins-flow-framework-state
[2024-03-13T01:20:08,785][INFO ][o.o.f.w.ProcessNode      ] [ip-172-31-56-214] Finished create_index.
[2024-03-13T01:20:08,785][INFO ][o.o.f.t.ProvisionWorkflowTransportAction] [ip-172-31-56-214] Provisioning completed successfully for workflow sN5lNY4B10dFQcqyDKnd
[...]

PR #763, opened 2024-07-06
Add allow_delete parameter to Deprovision API

### LINKED ISSUE #748: [FEATURE] Support optional force deletion of certain resources
Currently, certain resources are blocked from being deleted during deprovisioning, including indices, as a precaution to prevent users from accidentally deleting potential data.

By allowing the ability to delete such resources if the user explicitly wants to, this gives users more flexibility. It simplifies the experience of performing iterative testing / rapid prototyping on a workflow, where users may be constantly deprovisioning / reprovisioning resources.

On the frontend, the UX is focused on iterative testing of workflows, most of which involve index creation. Currently, this causes errors during reprovisioning due to the created index resource existing. There are frontend-only solutions for this problem, but ideally this is allowed directly from the backend for consistency and ease of use for API users.

Proposal: an `allow_delete` list param - a fine-grained parameter for deleting any specially-handled resources specified in the list of resource types passed
```
POST .../_deprovision?allow_delete=index,ingest_pipeline,foo
{
"workflow_id" : "abc"
}
```
Internally, we can add a delete index workflow step (and for any other resources that have this guardrail) that will only be executed when the parameter logic specifies as such.

Additionally, we may want some notification / visibility to users on such scenarios. See #579 for details.

dbwiddis:
### Description

Adds an `allow_delete` parameter to the Deprovision Workflow API which can contain a comma-delimited list of resource ids  which may be deleted.  

When the parameter does not contain resources that require it, the deprovision API returns a 403 response identifying the remaining resources.

Added `DeleteIndexStep`, `DeleteIngestPipelineStep`, and `DeleteSearchPipelineStep` which require the use of this parameter.

Added a denylist to prevent the use of these steps when provisioning workflows, and filtered the Get Workflow Steps API return to remove them from that list as well

### Issues Resolved

- #748
- #579

### Documentation

Please review:
 - https://github.com/opensearch-project/documentation-website/pull/7639

### Note to Reviewers

This implementation causes a change in API behavior.

Current behavior is to simply remove these unimplemented deletions from the resource list and silently fail, giving a 200 (OK) on successful deprovision while the resources remain.  (As a concrete example, if an index was created and included in the created resources list, deprovisioning would simply remove it from the resources list without taking any action (deleting the index).)

This is misleading, as an attempt to provision again will fail with a pre-existing index, so I think the change is justified, but should be made clear to anyone using existing scripting relying on this behavior.

I think 403 (Forbidden) is the best response type here, but I'm open to discussion here if you have a better proposal.

### DIFF
--- a/src/main/java/org/opensearch/flowframework/common/WorkflowResources.java
+++ b/src/main/java/org/opensearch/flowframework/common/WorkflowResources.java
@@ -18,7 +18,10 @@ import org.opensearch.flowframework.workflow.CreateIngestPipelineStep;
 import org.opensearch.flowframework.workflow.CreateSearchPipelineStep;
 import org.opensearch.flowframework.workflow.DeleteAgentStep;
 import org.opensearch.flowframework.workflow.DeleteConnectorStep;
+import org.opensearch.flowframework.workflow.DeleteIndexStep;
+import org.opensearch.flowframework.workflow.DeleteIngestPipelineStep;
 import org.opensearch.flowframework.workflow.DeleteModelStep;
+import org.opensearch.flowframework.workflow.DeleteSearchPipelineStep;
 import org.opensearch.flowframework.workflow.DeployModelStep;
 import org.opensearch.flowframework.workflow.NoOpStep;
 import org.opensearch.flowframework.workflow.RegisterAgentStep;
@@ -54,11 +57,11 @@ public enum WorkflowResources {
     /** Workflow steps for deploying/undeploying a model and associated created resource */
     DEPLOY_MODEL(DeployModelStep.NAME, WorkflowResources.MODEL_ID, UndeployModelStep.NAME),
     /** Workflow steps for creating an ingest-pipeline and associated created resource */
-    CREATE_INGEST_PIPELINE(CreateIngestPipelineStep.NAME, WorkflowResources.PIPELINE_ID, null), // TODO delete step
+    CREATE_INGEST_PIPELINE(CreateIngestPipelineStep.NAME, WorkflowResources.PIPELINE_ID, DeleteIngestPipelineStep.NAME),
     /** Workflow steps for creating an ingest-pipeline and associated created resource */
-    CREATE_SEARCH_PIPELINE(CreateSearchPipelineStep.NAME, WorkflowResources.PIPELINE_ID, null), // TODO delete step
+    CREATE_SEARCH_PIPELINE(CreateSearchPipelineStep.NAME, WorkflowResources.PIPELINE_ID, DeleteSearchPipelineStep.NAME),
     /** Workflow steps for creating an index and associated created resource */
-    CREATE_INDEX(CreateIndexStep.NAME, WorkflowResources.INDEX_NAME, NoOpStep.NAME),
+    CREATE_INDEX(CreateIndexStep.NAME, WorkflowResources.INDEX_NAME, DeleteIndexStep.NAME),
     /** Workflow steps for reindex a source index to destination index and associated created resource */
     REINDEX(ReindexStep.NAME, WorkflowResources.INDEX_NAME, NoOpStep.NAME),
     /** Workflow steps for registering/deleting an agent and the associated created resource */
--- a/src/main/java/org/opensearch/flowframework/transport/DeprovisionWorkflowTransportAction.java
+++ b/src/main/java/org/opensearch/flowframework/transport/DeprovisionWorkflowTransportAction.java
@@ -19,6 +19,7 @@ import org.opensearch.client.Client;
 import org.opensearch.common.inject.Inject;
 import org.opensearch.common.util.concurrent.ThreadContext;
 import org.opensearch.core.action.ActionListener;
+import org.opensearch.core.common.Strings;
 import org.opensearch.core.rest.RestStatus;
 import org.opensearch.flowframework.common.FlowFrameworkSettings;
 import org.opensearch.flowframework.exception.FlowFrameworkException;
@@ -28,6 +29,7 @@ import org.opensearch.flowframework.model.ResourceCreated;
 import org.opensearch.flowframework.model.State;
 import org.opensearch.flowframework.workflow.ProcessNode;
 import org.opensearch.flowframework.workflow.WorkflowData;
+import org.opensearch.flowframework.workflow.WorkflowStep;
 import org.opensearch.flowframework.workflow.WorkflowStepFactory;
 import org.opensearch.tasks.Task;
 import org.opensearch.threadpool.ThreadPool;
@@ -40,8 +42,10 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Set;
 import java.util.stream.Collectors;
 
+import static org.opensearch.flowframework.common.CommonValue.ALLOW_DELETE;
 import static org.opensearch.flowframework.common.CommonValue.DEPROVISION_WORKFLOW_THREAD_POOL;
 import static org.opensearch.flowframework.common.CommonValue.PROVISIONING_PROGRESS_FIELD;
 import static org.opensearch.flowframework.common.CommonValue.PROVISION_END_TIME_FIELD;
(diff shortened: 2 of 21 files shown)


ohltyler (approved):
At a high level LGTM, thanks for adding this! Meets the frontend requirements and will make deprovisioning/reprovisioning a much smoother experience. I concur with a 403 response on resources left over, I think that makes the most sense.

src/main/java/org/opensearch/flowframework/common/WorkflowResources.java
```diff
@@ -54,11 +57,11 @@ public enum WorkflowResources {
...
     DEPLOY_MODEL(DeployModelStep.NAME, WorkflowResources.MODEL_ID, UndeployModelStep.NAME),
     /** Workflow steps for creating an ingest-pipeline and associated created resource */
-    CREATE_INGEST_PIPELINE(CreateIngestPipelineStep.NAME, WorkflowResources.PIPELINE_ID, null), // TODO delete step
+    CREATE_INGEST_PIPELINE(CreateIngestPipelineStep.NAME, WorkflowResources.PIPELINE_ID, DeleteIngestPipelineStep.NAME),
     /** Workflow steps for creating an ingest-pipeline and associated created resource */
-    CREATE_SEARCH_PIPELINE(CreateSearchPipelineStep.NAME, WorkflowResources.PIPELINE_ID, null), // TODO delete step
+    CREATE_SEARCH_PIPELINE(CreateSearchPipelineStep.NAME, WorkflowResources.PIPELINE_ID, DeleteSearchPipelineStep.NAME),
     /** Workflow steps for creating an index and associated created resource */
-    CREATE_INDEX(CreateIndexStep.NAME, WorkflowResources.INDEX_NAME, NoOpStep.NAME),
+    CREATE_INDEX(CreateIndexStep.NAME, WorkflowResources.INDEX_NAME, DeleteIndexStep.NAME),
```
owaiskazi19, line 64: We removed this step because we don't want to let user delete any customer related data from flow framework. 
@amitgalitz can you also take a look here?

src/main/java/org/opensearch/flowframework/transport/DeprovisionWorkflowTransportAction.java
```diff
@@ -244,35 +263,49 @@ private void updateWorkflowState(
...
+                    Map.entry(RESOURCES_CREATED_FIELD, stateIndexResources)
                 ),
                 ActionListener.wrap(updateResponse -> {
                     logger.info("updated workflow {} state to COMPLETED", workflowId);
                 }, exception -> { logger.error("Failed to update workflow {} state", workflowId, exception); })
             );
             // give user list of remaining resources
+            StringBuilder message = new StringBuilder();
+            appendResourceInfo(message, "Failed to deprovision some resources: ", remainingResources);
+            appendResourceInfo(message, "These resources require the " + ALLOW_DELETE + " parameter to deprovision: ", deleteNotAllowed);
```
owaiskazi19, line 285: Can we merge these 2 messages?

src/main/java/org/opensearch/flowframework/common/WorkflowResources.java
amitgalitz: yeah I think though we have made the decision that its okay to do this with the additional parameter that gets added. Customers in an FGAC domain will still need permission to delete that index from security plugin.

src/main/java/org/opensearch/flowframework/transport/DeprovisionWorkflowTransportAction.java
dbwiddis: What do you mean by merge? They are appended onto the same string when relevant; see the test cases.

src/main/java/org/opensearch/flowframework/common/WorkflowResources.java
dbwiddis: > Customers in an FGAC domain will still need permission 

Does that happen automatically or is there further validation that has to be done?

amitgalitz: happens automatically, would be good to at least manually check :)

src/main/java/org/opensearch/flowframework/transport/DeprovisionWorkflowTransportAction.java
owaiskazi19: Ah the test cleared my doubt.

src/main/java/org/opensearch/flowframework/common/WorkflowResources.java
dbwiddis: > We removed this step because we don't want to let user delete any customer related data from flow framework. 

It wasn't quite that broad a decision.  We didn't want a deletion hidden behind an API that did not make the deletion clear, specifically "deprovision".  The addition of a deletion step actually allows users to use that step in a workflow and delete indices (or even delete all indices if they use `*`.  But it will have the word `delete` in the step and be obvious.  And won't give them permission to do anything they couldn't do anyway directly via the API.

Here's the original comment that prompted this feature:
 - https://github.com/opensearch-project/flow-framework/pull/574#discussion_r1525111295

src/main/java/org/opensearch/flowframework/workflow/DeleteSearchPipelineStep.java
```diff
@@ -0,0 +1,108 @@
...
+import java.util.Map;
+import java.util.Set;
+
+import static org.opensearch.flowframework.common.WorkflowResources.PIPELINE_ID;
+import static org.opensearch.flowframework.exception.WorkflowStepException.getSafeException;
+
+/**
+ * Step to delete a search pipeline
+ */
+public class DeleteSearchPipelineStep implements WorkflowStep {
```
owaiskazi19, line 31: Looks like except request the code is same for both the steps. We can create an abstract class for both DeleteSearch and DeleteIngest Pipeline step.

owaiskazi19 (commented):
Looks good overall with one comment

src/main/java/org/opensearch/flowframework/workflow/DeleteSearchPipelineStep.java
dbwiddis: Yeah, we could, but I don't think there's any benefit.

I did look at the abstract class for create when writing this and considered it.

Unfortunately the Create abstract class breaks the beauty of inheritance by doing this silly stuff internally, which it should not have done. (If you insist on combining the rest of the method, just call an abstract method here and implement it on the subclass.)
```json
            if (pipelineToBeCreated.equals(CreateSearchPipelineStep.NAME)) {
                PutSearchPipelineRequest putSearchPipelineRequest = new PutSearchPipelineRequest(
                    pipelineId,
                    configurationsBytes,
                    XContentType.JSON
                );
                clusterAdminClient.putSearchPipeline(putSearchPipelineRequest, putPipelineActionListener);
            } else {
                PutPipelineRequest putPipelineRequest = new PutPipelineRequest(pipelineId, configurationsBytes, XContentType.JSON);
                clusterAdminClient.putPipeline(putPipelineRequest, putPipelineActionListener);
            }
```

This appears to make some sense for the create step because of the overlap in code, but I'd propose that even that is a misguided refactoring.  All the "failed to update resource in state index" should be handled inside the `FlowFrameworkIndicesHandler` class.  That bit of code in the abstract class is [duplicated in almost every resource-creating step](https://github.com/search?q=repo%3Aopensearch-project%2Fflow-framework%20flowFrameworkIndicesHandler.updateResourceInStateIndex(&type=code).   So there's some refactoring to be done, but not into an abstract parent class!

Subtracting out the `FlowFrameworkIndicesHandler` code and the static constants that belong in each subclass, there's really not much left to put in the parent.  

And the delete steps are deprovision steps, so they don't even use that code; the deprovision API handles updating the state index.

The only real duplicated code I see is fetching the previous inputs via the ParseUtil (one method) and then pulling the PIPELINE_ID out of the map.  Perhaps the `client` instance field that could be put in the superclass.   That's essentially two lines of duplicated code (or 3 if you move the client field).   The rest is unique (static constants in the class, a different request class, and a different cluster admin client call. (Sure, we could maybe extract out the AcknowledgedResponse handling but that's still just a few lines of code.)

I don't think an abstract class is useful at all in this case, and may not be very useful in the Create case either once the indices handler class handles all that repeat code and the incorrect subclass implementations are properly removed.

owaiskazi19: Agreed. Since our lot of code is getting duplicated in multiple workflow steps, we can refactor or pull that code out in the parent abstract class. Can we have an issue for refactoring?

owaiskazi19 (approved):

PR #805, opened 2024-07-25
Handle Not Found deprovision exceptions as successful deletions

### LINKED ISSUE #803: [BUG] DeleteModelStep and DeleteAgentStep don't handle NOT_FOUND exceptions
### What is the bug?

When deleting a connector, the Delete action is called directly without checking for existence of the connector; the DeleteResponse includes a `DocWriteReponse.NOT_FOUND` specifically to handle this use case.

However, for deleting models and deleting agents, ML Commons does a Get first (to check for various other reasons we can not delete) and in the case of the model/agent not being found, throws an exception instead.

Thus it is impossible to deprovision a model or agent that does not exist.

1. Provision a workflow that creates a model and/or agent.
2. Manually delete the model and/or agent with ML Commons API
3. Attempt to deprovision the workflow.

NOT_FOUND is not an error when deleting, and the resource is removed.

dbwiddis:
### Description

Treats a NOT_FOUND exception on deprovisioning agents and models as successful deprovision.

### Related Issues

Fixes #803 

### Check List

### DIFF
--- a/src/main/java/org/opensearch/flowframework/workflow/DeleteAgentStep.java
+++ b/src/main/java/org/opensearch/flowframework/workflow/DeleteAgentStep.java
@@ -11,9 +11,11 @@ package org.opensearch.flowframework.workflow;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.opensearch.ExceptionsHelper;
+import org.opensearch.OpenSearchStatusException;
 import org.opensearch.action.delete.DeleteResponse;
 import org.opensearch.action.support.PlainActionFuture;
 import org.opensearch.core.action.ActionListener;
+import org.opensearch.core.rest.RestStatus;
 import org.opensearch.flowframework.exception.FlowFrameworkException;
 import org.opensearch.flowframework.exception.WorkflowStepException;
 import org.opensearch.flowframework.util.ParseUtils;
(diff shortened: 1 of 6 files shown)


owaiskazi19 (approved):

amitgalitz (approved):

src/main/java/org/opensearch/flowframework/workflow/DeleteAgentStep.java
```diff
@@ -85,9 +87,21 @@ public void onResponse(DeleteResponse deleteResponse) {
                 @Override
                 public void onFailure(Exception ex) {
                     Exception e = getSafeException(ex);
-                    String errorMessage = (e == null ? "Failed to delete agent " + agentId : e.getMessage());
-                    logger.error(errorMessage, e);
-                    deleteAgentFuture.onFailure(new WorkflowStepException(errorMessage, ExceptionsHelper.status(e)));
+                    // NOT_FOUND exception should be treated as successful deletion
```
amitgalitz, line 90: it would be good to double check what ml-commons returns in case when user has no permission to access the model when deleting. Either manually or through code

dbwiddis: @amitgalitz 
[Delete Model](https://github.com/opensearch-project/ml-commons/blob/5ad26df7dd23a6ed37a13d80548779579521b13f/plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java) returns NOT_FOUND if a get request for that model has false `isExists()`.  It returns FORBIDDEN for no permissions.

[Delete Agent](https://github.com/opensearch-project/ml-commons/blob/5ad26df7dd23a6ed37a13d80548779579521b13f/plugin/src/main/java/org/opensearch/ml/action/agents/DeleteAgentTransportAction.java) does the same.

Issue #822, opened 2024-08-04
[FEATURE] Create security-enabled/non-security enabled integration and BWC tests for reprovisioning


joshpalis:
### Is your feature request related to a problem?
Related to :
- issue : https://github.com/opensearch-project/flow-framework/issues/717
- PR :   https://github.com/opensearch-project/flow-framework/pull/804

### What solution would you like?
Create integration tests for fine grained reprovisioning for the following cases : 

- additive modifications to provisioned templates
- update modifications to provisioned templates
- BWC tests for workflow request serialization
- deletion of workflow nodes when reprovisioning
- Tests for supported update steps (update search/ingest pipelines, update index settings)
- Tests for non-supported update steps

PR #834, opened 2024-08-13
Adding reprovision integration tests

joshpalis:
### Description
Adds integration tests for reprovision param. 

### Related Issues
resolves https://github.com/opensearch-project/flow-framework/issues/822

### Check List


src/test/java/org/opensearch/flowframework/rest/FlowFrameworkRestApiIT.java
```diff
@@ -363,6 +363,281 @@ public void testCreateAndProvisionAgentFrameworkWorkflow() throws Exception {
...
+        // Begin with a template to register a local pretrained model
+        Template template = TestHelpers.createTemplateFromFile("registerremotemodel.json");
+
+        // Hit Create Workflow API to create agent-framework template, with template validation check and provision parameter
+        Response response;
+        if (!indexExistsWithAdminClient(".plugins-ml-config")) {
+            assertBusy(() -> assertTrue(indexExistsWithAdminClient(".plugins-ml-config")), 40, TimeUnit.SECONDS);
+            response = createWorkflowWithProvision(client(), template);
+        } else {
+            response = createWorkflowWithProvision(client(), template);
```
dbwiddis, line 376: This can be simplified by moving this creation outside the `if` conditional and only using it to wait on the ml config index.   

Which seems somewhat of a common step for almost all our tests, can we add it in the `waitToStart()` section with the `@Before` annotation, perhaps replacing my hacked "wait 25 seconds" workaround there?

src/test/java/org/opensearch/flowframework/rest/FlowFrameworkRestApiIT.java
```diff
@@ -363,6 +363,281 @@ public void testCreateAndProvisionAgentFrameworkWorkflow() throws Exception {
...
+        assertEquals(4, resourcesCreated.size());
+        resourceIds = resourcesCreated.stream().map(x -> x.workflowStepName()).collect(Collectors.toList());
+        assertTrue(resourceIds.contains("create_connector"));
+        assertTrue(resourceIds.contains("register_remote_model"));
+        assertTrue(resourceIds.contains("create_ingest_pipeline"));
+
+        // Retrieve pipeline by ID to ensure model ID is set correctly
+        String modelId = resourcesCreated.stream()
+            .filter(x -> x.workflowStepName().equals("register_remote_model"))
+            .findFirst()
```
dbwiddis, line 410: (Nit/Style) The resources created aren't really guaranteed to be sorted, so `findFirst()` seems to imply order matters.  If this is the only step with the resource, `findAny()` might work just as well. 

But while I love using streams, since you're doing two back-to-back iterations, this may be a case where a simple iteration over `resourcesCreated` will work, and it may make more sense to just string match on the resource type.

src/test/java/org/opensearch/flowframework/rest/FlowFrameworkRestApiIT.java
```diff
@@ -363,6 +363,281 @@ public void testCreateAndProvisionAgentFrameworkWorkflow() throws Exception {
...
+            .get()
+            .resourceId();
+        String pipelineId = resourcesCreated.stream()
+            .filter(x -> x.workflowStepName().equals("create_ingest_pipeline"))
+            .findFirst()
+            .get()
+            .resourceId();
+        GetPipelineResponse getPipelineResponse = getPipelines(pipelineId);
+        assertEquals(1, getPipelineResponse.pipelines().size());
+        assertTrue(getPipelineResponse.pipelines().get(0).getConfigAsMap().toString().contains(modelId));
```
dbwiddis, line 420: Is there a way to use the map key/value to extract the model ID rather than a string contains?

src/test/java/org/opensearch/flowframework/rest/FlowFrameworkRestApiIT.java
```diff
@@ -363,6 +363,281 @@ public void testCreateAndProvisionAgentFrameworkWorkflow() throws Exception {
...
+
+        // Reprovision template to add index which uses default ingest pipeline
+        template = TestHelpers.createTemplateFromFile("registerremotemodel-ingestpipeline-createindex.json");
+        response = reprovisionWorkflow(client(), workflowId, template);
+        assertEquals(RestStatus.CREATED, TestHelpers.restStatus(response));
+
+        resourcesCreated = getResourcesCreated(client(), workflowId, 10);
+        assertEquals(5, resourcesCreated.size());
+        resourceIds = resourcesCreated.stream().map(x -> x.workflowStepName()).collect(Collectors.toList());
```
dbwiddis, line 429: (Suggestion) if you collect into a map with workflow step name as the key and resource created as the value, you can just use the map later to look things up instead of continuing to iterate this list to find steps.

src/test/java/org/opensearch/flowframework/rest/FlowFrameworkRestApiIT.java
```diff
@@ -363,6 +363,281 @@ public void testCreateAndProvisionAgentFrameworkWorkflow() throws Exception {
...
+    }
+
+    public void testReprovisionWorkflowMidNodeAddition() throws Exception {
+        // Begin with a template to register a local pretrained model and create an index, no edges
+        Template template = TestHelpers.createTemplateFromFile("registerremotemodel-createindex.json");
+
+        // Hit Create Workflow API to create agent-framework template, with template validation check and provision parameter
+        Response response;
+        if (!indexExistsWithAdminClient(".plugins-ml-config")) {
+            assertBusy(() -> assertTrue(indexExistsWithAdminClient(".plugins-ml-config")), 40, TimeUnit.SECONDS);
```
dbwiddis, line 478: See above comment about moving this into the `@Before` method, perhaps replacing what's currently there.

dbwiddis (commented):
LGTM with a few suggestions.

All the resource matching lookups could probably be streamlined with a helper method producing a map.

src/test/java/org/opensearch/flowframework/rest/FlowFrameworkRestApiIT.java
joshpalis: Sure Ill make this modification

src/test/resources/template/registerremotemodel-createindex.json
```diff
@@ -0,0 +1,84 @@
+{
+    "name": "semantic search with local pretrained model",
+    "description": "Setting up semantic search, with a local pretrained embedding model",
+    "use_case": "SEMANTIC_SEARCH",
+    "version": {
```
amitgalitz, line 5: just wondering what happens on reprovisioning when we change version or the other fields not in workflows?

amitgalitz:
Can we add a test case where we reprovision a default use_case template?

src/test/resources/template/registerremotemodel-createindex.json
joshpalis: This will fail with message `Template does not contain any modifications`. We only check for modifications within the workflow nodes

src/test/java/org/opensearch/flowframework/rest/FlowFrameworkRestApiIT.java
joshpalis: Not directly, the model ID is part of the processor configuration string within the processors array field, so we would have to  iterate over the array and check for the string anyways. I think this is a simpler approach

src/test/java/org/opensearch/flowframework/rest/FlowFrameworkRestApiIT.java
```diff
@@ -48,15 +46,12 @@
...
-    private static AtomicBoolean waitToStart = new AtomicBoolean(true);
 
     @Before
     public void waitToStart() throws Exception {
         // ML Commons cron job runs every 10 seconds and takes 20+ seconds to initialize .plugins-ml-config index
-        // Delay on the first attempt for 25 seconds to allow this initialization and prevent flaky tests
-        if (waitToStart.getAndSet(false)) {
-            CountDownLatch latch = new CountDownLatch(1);
-            latch.await(25, TimeUnit.SECONDS);
+        if (!indexExistsWithAdminClient(".plugins-ml-config")) {
```
dbwiddis, line 53: Might be nice to use the ML Commons `CommonValue.ML_CONFIG_INDEX` constant here (static import).

src/test/java/org/opensearch/flowframework/workflow/UpdateIndexStepTests.java
```diff
@@ -136,6 +200,55 @@ public void testMissingSettings() throws InterruptedException {
...
+            getSettingsResponseListener.onResponse(new GetSettingsResponse(indexToSettings, indexToSettings));
+            return null;
+        }).when(indicesAdminClient).getSettings(any(), any());
+
+        // validate update settings request content
+        @SuppressWarnings({ "unchecked" })
+        ArgumentCaptor<UpdateSettingsRequest> updateSettingsRequestCaptor = ArgumentCaptor.forClass(UpdateSettingsRequest.class);
+
+        // Configurations has updated ingest pipeline default values of _none. Settings have regular and full names
+        String configurations =
```
dbwiddis, line 229: (Readability) This works, but in other tests I've seen us modify the first string with a regex search/replace.  Looking at this vs. line 154 I can't really visually tell what changed. I have to trust your comment (which is good that you included it, btw!)

dbwiddis (approved):
LGTM!   

Approved with suggestions, feel free to take them or ignore them.

amitgalitz (approved):

joshpalis:
Will raise the backport manually

Issue #668, opened 2024-04-18
[FEATURE] Add user level access control based on backend roles


amitgalitz:
#### What we have now:

If users want to send requests to the Flow-Framework plugin to create/get workflows, they should have appropriate permissions to perform the actions. The security plugin has two built-in roles to cover flow-framework use cases: flow_framework_read_access, and flow_framework_full_access.

### What solution would you like?

1. Add security-based access control so that only authorized users have access to workflows created/owned by other users. This feature is similar to AD/Alerting plugin resource control where a user can only access the detectors/monitors when at least one of the back-end roles of the user matches with that of the detectors/monitors.

2. Users are mapped to back-end roles based on which we filter the workflows/templates that can be accessed by them. Whenever a user sends an flow-framework API request, along with verifying if they have appropriate permissions, we also check them against authorized back-end roles. Each user can have more than one back-end role.

3. Back-end role-based access control enable an additional layer of security where a user can access only those workflows that are owned by same back-end role users.

4. We should have a setting to enabled filtering by backend role like other plugins have. 

We should follow exact same interface as AD has here : https://opensearch.org/docs/latest/observing-your-data/ad/security/#advanced-limit-access-by-backend-role

### Task Breakdown

In order to enable access control based on backend role filtering we will need to make modification in all APIs

In summary we want to be checking the backend role of the user making the request and comparing it to the backend role of the user that created the template. 

This feature will be enabled/disabled by a new setting: `filter_by_backend_roles`

#### Tasks
Tasks 1,2,3 can be done at the same time as task 4 and 5. (4, 5 don't depend on prior tasks)

1. Create common util methods to get a workflow and then compare if requested user has backend role required to access the resource
2. Block creation of workflows if user doesn’t have a backend role
3. Change following APIs with a check to see if user has permission to the resource using common method in task 1: update workflow, get workflow, provision workflow, get workflow status, deprovision workflow and delete workflow 
4. Create a common util method to add backend role for search requests (search workflows and search workflow state)
5. Update following API with additional backend role query in search request: search workflows and search workflow states
6. Security Integ tests for all APIs with backend roles considered
7. BWC tests if we make changes to the system index

### Additional Consideration:

An additional request that customers have asked for in the past include the ability to change the backend role configured with a particular resource after creation. This is something that hasn’t been implemented in AD but has been implemented in Alerting and ML-Commons through model groups. 

[...]

PR #838, opened 2024-08-16
Added user level access control based on backend roles

owaiskazi19:
### Description
Added user level access control based on backend roles.

1. Created common util methods to get a workflow and then compare if requested user has backend role required to access the resource
2. Blocked creation of workflows if user doesn’t have a backend role
3. Changed following APIs with a check to see if user has permission to the resource using common method in task 1: update workflow, get workflow, provision workflow, get workflow status, reprovision workflow, deprovision workflow and delete workflow
4. Created a common util method to add backend role for search requests (search workflows and search workflow state)
5. Updated following API with additional backend role query in search request: search workflows and search workflow states
[**For this task, I had to change the search queries to boolean to add a nested filter query**] @dbwiddis any thoughts?
7. Added Security Integaration tests for all APIs with backend roles considered (except reprovision)
8. Documentation PR - https://github.com/opensearch-project/documentation-website/pull/8076

Learning after working on this feature
1. How access role filtering is done on OpenSearch?
2. Played a lot with ThreadContext
3. How user level permissions are defined?
4. Writing Security integration tests 

Refer https://github.com/opensearch-project/flow-framework/issues/668#issuecomment-2294020696 for step by step workflow and `src/test/java/org/opensearch/flowframework/rest/FlowFrameworkSecureRestApiIT.java` file for complete functionality.

### Related Issues
Resolves #https://github.com/opensearch-project/flow-framework/issues/668

### Check List


src/main/java/org/opensearch/flowframework/transport/DeprovisionWorkflowTransportAction.java
```diff
@@ -86,51 +98,79 @@ public DeprovisionWorkflowTransportAction(
...
+        WorkflowRequest request,
+        ActionListener<WorkflowResponse> listener,
+        ThreadContext.StoredContext context
+    ) {
+        String workflowId = request.getWorkflowId();
+        String allowDelete = request.getParams().get(ALLOW_DELETE);
+        GetWorkflowStateRequest getStateRequest = new GetWorkflowStateRequest(workflowId, true);
+        logger.info("Querying state for workflow: {}", workflowId);
+        client.execute(GetWorkflowStateAction.INSTANCE, getStateRequest, ActionListener.wrap(response -> {
+            context.restore();
```
dbwiddis, line 154: Not sure the context is restored in the exception path here?  (We're relying on the try-with-resource wrapping this call.)   Consider using `runBefore()`.

src/main/java/org/opensearch/flowframework/util/ParseUtils.java
```diff
@@ -229,6 +247,218 @@ public static User getUserContext(Client client) {
...
+     * @param listener action listener
+     * @param function workflow function
+     * @param client node client
+     * @param clusterService cluster service
+     * @param xContentRegistry contentRegister to parse get response
+     */
+    public static void resolveUserAndExecute(
+        User requestedUser,
+        String workflowId,
+        boolean filterByEnabled,
```
dbwiddis, line 294: This unboxes the `Boolean` which we're pretty sure will never be null, otherwise would NPE here.  But we may as well just keep it boxed here to save a few nanoseconds.

src/main/java/org/opensearch/flowframework/transport/GetWorkflowTransportAction.java
```diff
@@ -103,4 +115,38 @@ protected void doExecute(Task task, WorkflowRequest request, ActionListener<GetW
...
+    private void executeGetRequest(
+        WorkflowRequest request,
+        ActionListener<GetWorkflowResponse> listener,
+        ThreadContext.StoredContext context
+    ) {
+        String workflowId = request.getWorkflowId();
+        GetRequest getRequest = new GetRequest(GLOBAL_CONTEXT_INDEX, workflowId);
+        logger.info("Querying workflow from global context: {}", workflowId);
+        client.get(getRequest, ActionListener.wrap(response -> {
+            context.restore();
```
dbwiddis, line 134: Similar comment here about exception path/runBefore.

src/main/java/org/opensearch/flowframework/transport/ProvisionWorkflowTransportAction.java
```diff
@@ -95,108 +111,144 @@ public ProvisionWorkflowTransportAction(
...
+    private void executeProvisionRequest(
+        WorkflowRequest request,
+        ActionListener<WorkflowResponse> listener,
+        ThreadContext.StoredContext context
+    ) {
+        String workflowId = request.getWorkflowId();
+        GetRequest getRequest = new GetRequest(GLOBAL_CONTEXT_INDEX, workflowId);
+        logger.info("Querying workflow from global context: {}", workflowId);
+        client.get(getRequest, ActionListener.wrap(response -> {
+            context.restore();
```
dbwiddis, line 169: Similar comment here about exception path/runBefore.

src/main/java/org/opensearch/flowframework/transport/ReprovisionWorkflowTransportAction.java
```diff
@@ -107,120 +123,149 @@ public ReprovisionWorkflowTransportAction(
...
+        ReprovisionWorkflowRequest request,
+        ActionListener<WorkflowResponse> listener,
+        ThreadContext.StoredContext context
+    ) {
+        String workflowId = request.getWorkflowId();
+        logger.info("Querying state for workflow: {}", workflowId);
+        // Retrieve state and resources created
+        GetWorkflowStateRequest getStateRequest = new GetWorkflowStateRequest(workflowId, true);
+        client.execute(GetWorkflowStateAction.INSTANCE, getStateRequest, ActionListener.wrap(response -> {
+            context.restore();
```
dbwiddis, line 173: Similar comment here about exception path/runBefore.

src/main/java/org/opensearch/flowframework/transport/handler/SearchHandler.java
```diff
@@ -0,0 +1,98 @@
...
+import static org.opensearch.flowframework.util.ParseUtils.isAdmin;
+import static org.opensearch.flowframework.util.RestHandlerUtils.getSourceContext;
+
+/**
+ * Handle general search request, check user role and return search response.
+ */
+public class SearchHandler {
+    private final Logger logger = LogManager.getLogger(SearchHandler.class);
+    private final Client client;
+    private volatile Boolean filterEnabled;
```
dbwiddis, line 34: would prefer `filterByBackendRole` name here

src/main/java/org/opensearch/flowframework/transport/handler/WorkflowFunction.java
```diff
@@ -0,0 +1,22 @@
...
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+package org.opensearch.flowframework.transport.handler;
+
+/**
+ * Interface for a workflow function.
+ *
+ */
+public interface WorkflowFunction {
```
dbwiddis, line 15: What does this interface give us that we don't already have with [`Runnable`](https://docs.oracle.com/javase/8/docs/api/java/lang/Runnable.html)?

src/main/java/org/opensearch/flowframework/transport/handler/WorkflowFunction.java
```diff
@@ -0,0 +1,22 @@
...
+    /**
+     * Performs this operation.
+     *
+     * Notes: don't forget to send back responses via channel if you process response with this method.
```
dbwiddis, line 19: This isn't clear what it means. Which channel since it's not passed as a parameter?  Should this be a `Consumer<Channel>`?

src/main/java/org/opensearch/flowframework/transport/CreateWorkflowTransportAction.java
```diff
@@ -82,20 +96,92 @@ public CreateWorkflowTransportAction(
...
+        User requestedUser,
+        String workflowId,
+        ActionListener<WorkflowResponse> listener,
+        WorkflowFunction function
+    ) {
+        try {
+            // Check if user has backend roles
+            // When filter by is enabled, block users creating/updating workflows who do not have backend roles.
+            if (filterByEnabled) {
+                String error = checkFilterByBackendRoles(requestedUser);
```
dbwiddis, line 146: Doesn't look like tests are covering the `filterByEnabled` case.

src/main/java/org/opensearch/flowframework/transport/GetWorkflowStateTransportAction.java
```diff
@@ -44,60 +48,96 @@ public class GetWorkflowStateTransportAction extends HandledTransportAction<GetW
...
+            if (r != null && r.isExists()) {
+                try (XContentParser parser = ParseUtils.createXContentParserFromRegistry(xContentRegistry, r.getSourceAsBytesRef())) {
+                    ensureExpectedToken(XContentParser.Token.START_OBJECT, parser.nextToken(), parser);
+                    WorkflowState workflowState = WorkflowState.parse(parser);
+                    listener.onResponse(new GetWorkflowStateResponse(workflowState, request.getAll()));
+                } catch (Exception e) {
+                    String errorMessage = "Failed to parse workflowState: " + r.getId();
+                    logger.error(errorMessage, e);
+                    listener.onFailure(new FlowFrameworkException(errorMessage, RestStatus.BAD_REQUEST));
+                }
```
dbwiddis, line 129: These aren't covered by tests.  Yeah, probably were missing coverage before you refactored, but we can add them now!

src/main/java/org/opensearch/flowframework/util/ParseUtils.java
```diff
@@ -229,6 +247,218 @@ public static User getUserContext(Client client) {
...
+        String userFieldName = "user";
+        String userBackendRoleFieldName = "user.backend_roles.keyword";
+        List<String> backendRoles = user.getBackendRoles() != null ? user.getBackendRoles() : ImmutableList.of();
+        // For normal case, user should have backend roles.
+        TermsQueryBuilder userRolesFilterQuery = QueryBuilders.termsQuery(userBackendRoleFieldName, backendRoles);
+        NestedQueryBuilder nestedQueryBuilder = new NestedQueryBuilder(userFieldName, userRolesFilterQuery, ScoreMode.None);
+        boolQueryBuilder.must(nestedQueryBuilder);
+        QueryBuilder query = searchSourceBuilder.query();
+        if (query == null) {
+            searchSourceBuilder.query(boolQueryBuilder);
```
dbwiddis, line 270: Where do we test this searchSourceBuilder?

src/main/java/org/opensearch/flowframework/util/ParseUtils.java
```diff
@@ -229,6 +247,218 @@ public static User getUserContext(Client client) {
...
+        } else if (query instanceof BoolQueryBuilder) {
+            ((BoolQueryBuilder) query).filter(boolQueryBuilder);
+        } else {
+            // e.g., wild card query
+            throw new FlowFrameworkException("Search API does not support queries other than BoolQuery", RestStatus.BAD_REQUEST);
```
dbwiddis, line 275: Why not?  We can wrap an existing query of any type in a BoolQuery `must()` and then add a filter.  See [example](https://github.com/opensearch-project/ml-commons/blob/9f4d2ce47c1f53b2003733b91f37a1246182e1ff/common/src/main/java/org/opensearch/sdk/client/LocalClusterIndicesClient.java#L227-L230)

src/main/java/org/opensearch/flowframework/util/ParseUtils.java
```diff
@@ -229,6 +247,218 @@ public static User getUserContext(Client client) {
...
+        }
+        return false;
+    }
+
+    /**
+     * Check if filter by backend roles is enabled and validate the requested user
+     * @param requestedUser the user to execute the request
+     * @return error message if validation fails, null otherwise
+     */
+    public static String checkFilterByBackendRoles(User requestedUser) {
```
dbwiddis, line 349: Returning a `String` or `null` and then testing that null to throw an exception seems an odd pattern.

Why not just have a void method here that throws an exception?  Most validate-style methods do this.  That would allow you to customize the RestStatus as well (403 vs. 404, etc.)

dbwiddis (commented):
Some suggestions.

> [For this task, I had to change the search queries to boolean to add a nested filter query] @dbwiddis any thoughts?

See comment below.  You can wrap any query inside a `BoolQuery` with a `must()` and then add a filter.  I linked to an example I just wrote last week.

src/main/java/org/opensearch/flowframework/transport/DeprovisionWorkflowTransportAction.java
dbwiddis: > It should be. Since it was passed from a try/catch block above. 

Right, the try-with-resources block will close it it but it seems inconsistent with explicitly restoring it here on this path.  Mostly just a consistency/readability issue.

src/main/java/org/opensearch/flowframework/util/ParseUtils.java
owaiskazi19: Yes, this looks better with no hits
```json
{
    "took": 11,
    "timed_out": false,
    "_shards": {
        "total": 1,
        "successful": 1,
        "skipped": 0,
        "failed": 0
    },
    "hits": {
        "total": {
            "value": 0,
            "relation": "eq"
        },
        "max_score": null,
        "hits": []
    }
}
```e

File: src/main/java/org/opensearch/flowframework/transport/handler/WorkflowFunction.java (hunk 8 above)
Comment (owaiskazi19) [reply src/main/java/org/opensearch/flowframework/transport/handler/WorkflowFunction.java]: Removed the interface itself

File: src/main/java/org/opensearch/flowframework/transport/CreateWorkflowTransportAction.java (hunk 9 above)
Comment (owaiskazi19) [reply src/main/java/org/opensearch/flowframework/transport/CreateWorkflowTransportAction.java]: We are now in the latest commit )

File: src/main/java/org/opensearch/flowframework/transport/GetWorkflowStateTransportAction.java (hunk 10 above)
Comment (owaiskazi19) [reply src/main/java/org/opensearch/flowframework/transport/GetWorkflowStateTransportAction.java:129]: Added couple of tests

File: src/main/java/org/opensearch/flowframework/util/ParseUtils.java (hunk 11 above)
Comment (owaiskazi19) [reply src/main/java/org/opensearch/flowframework/util/ParseUtils.java:268]: In the latest commit :p

File: src/main/java/org/opensearch/flowframework/transport/CreateWorkflowTransportAction.java [hunk 14]
```diff
@@ -82,20 +95,94 @@ public CreateWorkflowTransportAction(
...
+    private void resolveUserAndExecute(
+        User requestedUser,
+        String workflowId,
+        ActionListener<WorkflowResponse> listener,
+        Runnable function
+    ) {
+        try {
+            // Check if user has backend roles
+            // When filter by is enabled, block users creating/updating workflows who do not have backend roles.
+            if (filterByEnabled) {
```
dbwiddis, line 144: (Style) This unboxes the `Boolean`.  I happened to follow the code trail enough to know that the setting will return its default (false) value if it's not present and thus not be null, but that's implementation details I shouldn't have had to look at.  See https://sonarsource.atlassian.net/browse/RSPEC-5411

src/main/java/org/opensearch/flowframework/util/ParseUtils.java
```diff
@@ -225,10 +241,225 @@ public static Instant parseInstant(XContentParser parser) throws IOException {
...
+        String workflowId,
+        Boolean filterByEnabled,
+        ActionListener listener,
+        Runnable function,
+        Client client,
+        ClusterService clusterService,
+        NamedXContentRegistry xContentRegistry
+    ) {
+        try {
+            if (requestedUser == null || !filterByEnabled) {
```
dbwiddis, line 302: Similar comment here on `filterByEnabled` Boolean.

src/main/java/org/opensearch/flowframework/transport/CreateWorkflowTransportAction.java
```diff
@@ -63,6 +69,10 @@ public class CreateWorkflowTransportAction extends HandledTransportAction<Workfl
     private final Client client;
     private final FlowFrameworkSettings flowFrameworkSettings;
     private final PluginsService pluginsService;
+    private volatile Boolean filterByEnabled;
+    private final Settings settings;
```
dbwiddis, line 73: (Nit) You never use this value.

src/main/java/org/opensearch/flowframework/transport/handler/SearchHandler.java
```diff
@@ -0,0 +1,98 @@
...
+            // Case 1: user == null when 1. Security is disabled. 2. When user is super-admin
+            // Case 2: If Security is enabled and filter is disabled, proceed with search as
+            // user is already authenticated to hit this API.
+            // case 3: user is admin which means we don't have to check backend role filtering
+            client.search(request, ActionListener.runBefore(listener, context::restore));
+        } else {
+            // Security is enabled, filter is enabled and user isn't admin
+            try {
+                ParseUtils.addUserBackendRolesFilter(user, request.source());
+                logger.debug("Filtering result by " + user.getBackendRoles());
```
dbwiddis, line 91: (Nit)
```suggestion
                logger.debug("Filtering result by {}", user.getBackendRoles());
```

src/main/java/org/opensearch/flowframework/util/ParseUtils.java
```diff
@@ -225,10 +241,225 @@ public static Instant parseInstant(XContentParser parser) throws IOException {
...
+     * @param function workflow function
+     * @param client node client
+     * @param clusterService cluster service
+     * @param xContentRegistry contentRegister to parse get response
+     */
+    public static void resolveUserAndExecute(
+        User requestedUser,
+        String workflowId,
+        Boolean filterByEnabled,
+        ActionListener listener,
```
dbwiddis, line 295: (Nit)
```suggestion
        ActionListener<? extends ActionResponse> listener,
```

dbwiddis (approved):
LGTM with some style nits.

src/main/java/org/opensearch/flowframework/util/ParseUtils.java
```diff
@@ -225,10 +242,225 @@ public static Instant parseInstant(XContentParser parser) throws IOException {
...
+     * @param requestUser the user to execute the request
+     * @param workflowId workflow id
+     * @param filterByEnabled filter by enabled setting
+     * @param listener action listener
+     * @param function workflow function
+     * @param client node client
+     * @param clusterService cluster service
+     * @param xContentRegistry contentRegister to parse get response
+     */
+    public static void getWorkflow(
```
joshpalis, line 376: Non-blocking, but seems to me these methods should live outside of the ParseUtils class, since they dont have anything to do with parsing.

joshpalis (approved):
Overall looks good to me, just some non-blocking comments. Thanks @owaiskazi19 for add this in

src/main/java/org/opensearch/flowframework/util/ParseUtils.java
owaiskazi19: Agreed. We don't actually parse here except the template but I wanted to keep it in a common file. Probably, will need another one of such.

Issue #780, opened 2024-07-12
[FEATURE] Update WorkflowState resources during deprovisioning


dbwiddis:
### Is your feature request related to a problem?

Because WorkflowState updates have only been additive, the Deprovisioning workflow only updates once at the end of deprovisioning with a "remaining resources" list.  There is no check whether this was successful (see #691) and no concurrency guarantees.

### What solution would you like?

Following implementation of #779, use the iterative update to remove resources from the resource list as they are deprovisioned, so the final result will not be sent until the state updates were all completed.

dbwiddis:
This should be a simple change to the [`ToolStep`](https://github.com/opensearch-project/flow-framework/blob/43a9f871de50fc57758efd57c0f8899b22cecc3a/src/main/java/org/opensearch/flowframework/workflow/ToolStep.java#L51-L94) class adding the new optional parameter.
 - please also update the test case
 - please also migrate the input/output set per https://github.com/opensearch-project/flow-framework/issues/535

PR #894, opened 2024-10-02
Update workflow state without using painless script 

### LINKED ISSUE #779: [REFACTOR] Improve Workflow State Resource updates

Currently the Workflow State resource document is only additive, using a Painless script during provisioning to append to a list.  This prevented iteratively updating it during deprovisioning, leading to open questions like #691.

Following #778 we will have the ability to update any individual element of the Workflow State which will bring these benefits:
 - Allow for incremental addition to the resources created without using Painless, improving portability
 - Allow for incremental removal from the resources created during deprovisioning, allowing the full deprovisioning to depend on this update. (See #790) 
 - Allow for the easy removal of the "error" field on subsequent successful provisioning

Replace the painless script with appropriate code from #778.  Use `ifSeqNo()` and `ifPrimaryTerm()` on the update requests to enforce concurrency, and retry if it fails.

dbwiddis:
### Description

Removes methods using painless script for workflow state updates.

### Related Issues

Fixes #779 (except for the part that will be covered by #780)

### Check List

### DIFF
--- a/src/main/java/org/opensearch/flowframework/indices/FlowFrameworkIndicesHandler.java
+++ b/src/main/java/org/opensearch/flowframework/indices/FlowFrameworkIndicesHandler.java
@@ -691,7 +691,7 @@ public class FlowFrameworkIndicesHandler {
                 getAndUpdateResourceInStateDocumentWithRetries(
                     workflowId,
                     newResource,
-                    3,
+                    5,
                     ActionListener.runBefore(listener, context::restore)
                 );
             }
--- a/src/main/java/org/opensearch/flowframework/indices/FlowFrameworkIndicesHandler.java
+++ b/src/main/java/org/opensearch/flowframework/indices/FlowFrameworkIndicesHandler.java
@@ -82,6 +82,8 @@ public class FlowFrameworkIndicesHandler {
     private static final Map<String, AtomicBoolean> indexMappingUpdated = new HashMap<>();
     private static final Map<String, Object> indexSettings = Map.of("index.auto_expand_replicas", "0-1");
     private final NamedXContentRegistry xContentRegistry;
+    // Retries in case of simultaneous updates
+    private static final int RETRIES = 5;
 
     /**
      * constructor
@@ -599,7 +601,7 @@ public class FlowFrameworkIndicesHandler {
                 updatedDocument.toXContent(builder, null);
                 updateRequest.doc(builder);
                 updateRequest.setRefreshPolicy(WriteRequest.RefreshPolicy.IMMEDIATE);
-                updateRequest.retryOnConflict(5);
+                updateRequest.retryOnConflict(RETRIES);
                 client.update(updateRequest, ActionListener.runBefore(listener, context::restore));
             } catch (Exception e) {
                 String errorMessage = "Failed to update " + WORKFLOW_STATE_INDEX + " entry : " + documentId;
@@ -630,7 +632,7 @@ public class FlowFrameworkIndicesHandler {
                 Map<String, Object> updatedContent = new HashMap<>(updatedFields);
                 updateRequest.doc(updatedContent);
                 updateRequest.setRefreshPolicy(WriteRequest.RefreshPolicy.IMMEDIATE);
-                updateRequest.retryOnConflict(5);
+                updateRequest.retryOnConflict(RETRIES);
                 // TODO: decide what condition can be considered as an update conflict and add retry strategy
                 client.update(updateRequest, ActionListener.runBefore(listener, context::restore));
             } catch (Exception e) {
@@ -691,7 +693,7 @@ public class FlowFrameworkIndicesHandler {
                 getAndUpdateResourceInStateDocumentWithRetries(
                     workflowId,
                     newResource,
-                    5,
+                    RETRIES,
                     ActionListener.runBefore(listener, context::restore)
                 );
             }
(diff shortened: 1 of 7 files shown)


src/main/java/org/opensearch/flowframework/indices/FlowFrameworkIndicesHandler.java
```diff
@@ -716,26 +679,93 @@ public void addResourceToStateIndex(
...
+        if (!doesIndexExist(WORKFLOW_STATE_INDEX)) {
+            String errorMessage = "Failed to update state for " + workflowId + " due to missing " + WORKFLOW_STATE_INDEX + " index";
+            logger.error(errorMessage);
+            listener.onFailure(new FlowFrameworkException(errorMessage, RestStatus.NOT_FOUND));
+        } else {
+            try (ThreadContext.StoredContext context = client.threadPool().getThreadContext().stashContext()) {
+                getAndUpdateResourceInStateDocumentWithRetries(
+                    workflowId,
+                    newResource,
+                    3,
```
amitgalitz, line 694: Is 3 retries a standard? we use 5 in other locations

joshpalis (approved):

src/main/java/org/opensearch/flowframework/indices/FlowFrameworkIndicesHandler.java
```diff
@@ -716,26 +679,93 @@ public void addResourceToStateIndex(
...
+    private void handleStateUpdateException(
+        String workflowId,
+        ResourceCreated newResource,
+        int retries,
+        ActionListener<WorkflowData> listener,
+        Exception e
+    ) {
+        if (e instanceof VersionConflictEngineException && retries > 0) {
+            // Retry if we haven't exhausted retries
+            getAndUpdateResourceInStateDocumentWithRetries(workflowId, newResource, retries - 1, listener);
```
owaiskazi19, line 757: Why this recursive calling here when we can pass retryonconflict?

src/test/java/org/opensearch/flowframework/indices/FlowFrameworkIndicesHandlerTests.java
```diff
@@ -537,6 +606,147 @@ public void testAddResourceToStateIndex() throws IOException {
...
+        }).doAnswer(invocation -> {
+            ActionListener<UpdateResponse> responseListener = invocation.getArgument(1);
+            responseListener.onFailure(conflictException);
+            return null;
+        }).doAnswer(invocation -> {
+            // we'll never get here
+            ActionListener<UpdateResponse> responseListener = invocation.getArgument(1);
+            responseListener.onResponse(updateResponse);
+            return null;
+        }).when(client).update(any(UpdateRequest.class), any());
```
owaiskazi19, line 735: This is great. We should move this to a common test class for other tests to use retry listeners

owaiskazi19 (commented):
LGTM with minor comments

src/main/java/org/opensearch/flowframework/indices/FlowFrameworkIndicesHandler.java
```diff
@@ -608,99 +610,60 @@ public void updateFlowFrameworkSystemIndexDoc(
...
         } else {
             try (ThreadContext.StoredContext context = client.threadPool().getThreadContext().stashContext()) {
-                DeleteRequest deleteRequest = new DeleteRequest(WORKFLOW_STATE_INDEX, documentId);
-                deleteRequest.setRefreshPolicy(WriteRequest.RefreshPolicy.IMMEDIATE);
-                client.delete(deleteRequest, ActionListener.runBefore(listener, context::restore));
+                UpdateRequest updateRequest = new UpdateRequest(WORKFLOW_STATE_INDEX, documentId);
+                Map<String, Object> updatedContent = new HashMap<>(updatedFields);
+                updateRequest.doc(updatedContent);
+                updateRequest.setRefreshPolicy(WriteRequest.RefreshPolicy.IMMEDIATE);
+                updateRequest.retryOnConflict(5);
```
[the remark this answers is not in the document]
dbwiddis: I don't know why they are different and I don't think conflicts are going to be common.  These are updates to a single workflow's state index, and the only time there would be a conflict is if two (or more) resources are simultaneously trying to update the resource list.  I'd have to think this is an extremely rare situation, if at all.  Sure, I can make it a constant.

src/main/java/org/opensearch/flowframework/indices/FlowFrameworkIndicesHandler.java
dbwiddis: RetryOnConflict only applies when you don't care what the existing value is but you're updating a field.  In our case we need to fetch the existing resource list, add to it, and then put that resource list back.  

If the document changed in between us getting and updating, retryOnConflict would miss it.  We use the `seqNo` and `primaryTerm` from the document we "got" to make sure that's the same document we're putting back in; if the doc has been updated, retrying won't do us any good, we need to retry the whole get-and-put which is what this does.

Also a minor note, while it looks recursive, it's just a chained callback not unlike all our other transport calls, so it won't fill up the stack.

src/test/java/org/opensearch/flowframework/indices/FlowFrameworkIndicesHandlerTests.java
dbwiddis: Not sure what you mean here?

owaiskazi19: I was thinking if we can have a single method which can be used across tests to test retry listeners but I don't think so there would be more case. So it should be fine

src/main/java/org/opensearch/flowframework/indices/FlowFrameworkIndicesHandler.java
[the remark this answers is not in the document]
owaiskazi19: Should we update the `3` also with the constant now?

owaiskazi19 (approved):

amitgalitz (approved):

PR #898, opened 2024-10-03
Incrementally remove resources from workflow state during deprovisioning

### LINKED ISSUE #691: [DISCUSS] How should we handle a successful deprovision with a failure in workflow state update
Coming from https://github.com/opensearch-project/flow-framework/pull/689#discussion_r1585173319

When deprovisioning, successfully deleting all resources gives a successful response to the user.

There is a possibility of an unexpected failure resetting the state document to NOT_STARTED (if template exists) or deleting it (if template doesn't exist).   Presently this state index failure is only logged.

Keep this status quo, as the failures are extremely unlikely to occur, and can be corrected with another deprovisioning.  

Waiting to return a result to a user until the state document is updated, and returning a more verbose error message of the form "resources were successfully deprovisioned, but the workflow state update failed, try deprovisioning again.  This may or may not succeed and may be confusing.

### LINKED ISSUE #780: [FEATURE] Update WorkflowState resources during deprovisioning

dbwiddis:
### Description

Removes resources from the workflow state list during deprovision incrementally.  Retains the existing full-list update as a backup option.

### Related Issues

Resolves #780
Resolves #691 

### Check List

### DIFF
--- a/src/main/java/org/opensearch/flowframework/indices/FlowFrameworkIndicesHandler.java
+++ b/src/main/java/org/opensearch/flowframework/indices/FlowFrameworkIndicesHandler.java
@@ -11,6 +11,7 @@ package org.opensearch.flowframework.indices;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.opensearch.ExceptionsHelper;
+import org.opensearch.action.DocWriteRequest.OpType;
 import org.opensearch.action.admin.indices.create.CreateIndexRequest;
 import org.opensearch.action.admin.indices.create.CreateIndexResponse;
 import org.opensearch.action.admin.indices.mapping.put.PutMappingRequest;
@@ -693,6 +694,7 @@ public class FlowFrameworkIndicesHandler {
                 getAndUpdateResourceInStateDocumentWithRetries(
                     workflowId,
                     newResource,
+                    OpType.INDEX,
                     RETRIES,
                     ActionListener.runBefore(listener, context::restore)
                 );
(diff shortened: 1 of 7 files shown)


src/main/java/org/opensearch/flowframework/indices/FlowFrameworkIndicesHandler.java
```diff
@@ -732,41 +764,51 @@ private void getAndUpdateResourceInStateDocumentWithRetries(
...
-            + " on step "
-            + newResource.workflowStepId()
-            + " with "
-            + newResource.resourceType()
-            + " "
-            + newResource.resourceId();
+        StringBuilder sb = new StringBuilder("Failed to update workflow state for ");
+        sb.append(workflowId).append(" on step ").append(newResource.workflowStepId());
+        sb.append(" to ").append(operation.equals(OpType.DELETE) ? "delete" : "add");
+        sb.append(" resource ").append(newResource.resourceType()).append(" ").append(newResource.resourceId());
```
owaiskazi19, line 810:
```suggestion
        String message = String.format(
    "Failed to update workflow state for %s on step %s to %s resource %s %s",
    workflowId,
    newResource.workflowStepId(),
    operation.equals(OpType.DELETE) ? "delete" : "add",
    newResource.resourceType(),
    newResource.resourceId()
);
```

owaiskazi19 (approved):

src/main/java/org/opensearch/flowframework/indices/FlowFrameworkIndicesHandler.java
dbwiddis: @owaiskazi19 `String.format()` has to parse the format string which [takes a couple dozen milliseconds](https://stackoverflow.com/a/44124045/1161484).  Granted this will be a rarely-encountered code path, but in addition to the added time, `String.format()` has to deal with locale (we have to specify locale or accept unpredictable output strings that can cause flaky tests), and can be a source of additional exceptions if you use the wrong format string, so I really like to avoid using it in exception catch blocks (which this `handle()` method is in).

While there's negligible difference between concatenation vs. `StringBuilder` I thought the way I separated the lines here was more readable than the concatenation version I replaced with it. 

The log message substitution stuff also works, but unfortunately since we're using the message both in an exception and log message, we have to do this stuff.

All that said, it looks like I can use the `ParameterizedMessageFactory` to construct this message in log4j2 style and then re-use it twice.   Which we should also do elsewhere in this repo...

dbwiddis: Implemented in https://github.com/opensearch-project/flow-framework/pull/898/commits/86ac8ea862ba59abcbdf036fd637159e751e3be4 and opened an issue to do elsewhere.

amitgalitz (approved):

Issue #651, opened 2024-04-06
[FEATURE] Add test(s) to compare workflow steps configuration to API specification


dbwiddis:
### Is your feature request related to a problem?

Workflow Steps are tightly coupled to the APIs they implement.   

The APIs are hosted in external repositories, and there is not signal to our plugin that an API has added features.

OpenSearch APIs have OpenAPI spec files written for them: https://github.com/opensearch-project/opensearch-api-specification

I have opened an issue on ML Commons requesting they do the same: https://github.com/opensearch-project/ml-commons/issues/2297

We should establish automated testing that runs on a periodic basis (Scheduled GitHub Action runs) to test whether we have implemented all API parameters.

### What solution would you like?

1. Write code to remotely fetch the API specification from wherever it resides.  For example, OpenSearch APIs are published here: https://github.com/opensearch-project/opensearch-api-specification/tree/main/spec .  It may be acceptable to fetch these file via wget in the GHA script to make it available locally, but if the effort to use a testing client (like we do in integ tests) is low, that would be preferable.
2. Parse the YAML into an object.  This may be complex: for example, creating an index [references a request body with a path](https://github.com/opensearch-project/opensearch-api-specification/blob/18795533eac735276a141703c8f7b5de9bc0ab35/spec/namespaces/_core.yaml#L1065-L1066) that has to be looked up [elsewhere](https://github.com/opensearch-project/opensearch-api-specification/blob/18795533eac735276a141703c8f7b5de9bc0ab35/spec/namespaces/indices.yaml#L1717-L1732).
3. Use the [WorkflowSteps](https://github.com/opensearch-project/flow-framework/blob/f6e90362f175e3439e902bd440df3f84dbcd6410/src/main/java/org/opensearch/flowframework/workflow/WorkflowStepFactory.java#L120) enum to compare the schema/content/requestBody definitions with the required and optional inputs.  The enum may need to be updated to include other information for comparison.

### What alternatives have you considered?

Manually following change logs, documentation, and PRs

### Do you have any additional context?

It may not be necessary to go deeply into object parsing; just the first level presence or absence of an object key in the request body is likely sufficient.

PR #900, opened 2024-10-04
Add ApiSpecFetcher for Fetching and Comparing API Specifications

junweid62:
### Description

- Introduced a new `ApiSpecFetcher` utility to programmatically fetch and compare OpenAPI specifications.

- Supports comparing required input fields between an OpenAPI spec and predefined enum values.

- The current implementation serves as a foundational mechanism, with the comparison logic successfully integrated into the `Register Agent` workflow.

- Future PRs will extend this mechanism to other workflow steps, ensuring consistent validation across the API specification.

### Related Issues
Part of #651

### Check List
- [ ] Commits are signed per the DCO using `--signoff`.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -179,6 +179,15 @@ dependencies {
     implementation "org.glassfish:jakarta.json:2.0.1"
     implementation "org.eclipse:yasson:3.0.4"
     implementation "com.google.code.gson:gson:2.11.0"
+    // Swagger-Parser dependencies for API consistency tests
+    implementation 'io.swagger.core.v3:swagger-models:2.2.23'
+    implementation 'io.swagger.parser.v3:swagger-parser-core:2.1.22'
+    implementation 'io.swagger.parser.v3:swagger-parser:2.1.22'
+    implementation 'io.swagger.parser.v3:swagger-parser-v3:2.1.22'
+    implementation 'com.fasterxml.jackson.core:jackson-databind:2.17.0'
+    implementation 'com.fasterxml.jackson.datatype:jackson-datatype-jsr310:2.13.0'
+    implementation 'com.fasterxml.jackson.core:jackson-annotations:2.17.0'
+    implementation 'io.swagger.core.v3:swagger-core:2.2.21'
 
     // ZipArchive dependencies used for integration tests
     zipArchive group: 'org.opensearch.plugin', name:'opensearch-ml-plugin', version: "${opensearch_build}"
--- a/build.gradle
+++ b/build.gradle
@@ -33,7 +33,7 @@ buildscript {
         bwcFilePath = "src/test/resources/org/opensearch/flowframework/bwc/"
         bwcFlowFrameworkPath = bwcFilePath + "flowframework/"
 
-        isSameMajorVersion = opensearch_version.split("\\.")[0] == bwcVersionShort.split("\\.")
+        isSameMajorVersion = opensearch_version.split("\\.")[0] == bwcVersionShort.split("\\.")[0]
         swaggerVersion = "2.1.22"
         jacksonVersion = "2.18.0"
         swaggerCoreVersion = "2.2.23"
--- a/build.gradle
+++ b/build.gradle
@@ -24,7 +24,7 @@ buildscript {
         opensearch_no_snapshot = opensearch_build.replace("-SNAPSHOT","")
         System.setProperty('tests.security.manager', 'false')
         common_utils_version = System.getProperty("common_utils.version", opensearch_build)
-
+        swaggerCoreVersion = "2.2.23"
         bwcVersionShort = "2.12.0"
         bwcVersion = bwcVersionShort + ".0"
         bwcOpenSearchFFDownload = 'https://ci.opensearch.org/ci/dbc/distribution-build-opensearch/' + bwcVersionShort + '/latest/linux/x64/tar/builds/' +
@@ -33,7 +33,11 @@ buildscript {
         bwcFilePath = "src/test/resources/org/opensearch/flowframework/bwc/"
         bwcFlowFrameworkPath = bwcFilePath + "flowframework/"
 
-        isSameMajorVersion = opensearch_version.split("\\.")[0] == bwcVersionShort.split("\\.")[0]
+        isSameMajorVersion = opensearch_version.split("\\.")[0] == bwcVersionShort.split("\\.")
+        swaggerVersion = "2.1.22"
+        jacksonVersion = "2.18.0"
+        swaggerCoreVersion = "2.2.23"
+
     }
 
     repositories {
@@ -180,14 +184,14 @@ dependencies {
     implementation "org.eclipse:yasson:3.0.4"
     implementation "com.google.code.gson:gson:2.11.0"
     // Swagger-Parser dependencies for API consistency tests
-    implementation 'io.swagger.core.v3:swagger-models:2.2.23'
-    implementation 'io.swagger.parser.v3:swagger-parser-core:2.1.22'
-    implementation 'io.swagger.parser.v3:swagger-parser:2.1.22'
-    implementation 'io.swagger.parser.v3:swagger-parser-v3:2.1.22'
-    implementation 'com.fasterxml.jackson.core:jackson-databind:2.17.0'
-    implementation 'com.fasterxml.jackson.datatype:jackson-datatype-jsr310:2.13.0'
-    implementation 'com.fasterxml.jackson.core:jackson-annotations:2.17.0'
-    implementation 'io.swagger.core.v3:swagger-core:2.2.21'
+    implementation "io.swagger.core.v3:swagger-models:${swaggerCoreVersion}"
+    implementation "io.swagger.core.v3:swagger-core:${swaggerCoreVersion}"
+    implementation "io.swagger.parser.v3:swagger-parser-core:${swaggerVersion}"
+    implementation "io.swagger.parser.v3:swagger-parser:${swaggerVersion}"
+    implementation "io.swagger.parser.v3:swagger-parser-v3:${swaggerVersion}"
+    implementation "com.fasterxml.jackson.core:jackson-databind:${jacksonVersion}"
+    implementation "com.fasterxml.jackson.datatype:jackson-datatype-jsr310:${jacksonVersion}"
+    implementation "com.fasterxml.jackson.core:jackson-annotations:${jacksonVersion}"
 
     // ZipArchive dependencies used for integration tests
     zipArchive group: 'org.opensearch.plugin', name:'opensearch-ml-plugin', version: "${opensearch_build}"
--- a/src/main/java/org/opensearch/flowframework/common/CommonValue.java
+++ b/src/main/java/org/opensearch/flowframework/common/CommonValue.java
@@ -234,10 +234,7 @@ public class CommonValue {
     /** The field name for reindex source index substitution */
     public static final String REINDEX_SOURCE_INDEX = "reindex.source_index";
 
-    /*
-     * Constants associated with API specification
-     */
-
+    /**URI for the YAML file of the ML Commons API specification.*/
     public static final String ML_COMMONS_API_SPEC_YAML_URI =
         "https://raw.githubusercontent.com/opensearch-project/opensearch-api-specification/refs/heads/main/spec/namespaces/ml.yaml";
 
--- a/src/main/java/org/opensearch/flowframework/common/CommonValue.java
+++ b/src/main/java/org/opensearch/flowframework/common/CommonValue.java
@@ -233,4 +233,12 @@ public class CommonValue {
     public static final String CREATE_INGEST_PIPELINE_MODEL_ID = "create_ingest_pipeline.model_id";
     /** The field name for reindex source index substitution */
     public static final String REINDEX_SOURCE_INDEX = "reindex.source_index";
+
+    /*
+     * Constants associated with API specification
+     */
+
+    public static final String ML_COMMONS_API_SPEC_YAML_URI =
+        "https://raw.githubusercontent.com/opensearch-project/opensearch-api-specification/refs/heads/main/spec/namespaces/ml.yaml";
+
 }
--- a/src/main/java/org/opensearch/flowframework/exception/ApiSpecParseException.java
+++ b/src/main/java/org/opensearch/flowframework/exception/ApiSpecParseException.java
@@ -1,0 +1,46 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+package org.opensearch.flowframework.exception;
+
+import org.opensearch.OpenSearchException;
+
+/**
+ * Custom exception to be thrown when an error occurs during the parsing of an API specification.
+ */
+public class ApiSpecParseException extends OpenSearchException {
+
+    /**
+     * Constructor with message.
+     *
+     * @param message The detail message.
+     */
+    public ApiSpecParseException(String message) {
+        super(message);
+    }
+
+    /**
+     * Constructor with message and cause.
+     *
+     * @param message The detail message.
+     * @param cause The cause of the exception.
+     */
+    public ApiSpecParseException(String message, Throwable cause) {
+        super(message, cause);
+    }
+
+    /**
+     * Constructor with message and list of detailed errors.
+     *
+     * @param message The detail message.
+     * @param detailedErrors The list of errors encountered during the parsing process.
+     */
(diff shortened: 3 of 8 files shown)


src/test/java/org/opensearch/flowframework/util/ApiSpecFetcherTests.java
```diff
@@ -0,0 +1,106 @@
...
+
+    private ApiSpecFetcher apiSpecFetcher;
+    private URI apiSpecUri;
+
+    @Before
+    public void setUp() throws Exception {
+        super.setUp();
+        this.apiSpecFetcher = new ApiSpecFetcher(); // Initialize your fetcher
+        this.apiSpecUri = new URI(
+            "https://raw.githubusercontent.com/opensearch-project/opensearch-api-specification/refs/heads/main/spec/namespaces/ml.yaml"
```
dbwiddis, line 32: Let's make this a constant.

src/main/java/org/opensearch/flowframework/util/ApiSpecFetcher.java
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    /**
+     * Parses the OpenAPI specification directly from the URI.
+     *
+     * @param apiSpecUri URI to the API specification (can be file path or web URI).
+     * @return Parsed OpenAPI object.
+     * @throws ApiSpecParseException If parsing fails.
+     */
+    public OpenAPI fetchApiSpec(URI apiSpecUri) {
```
dbwiddis, line 55: Looking at the test classes, we always create this URI from a String, then you spend a lot of time undoing the string in this method.  Do we need a URI argument here or can we just take a string?

dbwiddis (commented):
Looks great!  A few suggestions.

src/main/java/org/opensearch/flowframework/exception/ApiSpecParseException.java
```diff
@@ -0,0 +1,46 @@
...
+ * compatible open source license.
+ */
+package org.opensearch.flowframework.exception;
+
+import org.opensearch.OpenSearchException;
+
+/**
+ * Custom exception to be thrown when an error occurs during the parsing of an API specification.
+ */
+public class ApiSpecParseException extends OpenSearchException {
```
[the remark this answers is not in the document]
owaiskazi19: We have FlowFrameworkException which extends from `OpenSearchException`, should we utilize the same class for `ApiSpecParseException`?

src/main/java/org/opensearch/flowframework/util/ApiSpecFetcher.java
```diff
@@ -0,0 +1,115 @@
...
+import io.swagger.v3.oas.models.parameters.RequestBody;
+import io.swagger.v3.parser.OpenAPIV3Parser;
+import io.swagger.v3.parser.core.models.ParseOptions;
+import io.swagger.v3.parser.core.models.SwaggerParseResult;
+
+/**
+ * Utility class for fetching and parsing OpenAPI specifications.
+ */
+public class ApiSpecFetcher {
+    private static final Logger LOGGER = LogManager.getLogger(ApiSpecFetcher.class);
```
owaiskazi19, line 35:
```suggestion
    private static final Logger logger = LogManager.getLogger(ApiSpecFetcher.class);
```

src/test/java/org/opensearch/flowframework/workflow/RegisterAgentTests.java
```diff
@@ -150,4 +154,21 @@ public void testRegisterAgentFailure() throws IOException {
...
     }
+
+    public void testApiSpecRegisterAgentInputParamComparison() throws Exception {
+        List<String> requiredEnumParams = WorkflowStepFactory.WorkflowSteps.REGISTER_AGENT.inputs();
+
+        ApiSpecFetcher apiSpecFetcher = new ApiSpecFetcher();
+        boolean isMatch = apiSpecFetcher.compareRequiredFields(
+            requiredEnumParams,
+            new URI(
+                "https://raw.githubusercontent.com/opensearch-project/opensearch-api-specification/refs/heads/main/spec/namespaces/ml.yaml"
```
owaiskazi19, line 165: We can have a constant for it?

src/test/java/org/opensearch/flowframework/workflow/RegisterAgentTests.java
```diff
@@ -150,4 +154,21 @@ public void testRegisterAgentFailure() throws IOException {
         assertTrue(ex.getCause() instanceof FlowFrameworkException);
         assertEquals("Failed to register the agent", ex.getCause().getMessage());
     }
+
+    public void testApiSpecRegisterAgentInputParamComparison() throws Exception {
```
owaiskazi19, line 158: Can we have a success test as well?

owaiskazi19 (commented):
Great start!

src/main/java/org/opensearch/flowframework/util/ApiSpecFetcher.java
junweid62: I think string would be good, I simply followed the pattern in opensearch-java before.

src/main/java/org/opensearch/flowframework/exception/ApiSpecParseException.java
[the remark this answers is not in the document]
junweid62: I see the point about having `FlowFrameworkException`, but I believe `ApiSpecParseException` serves a distinct purpose by clearly indicating issues related to API spec parsing. Keeping it separate helps with better error categorization and makes it easier to track and handle specific parsing-related problems.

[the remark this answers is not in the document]
dbwiddis: @owaiskazi19 I agree with @junweid62 that there's room for a separate exception type here.  We primarily use FFE during operation, but this exception is almost exclusively used in tests.

That said, all the code here is in `main` but primarily meant for `test` so we _could_ move it there; even if just temporarily until we decide that we want to create an `ApiSpecStep` for our workflows.   WDYT?

src/main/java/org/opensearch/flowframework/util/ApiSpecFetcher.java
dbwiddis: +1 for copying existing code, but they probably have other external API requirements/integrations expecting that parameter.  I don't mind leaving it as-is in the interest of keeping code consistent (although a comment pointing to the source of the code for later cross-referencing would be nice).

build.gradle
```diff
@@ -33,7 +33,11 @@ buildscript {
         bwcFilePath = "src/test/resources/org/opensearch/flowframework/bwc/"
         bwcFlowFrameworkPath = bwcFilePath + "flowframework/"
 
-        isSameMajorVersion = opensearch_version.split("\\.")[0] == bwcVersionShort.split("\\.")[0]
+        isSameMajorVersion = opensearch_version.split("\\.")[0] == bwcVersionShort.split("\\.")
```
dbwiddis, line 36: Is this change intentional?

src/main/java/org/opensearch/flowframework/common/CommonValue.java
```diff
@@ -233,4 +233,12 @@ private CommonValue() {}
     public static final String CREATE_INGEST_PIPELINE_MODEL_ID = "create_ingest_pipeline.model_id";
     /** The field name for reindex source index substitution */
     public static final String REINDEX_SOURCE_INDEX = "reindex.source_index";
+
+    /*
+     * Constants associated with API specification
+     */
+
+    public static final String ML_COMMONS_API_SPEC_YAML_URI =
```
dbwiddis, line 241: Add a `/**` comment for the constant to make it part of the javadoc.  You can also put it on one line.  Not sure it needs its own category yet but won't hurt to leave that there for now.

src/main/java/org/opensearch/flowframework/util/ApiSpecFetcher.java
```diff
@@ -0,0 +1,115 @@
...
+    private static final ParseOptions PARSE_OPTIONS = new ParseOptions();
+    private static final OpenAPIV3Parser PARSER = new OpenAPIV3Parser();
```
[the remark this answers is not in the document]
dbwiddis: I think you changed the pares options name here (which was ok before) instead of the actual parser name (my suggestion) :)

src/main/java/org/opensearch/flowframework/util/ApiSpecFetcher.java
```diff
@@ -0,0 +1,110 @@
...
+        return false;
+    }
+
+    private static Content getContent(RestRequest.Method method, PathItem pathItem) throws IllegalArgumentException, ApiSpecParseException {
+        Operation operation = switch (method) {
+            case RestRequest.Method.POST -> pathItem.getPost();
+            case RestRequest.Method.GET -> pathItem.getGet();
+            case RestRequest.Method.PUT -> pathItem.getPut();
+            case RestRequest.Method.DELETE -> pathItem.getDelete();
+            default -> throw new IllegalArgumentException("Unsupported HTTP method: " + method);
```
dbwiddis, line 96: CodeCov indicates you haven't added tests for this failure case (and 2 below).

src/test/java/org/opensearch/flowframework/exception/ApiSpecParseExceptionTests.java
```diff
@@ -0,0 +1,42 @@
...
+    public void testApiSpecParseException() {
+        ApiSpecParseException exception = new ApiSpecParseException("API spec parsing failed");
+        assertTrue(true);
+        assertEquals("API spec parsing failed", exception.getMessage());
+    }
+
+    public void testApiSpecParseExceptionWithCause() {
+        Throwable cause = new RuntimeException("Underlying issue");
+        ApiSpecParseException exception = new ApiSpecParseException("API spec parsing failed", cause);
+        assertTrue(true);
```
dbwiddis, line 27: image

dbwiddis (commented):
A few more tweaks.

src/main/java/org/opensearch/flowframework/util/ApiSpecFetcher.java
dbwiddis: One more note:  we will have to backport this to 2.x which still supports JDK 11.   While I love the new functionality here, we may want to stick with old-style case statements.

src/test/java/org/opensearch/flowframework/exception/ApiSpecParseExceptionTests.java
junweid62: Haha, I just realized how funny this is. This will be the last time I use a code refactor suggestion without thinking it through.
Screenshot 2024-10-07 at 1 48 19 PM

src/test/java/org/opensearch/flowframework/workflow/RegisterAgentTests.java
owaiskazi19: Why do we have the below then?
```
 assertTrue("API spec input params do match enum required params", isMatch);
```

src/main/java/org/opensearch/flowframework/util/ApiSpecFetcher.java
```diff
@@ -0,0 +1,110 @@
...
+
+            List<String> requiredApiParams = schema.getRequired();
+            if (requiredApiParams != null && !requiredApiParams.isEmpty()) {
+                return new HashSet<>(requiredEnumParams).equals(new HashSet<>(requiredApiParams));
+            }
+        }
+        return false;
+    }
+
+    private static Content getContent(RestRequest.Method method, PathItem pathItem) throws IllegalArgumentException, ApiSpecParseException {
```
owaiskazi19, line 90: We can have a test for this method if we are not testing it at other places

src/test/java/org/opensearch/flowframework/workflow/RegisterAgentTests.java
junweid62: sure , we only need to verify the isMatch is true, will change

src/main/java/org/opensearch/flowframework/util/ApiSpecFetcher.java
junweid62: I add more test included this private method.

src/main/java/org/opensearch/flowframework/exception/ApiSpecParseException.java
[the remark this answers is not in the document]
dbwiddis: Agreed, it's fine in main and we will probably use it in the future.

src/test/java/org/opensearch/flowframework/util/ApiSpecFetcherTests.java
```diff
@@ -0,0 +1,126 @@
...
+
+        boolean comparisonResult = ApiSpecFetcher.compareRequiredFields(wrongRequiredParams, ML_COMMONS_API_SPEC_YAML_URI, path, method);
+
+        assertFalse("The required fields should not match for incorrect input", comparisonResult);
+    }
+
+    public void testCompareRequiredFieldsThrowsException() throws Exception {
+        String invalidUri = "http://invalid-url.com/fail.yaml";
+        String path = "/_plugins/_ml/agents/_register";
+        RestRequest.Method method = RestRequest.Method.POST;
```
dbwiddis, line 76: If you change this to `PUT` you'll cover another line in your switch/case.

src/test/java/org/opensearch/flowframework/util/ApiSpecFetcherTests.java
```diff
@@ -0,0 +1,126 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+package org.opensearch.flowframework.util;
+
+import org.opensearch.flowframework.exception.ApiSpecParseException;
+import org.opensearch.rest.RestRequest;
```
dbwiddis, line 12: (Style) If you import `RestRequest.Method` you'll have less verbose class definitions later.

src/main/java/org/opensearch/flowframework/util/ApiSpecFetcher.java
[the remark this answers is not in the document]
dbwiddis: This comment wasn't addressed.

dbwiddis (approved):
Approving with a few last items to clean up

build.gradle
```diff
@@ -179,6 +183,15 @@ dependencies {
...
     implementation "com.google.code.gson:gson:2.11.0"
+    // Swagger-Parser dependencies for API consistency tests
+    implementation "io.swagger.core.v3:swagger-models:${swaggerCoreVersion}"
+    implementation "io.swagger.core.v3:swagger-core:${swaggerCoreVersion}"
+    implementation "io.swagger.parser.v3:swagger-parser-core:${swaggerVersion}"
+    implementation "io.swagger.parser.v3:swagger-parser:${swaggerVersion}"
+    implementation "io.swagger.parser.v3:swagger-parser-v3:${swaggerVersion}"
+    implementation "com.fasterxml.jackson.core:jackson-databind:${jacksonVersion}"
+    implementation "com.fasterxml.jackson.datatype:jackson-datatype-jsr310:${jacksonVersion}"
+    implementation "com.fasterxml.jackson.core:jackson-annotations:${jacksonVersion}"
```
owaiskazi19, line 194: jackson dependencies are forced on L205. L192-194 should be moved under configurations as well

junweid62: Thank you for your suggestion. I tried moving the Jackson dependencies under configurations with force, but it seems that doing so doesn't properly import the dependencies. The force directive only enforces specific versions if the dependencies are already present, but it doesn't add them to the project on its own. Therefore, the implementation lines in dependencies are still necessary to ensure the Jackson libraries are actually included.

owaiskazi19: > only enforces specific versions 

And you are using the same version, right? 2.18.0. Any error you are seeing?

junweid62: Yes, I am using version 2.18.0 as specified in the force directive. The error I'm encountering is:

`java.lang.ClassNotFoundException: com.fasterxml.jackson.databind.Module`

It seems like the jackson-databind dependency isn't being imported properly, which is likely causing the issue. Even though the version is forced, the dependency itself doesn't get added unless it's explicitly declared. That's why I still need the implementation lines in the dependencies block to ensure that these libraries are available in the project.

owaiskazi19: I see. Let's discuss more on this. Approving the PR

owaiskazi19 (approved):

src/main/java/org/opensearch/flowframework/util/ApiSpecFetcher.java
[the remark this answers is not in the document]
owaiskazi19: It is now :)

Issue #526, opened 2024-02-16
Support workflow versioning


ohltyler:
## Background
Proposed @opensearch-project/opensearch-ux mockups for the frontend plugin introduces a concept of deployments within a single workflow. This allows for easy prototyping for a particular use-case, allowing users to construct and test different workflow configurations, persisting the history and associated resources with each deployment for reference.

![Screenshot 2024-02-16 at 2 47 50 PM](https://github.com/opensearch-project/flow-framework/assets/62119629/35608556-704f-4375-b7c5-1b295cf45380)

The current backend design does not support this paradigm. One approach to support this is introducing versioning to workflows, opening up functionality to iterate on a workflow at the API level to persist and provision multiple versions of a single workflow. Originally discussed in #109, this issue expands on this idea.

## Implementation
Currently there is a `version` field (see [here](https://github.com/opensearch-project/documentation-website/blob/main/_automating-workflows/workflow-tutorial.md#complete-json-workflow-template)) but it is not being used or validated yet. We can refactor this into being a required field with a `String` value that users can set to whatever they want. This way, there will always be a default `version` for a workflow. (If we need to persist `compatibility`, we can refactor that into a standalone template field)

Internally, workflows with the same ID but different versions can behave identically to the current workflow implementation; users can perform CRUD operations on them, they each persist their own provisioning state, created/associated resources, etc.

To handle defaults at the API level, backend logic can filter through the workflow ID, and if there are multiple instances of that workflow ID under different versions, choose the one with the latest `last_update_time` or equivalent such field.

## API changes
Users need the capability to run CRUD operations on individual versions of a workflow if they want. Users who don't want/care about versioning shouldn't have to, with abstracted out logic and sufficient defaults. Most changes listed below are just adding an additional path to existing APIs that include a version ID.

### Create workflow
Users specify a (now mandatory) `version` field in the template.
```
POST /_plugins/_flow_framework/workflow
{
   "version": "my-version-1",
   "name": "my-workflow-name",
   ...
}
```

### [NEW] Create new version of a workflow
Same as create, but now include the already-created `workflow-id` in the path. The JSON body should include a non-name-clashing `version` with any existing versions for that workflow, else throw an error. See open questions below on what to include in this JSON body.
```
POST /_plugins/_flow_framework/workflow/<workflow-id>
{
   "version": "my-version-2",
   "name": "my-workflow-name",
   ...
}
```

### Update workflow
[...]

PR #918, opened 2024-10-17
Fixed Template Update Location and Improved Logger Statements in ReprovisionWorkflowTransportAction

### LINKED ISSUE #870: [BUG] Reprovision Workflow IT is flaky
### What is the bug?

`FlowFrameworkRestApiIT.testReprovisionWorkflow()` is flaky.  During 2.17.0 release it required multiple retries to pass on Windows x86.  There are occasional GitHub CI failures on other platforms, see https://github.com/opensearch-project/flow-framework/actions/runs/10777769278/job/29887668921#step:4:59 (Ubuntu) and https://github.com/opensearch-project/flow-framework/actions/runs/10777865339/job/29888745661#step:4:49 (macOS)

```
FlowFrameworkRestApiIT > testReprovisionWorkflow FAILED
    org.junit.ComparisonFailure: expected:<[COMPLET]ED> but was:<[FAIL]ED>
        at __randomizedtesting.SeedInfo.seed([4D7D6[49](https://github.com/opensearch-project/flow-framework/actions/runs/10777865339/job/29888745661#step:4:50)945752BD3:196DC17D8E477B7D]:0)
        at org.junit.Assert.assertEquals(Assert.java:117)
        at org.junit.Assert.assertEquals(Assert.java:146)
        at org.opensearch.flowframework.FlowFrameworkRestTestCase.getAndAssertWorkflowStatus(FlowFrameworkRestTestCase.java:7[53](https://github.com/opensearch-project/flow-framework/actions/runs/10777865339/job/29888745661#step:4:54))
        at org.opensearch.flowframework.FlowFrameworkRestTestCase.lambda$getResourcesCreated$7(FlowFrameworkRestTestCase.java:805)
        at org.opensearch.test.OpenSearchTestCase.assertBusy(OpenSearchTestCase.java:1136)
        at org.opensearch.flowframework.FlowFrameworkRestTestCase.getResourcesCreated(FlowFrameworkRestTestCase.java:804)
        at org.opensearch.flowframework.rest.FlowFrameworkRestApiIT.testReprovisionWorkflow(FlowFrameworkRestApiIT.java:471)
```

Run our CI repeatedly.  It fails reasonably often.

Tests consistently pass.

### What is your host/environment?

Reproducible on Windows, Linux, macOS as shown above

The failure is occurring consistently on the attempt to reprovision to remove default ingest pipeline (setting it to `_none`, which is the `SearchPipelineService.NOOP_PIPELINE_ID`).

The change in default pipeline name is the only difference in the two templates.

saimedhi:
### Description
This PR fixes the template update location in ReprovisionWorkflowTransportAction.java and enhances logger statements for better clarity and debugging.

### Related Issues
closes #870 

### Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,10 +7,10 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 ### Features
 ### Enhancements
 ### Bug Fixes
+- Fixed Template Update Location and Improved Logger Statements in ReprovisionWorkflowTransportAction ([#918](https://github.com/opensearch-project/flow-framework/pull/918))
 ### Infrastructure
 - Set Java target compatibility to JDK 21 ([#730](https://github.com/opensearch-project/flow-framework/pull/730))
 - Add knowledge base alert agent into sample templates ([#874](https://github.com/opensearch-project/flow-framework/pull/874))
-
 ### Documentation
 - Add alert summary agent template ([#873](https://github.com/opensearch-project/flow-framework/pull/873))
 
@@ -25,6 +25,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 - Incrementally remove resources from workflow state during deprovisioning ([#898](https://github.com/opensearch-project/flow-framework/pull/898))
 
 ### Bug Fixes
+- Fixed Template Update Location and Improved Logger Statements in ReprovisionWorkflowTransportAction ([#918](https://github.com/opensearch-project/flow-framework/pull/918))
 ### Infrastructure
 ### Documentation
 - Add query assist data summary agent into sample templates ([#875](https://github.com/opensearch-project/flow-framework/pull/875))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,10 +7,10 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 ### Features
 ### Enhancements
 ### Bug Fixes
-- Fixed Template Update Location and Improved Logger Statements in ReprovisionWorkflowTransportAction ([#918](https://github.com/opensearch-project/flow-framework/pull/918))
 ### Infrastructure
 - Set Java target compatibility to JDK 21 ([#730](https://github.com/opensearch-project/flow-framework/pull/730))
 - Add knowledge base alert agent into sample templates ([#874](https://github.com/opensearch-project/flow-framework/pull/874))
+
 ### Documentation
 - Add alert summary agent template ([#873](https://github.com/opensearch-project/flow-framework/pull/873))
 
@@ -26,9 +26,11 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 
 ### Bug Fixes
 - Fixed Template Update Location and Improved Logger Statements in ReprovisionWorkflowTransportAction ([#918](https://github.com/opensearch-project/flow-framework/pull/918))
+
 ### Infrastructure
 ### Documentation
 - Add query assist data summary agent into sample templates ([#875](https://github.com/opensearch-project/flow-framework/pull/875))
+
 ### Maintenance
 ### Refactoring
 - Update workflow state without using painless script ([#894](https://github.com/opensearch-project/flow-framework/pull/894))
--- a/src/main/java/org/opensearch/flowframework/transport/ReprovisionWorkflowTransportAction.java
+++ b/src/main/java/org/opensearch/flowframework/transport/ReprovisionWorkflowTransportAction.java
@@ -271,8 +271,9 @@ public class ReprovisionWorkflowTransportAction extends HandledTransportAction<R
     ) {
         try {
             threadPool.executor(PROVISION_WORKFLOW_THREAD_POOL).execute(() -> {
-                updateTemplate( template,workflowId);
-                executeWorkflow(template, workflowSequence, workflowId); });
+                updateTemplate(template, workflowId);
+                executeWorkflow(template, workflowSequence, workflowId);
+            });
         } catch (Exception exception) {
             listener.onFailure(new FlowFrameworkException("Failed to execute workflow " + workflowId, ExceptionsHelper.status(exception)));
         }
@@ -283,16 +284,11 @@ public class ReprovisionWorkflowTransportAction extends HandledTransportAction<R
      * @param template The template to store after reprovisioning completes successfully
      * @param workflowId The workflowId associated with the workflow that is executing
      */
-    private void updateTemplate ( Template template, String workflowId) {
-        flowFrameworkIndicesHandler.updateTemplateInGlobalContext(
-                workflowId,
-                template,
-                ActionListener.wrap(templateResponse -> {
-                    logger.info("Updated template for {}", workflowId);
-                }, exception -> {
-                    logger.error("Failed to update use case template for {}", workflowId, exception);
-                }),
-                true  // ignores NOT_STARTED state if request is to reprovision
+    private void updateTemplate(Template template, String workflowId) {
+        flowFrameworkIndicesHandler.updateTemplateInGlobalContext(workflowId, template, ActionListener.wrap(templateResponse -> {
+            logger.info("Updated template for {}", workflowId);
+        }, exception -> { logger.error("Failed to update use case template for {}", workflowId, exception); }),
+            true  // ignores NOT_STARTED state if request is to reprovision
         );
     }
 
--- a/src/main/java/org/opensearch/flowframework/transport/ReprovisionWorkflowTransportAction.java
+++ b/src/main/java/org/opensearch/flowframework/transport/ReprovisionWorkflowTransportAction.java
@@ -271,9 +271,8 @@ public class ReprovisionWorkflowTransportAction extends HandledTransportAction<R
     ) {
         try {
             threadPool.executor(PROVISION_WORKFLOW_THREAD_POOL).execute(() -> {
-                updateTemplate(template, workflowId);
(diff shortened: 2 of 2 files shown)


src/main/java/org/opensearch/flowframework/transport/ReprovisionWorkflowTransportAction.java
```diff
@@ -270,27 +270,45 @@ private void executeWorkflowAsync(
...
 
     /**
-     * Executes the given workflow sequence
+     * Replace template document
      * @param template The template to store after reprovisioning completes successfully
+     * @param workflowId The workflowId associated with the workflow that is executing
+     */
+    private void updateTemplate(Template template, String workflowId) {
+        flowFrameworkIndicesHandler.updateTemplateInGlobalContext(workflowId, template, ActionListener.wrap(templateResponse -> {
+            logger.info("Updated template for {}", workflowId, State.COMPLETED);
```
dbwiddis, line 289: The `State.COMPLETED` here is ignored as there's only one `{}` parameter. Which is a good thing because I thought it was out of place anyway, and should just be deleted.

src/main/java/org/opensearch/flowframework/transport/ReprovisionWorkflowTransportAction.java
```diff
@@ -270,27 +270,45 @@ private void executeWorkflowAsync(
...
+        }, exception -> {
+            String errorMessage = "Failed to update use case template for " + workflowId;
+            logger.error(errorMessage, exception);
```
dbwiddis, line 292: Just put the message inline in the `logger.error()` with the substitution parameters.  (Yes I know you just refactored what was already there but we can improve it!)

src/main/java/org/opensearch/flowframework/transport/ReprovisionWorkflowTransportAction.java
```diff
@@ -270,27 +270,45 @@ private void executeWorkflowAsync(
...
+    private void executeWorkflow(List<ProcessNode> workflowSequence, String workflowId) {
         String currentStepId = "";
         try {
             Map<String, PlainActionFuture<?>> workflowFutureMap = new LinkedHashMap<>();
             for (ProcessNode processNode : workflowSequence) {
                 List<ProcessNode> predecessors = processNode.predecessors();
                 logger.info(
                     "Queueing process [{}].{}",
-                    processNode.id(),
+                    processNode.id() + processNode.workflowStep().getName(),
```
dbwiddis, line 311: Have some sort of delimiter here, otherwise these two strings will run together.  As written assuming we're using the semantic search template, this will appear as `Queueing process [create_ingest_pipelineupdate_ingest_pipeline].`.  

Not sure the best way to do this but perhaps `[create_ingest_pipeline (update_ingest_pipeline)]`?  Open to other ideas.

dbwiddis (commented):
I see it's in draft, offering an initial review.

src/main/java/org/opensearch/flowframework/transport/ReprovisionWorkflowTransportAction.java
```diff
@@ -290,7 +306,7 @@ private void executeWorkflow(Template template, List<ProcessNode> workflowSequen
                 List<ProcessNode> predecessors = processNode.predecessors();
                 logger.info(
                     "Queueing process [{}].{}",
-                    processNode.id(),
+                    String.format("%s (type: %s)", processNode.id(), processNode.workflowStep().getName()),
```
dbwiddis, line 309: You shouldn't need to use `String.format()` with logger substitutions.  This should be achievable with something like `"Queueing Process [{} (type: {})].()"`

dbwiddis (commented):
LGTM so far!

amitgalitz:
Overall LGTM, just a quick question, in the case the reprovisioning itself fails, will have an issue that the template is updated and we need to roll that back?

amitgalitz (approved):

src/main/java/org/opensearch/flowframework/transport/ReprovisionWorkflowTransportAction.java
```diff
@@ -270,12 +270,28 @@ private void executeWorkflowAsync(
         ActionListener<WorkflowResponse> listener
     ) {
         try {
-            threadPool.executor(PROVISION_WORKFLOW_THREAD_POOL).execute(() -> { executeWorkflow(template, workflowSequence, workflowId); });
+            threadPool.executor(PROVISION_WORKFLOW_THREAD_POOL).execute(() -> {
+                updateTemplate(template, workflowId);
```
joshpalis, line 274: One major concern I have here is error handling during subsequent executions. 

Reprovisioning depends on comparing the old template stored in the index with the new template configuration provided in the request body. We compare each node and identify which steps have been modified, creating a new workflow to either create new or update existing resources. 

Let's say we attempt to reprovision a workflow and it executes the first few steps and this fails at the end due to some transitive error. Since we've already indexed the updated template, we would be comparing the new template with itself. We will need to handle this case more gracefully and perhaps consider rolling back the attempted modification if a failure occurs, essentially returning the cluster to its state prior to the reprovision command. This is a similar approach that CloudFormation templates currently take.

dbwiddis: >  We will need to handle this case more gracefully and perhaps consider rolling back the attempted modification if a failure occurs, 

This is "fine grained deprovisioning" and the current code is not set up to support that.  To get to this level of granularity we're going to need to really figure out versioning (See #526).

For now let's focus on what the correct action is to take on this PR.

Currently when we provision, we save the template (in the desired to-be state) and then attempt to provision it.  If it's successful we know the template/state match.  If it fails, we don't know the state.  We may have partially deployed resources.  At this point, the only correct action is to deprovision (which ignores the template, and just gets rid of resources).

So, now consider reprovisioning.  Current state is to keep the old template there, and only update it after provisioning completes.  (But we currently set the workflow state to complete prior to updating the template which leads to a possible race condition (see the linked bug) where we can quickly reprovision again but not have the current template to compare to.   We absolutely must update the template before setting state to completed.

But why not update it earlier to match the behavior of provisioning?  Attempt to do so, if it fails, the only thing we can do is deprovision.

Here's a concrete example.
1. Provision a template with Step A and Step B.   
2. On success, state index has Resources A and B.
3. Now reprovision with a template with Step A, Step X, Step B, Step Y.
4. Step X completes successfully, state index has Resources A, B, and X.
5. Step Y fails for some reason.  

IF we haven't updated the template it has the old "A, B" steps and looking at the state you may wonder where X comes from or what additional steps are missing.  Worse, if you attempt to reprovision from here, you may get a conflict trying to add X again.  The only thing we can do is deprovision and start again from scratch, but we didn't save the "updated" template so we still need to save that template and provision with it.

If we have updated the template it has A, X, B, Y, and we can see "oh, we can manually add Y" if we're doing it interactively.  Or, we can just give up on it, deprovision completely, and provision completely... in which case the correct 4-step template is already there.  

In either failure case, we have to deprovision to recover automaticaly.   "Reprovisioning" is just a short cut to the deprovisoin/provision sequence in the happy case.  

Once you have a failure, you can't/shouldn't even try to reprovision because no matter what, the state index is out of sync.  In this case, I think having the latest "desired" template lets you do the full deprovision + provision existing template steps to get back in sync.

CHANGELOG.md
```diff
@@ -25,6 +25,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 - Incrementally remove resources from workflow state during deprovisioning ([#898](https://github.com/opensearch-project/flow-framework/pull/898))
 
 ### Bug Fixes
+- Fixed Template Update Location and Improved Logger Statements in ReprovisionWorkflowTransportAction ([#918](https://github.com/opensearch-project/flow-framework/pull/918))
 ### Infrastructure
```
dbwiddis, line 29: Nit: Needs a line of whitespace before the markdown header.
```suggestion
- Fixed Template Update Location and Improved Logger Statements in ReprovisionWorkflowTransportAction ([#918](https://github.com/opensearch-project/flow-framework/pull/918))

### Infrastructure
```

CHANGELOG.md
```diff
@@ -7,10 +7,10 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 ### Features
 ### Enhancements
 ### Bug Fixes
+- Fixed Template Update Location and Improved Logger Statements in ReprovisionWorkflowTransportAction ([#918](https://github.com/opensearch-project/flow-framework/pull/918))
```
dbwiddis, line 10: We are backporting this so it shouldn't go into 3.x section.

dbwiddis (approved):
Approved pending changelog fix and resolving @joshpalis comment.

src/main/java/org/opensearch/flowframework/transport/ReprovisionWorkflowTransportAction.java
joshpalis: Thanks for the input Dan. These changes look good to me then until we can implement fine-grained deprovisioning. In this case, we can rely on documentation to indicate to the user to deprovision if reprovision failed.

joshpalis (approved):

CHANGELOG.md
```diff
@@ -7,10 +7,10 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
...
 ### Infrastructure
 - Set Java target compatibility to JDK 21 ([#730](https://github.com/opensearch-project/flow-framework/pull/730))
 - Add knowledge base alert agent into sample templates ([#874](https://github.com/opensearch-project/flow-framework/pull/874))
-
```
dbwiddis, line 13: Don't remove this linebreak

PR #943, opened 2024-11-04
Replace String concatenation with Log4j ParameterizedMessage for readability

### LINKED ISSUE #905: Replace String concatenation with Log4j ParameterizedMessage for readability

Log4j includes parameter substitution which improves readability of log messages and has improved performance and stability over similar `String.format()`:
```
logger.info("Updating {} with {}.", baz.getFoo(), baz.getBar());
```
This works well when we are only logging a message.  However, we have many cases in our Transport Action exception handling where we use the same message in both the log and exception message (I know because I wrote them!).  One such example:
https://github.com/opensearch-project/flow-framework/blob/7a93d6c543738c7250313b0f66f3f794f42f5f1b/src/main/java/org/opensearch/flowframework/transport/ProvisionWorkflowTransportAction.java#L140-L144

Log4J's `ParameterizedMessage` class provides a way to construct the message the same way for both cases.

```java
import org.apache.logging.log4j.message.ParameterizedMessageFactory;

} catch (Exception e) {
    String errorMessage = ParameterizedMessageFactory.INSTANCE.newMessage(
        "Failed to retrieve template from global context for workflow {}",
        workflowId
    ).getFormattedMessage();
    listener.onFailure(new FlowFrameworkException(errorMessage, ExceptionsHelper.status(e)));
}
```

Example commit inspiring this issue: https://github.com/opensearch-project/flow-framework/pull/898/commits/86ac8ea862ba59abcbdf036fd637159e751e3be4

1. Keep the current concatenation (less readable)
2. Replace with `StringBuilder` implementation (less readable)
3. Replace with `String.format()` (performance, can throw exceptions, locale concerns)

There are lots of these.  As a good first issue, doing any one of them, or perhaps a class at a time, would be a welcome contribution!

In addition, since most of them follow the same pattern, a new method (per class) of `handleException(Exception e, String formatString, Object... args)` would probably be useful.

KirrTap:
### Description
I have replaced string concatenation with Log4j ParameterizedMessage in 40 files.

### Related Issues
Resolves #905 

### Check List


src/main/java/org/opensearch/flowframework/rest/RestCreateWorkflowAction.java
```diff
@@ -242,7 +244,9 @@ protected RestChannelConsumer prepareRequest(RestRequest request, NodeClient cli
                     XContentBuilder exceptionBuilder = ex.toXContent(channel.newErrorBuilder(), ToXContent.EMPTY_PARAMS);
                     channel.sendResponse(new BytesRestResponse(ex.getRestStatus(), exceptionBuilder));
                 } catch (IOException e) {
-                    String errorMessage = "IOException: Failed to send back create workflow exception";
+                    String errorMessage = ParameterizedMessageFactory.INSTANCE.newMessage(
+                        "IOException: Failed to send back create workflow exception"
```
dbwiddis, line 248: No need to rewrite this one (and many similar), EXCEPT... if you want, you could create a constant in the `CommonValue` class, something like `FAILED_TO_SEND_EXCEPTION` with the value "IOException: Failed to send back {} exception".  Then here you'd add the string "create workflow" and in the other similar ones you'd change it (get, delete, etc.).

src/main/java/org/opensearch/flowframework/rest/RestDeleteWorkflowAction.java
```diff
@@ -92,7 +93,9 @@ protected BaseRestHandler.RestChannelConsumer prepareRequest(RestRequest request
                     XContentBuilder exceptionBuilder = ex.toXContent(channel.newErrorBuilder(), ToXContent.EMPTY_PARAMS);
                     channel.sendResponse(new BytesRestResponse(ex.getRestStatus(), exceptionBuilder));
                 } catch (IOException e) {
-                    String errorMessage = "IOException: Failed to send back delete workflow exception";
+                    String errorMessage = ParameterizedMessageFactory.INSTANCE.newMessage(
+                        "IOException: Failed to send back delete workflow exception"
```
dbwiddis, line 97: see above; no need to replace a plain string but consider parameterizing "delete"

src/main/java/org/opensearch/flowframework/rest/RestDeprovisionWorkflowAction.java
```diff
@@ -94,7 +95,9 @@ protected BaseRestHandler.RestChannelConsumer prepareRequest(RestRequest request
                     XContentBuilder exceptionBuilder = ex.toXContent(channel.newErrorBuilder(), ToXContent.EMPTY_PARAMS);
                     channel.sendResponse(new BytesRestResponse(ex.getRestStatus(), exceptionBuilder));
                 } catch (IOException e) {
-                    String errorMessage = "IOException: Failed to send back deprovision workflow exception";
+                    String errorMessage = ParameterizedMessageFactory.INSTANCE.newMessage(
+                        "IOException: Failed to send back deprovision workflow exception"
```
dbwiddis, line 99: see above; no need to replace a plain string but consider parameterizing "deprovision workflow"

src/main/java/org/opensearch/flowframework/rest/RestGetWorkflowAction.java
```diff
@@ -91,7 +92,9 @@ protected BaseRestHandler.RestChannelConsumer prepareRequest(RestRequest request
                     XContentBuilder exceptionBuilder = ex.toXContent(channel.newErrorBuilder(), ToXContent.EMPTY_PARAMS);
                     channel.sendResponse(new BytesRestResponse(ex.getRestStatus(), exceptionBuilder));
                 } catch (IOException e) {
-                    String errorMessage = "IOException: Failed to send back get workflow exception";
+                    String errorMessage = ParameterizedMessageFactory.INSTANCE.newMessage(
+                        "IOException: Failed to send back get workflow exception"
```
dbwiddis, line 96: see above; no need to replace a plain string but consider parameterizing "get workflow"

src/main/java/org/opensearch/flowframework/rest/RestGetWorkflowStateAction.java
```diff
@@ -87,7 +88,9 @@ protected BaseRestHandler.RestChannelConsumer prepareRequest(RestRequest request
                     XContentBuilder exceptionBuilder = ex.toXContent(channel.newErrorBuilder(), ToXContent.EMPTY_PARAMS);
                     channel.sendResponse(new BytesRestResponse(ex.getRestStatus(), exceptionBuilder));
                 } catch (IOException e) {
-                    String errorMessage = "IOException: Failed to send back get workflow state exception";
+                    String errorMessage = ParameterizedMessageFactory.INSTANCE.newMessage(
+                        "IOException: Failed to send back get workflow state exception"
```
dbwiddis, line 92: see above; no need to replace a plain string but consider parameterizing "get workflow state"

src/main/java/org/opensearch/flowframework/rest/RestGetWorkflowStepAction.java
```diff
@@ -91,7 +92,9 @@ protected RestChannelConsumer prepareRequest(RestRequest request, NodeClient cli
                     XContentBuilder exceptionBuilder = ex.toXContent(channel.newErrorBuilder(), ToXContent.EMPTY_PARAMS);
                     channel.sendResponse(new BytesRestResponse(ex.getRestStatus(), exceptionBuilder));
                 } catch (IOException e) {
-                    String errorMessage = "IOException: Failed to send back get workflow step exception";
+                    String errorMessage = ParameterizedMessageFactory.INSTANCE.newMessage(
+                        "IOException: Failed to send back get workflow step exception"
```
dbwiddis, line 96: see above; no need to replace a plain string but consider parameterizing "get workflow step"

src/main/java/org/opensearch/flowframework/rest/RestProvisionWorkflowAction.java
```diff
@@ -98,7 +99,9 @@ protected RestChannelConsumer prepareRequest(RestRequest request, NodeClient cli
                     XContentBuilder exceptionBuilder = ex.toXContent(channel.newErrorBuilder(), ToXContent.EMPTY_PARAMS);
                     channel.sendResponse(new BytesRestResponse(ex.getRestStatus(), exceptionBuilder));
                 } catch (IOException e) {
-                    String errorMessage = "IOException: Failed to send back provision workflow exception";
+                    String errorMessage = ParameterizedMessageFactory.INSTANCE.newMessage(
+                        "IOException: Failed to send back provision workflow exception"
```
dbwiddis, line 103: see above; no need to replace a plain string but consider parameterizing "provision workflow"

src/main/java/org/opensearch/flowframework/transport/CreateWorkflowTransportAction.java
```diff
@@ -256,7 +263,9 @@ private void createExecute(WorkflowRequest request, User user, ActionListener<Wo
                                                         ActionListener.wrap(provisionResponse -> {
                                                             listener.onResponse(new WorkflowResponse(provisionResponse.getWorkflowId()));
                                                         }, exception -> {
-                                                            String errorMessage = "Provisioning failed.";
+                                                            String errorMessage = ParameterizedMessageFactory.INSTANCE.newMessage(
+                                                                "Provisioning failed."
```
dbwiddis, line 267: plain string

src/main/java/org/opensearch/flowframework/transport/CreateWorkflowTransportAction.java
```diff
@@ -274,7 +283,9 @@ private void createExecute(WorkflowRequest request, User user, ActionListener<Wo
                                                     listener.onResponse(new WorkflowResponse(globalContextResponse.getId()));
                                                 }
                                             }, exception -> {
-                                                String errorMessage = "Failed to save workflow state";
+                                                String errorMessage = ParameterizedMessageFactory.INSTANCE.newMessage(
+                                                    "Failed to save workflow state"
```
dbwiddis, line 287: plain string

dbwiddis (commented):
Thanks for submitting this PR!

Appreciate all the work -- one thing to note is that you don't need to do anything with values that are already just plain strings with nothing to insert into them.  Just leave those alone.

Except, as indicated, when there are a lot of strings with similar text except for a few words, you might consider making a log message constant in the CommonValue class and using it in those places.

I stopped after several files, will let you submit an update before a more complete review.

dbwiddis (approved):
Thanks for this PR!  

One last request, please add an entry in CHANGELOG.md under the "Refactoring" header linking to this PR.  After that it should be ready to merge.

owaiskazi19 (approved):
Thanks @KirrTap for the change

owaiskazi19:
@KirrTap can you do a manual backport to 2.x change and raise a PR?

PR #952, opened 2024-11-08
Remove useCase and defaultParams field in WorkflowRequest

### LINKED ISSUE #758: [BUG] WorkflowRequest doesn't serialize the use case and default parameters over transport
### What is the bug?

The `WorkflowRequest` class doesn't serialize the use case and default parameters over transport.

The `writeTo()` only writes the provision parameter and params map:
https://github.com/opensearch-project/flow-framework/blob/cf1016fd72e0f9196db498cbb52f4a57bafadade/src/main/java/org/opensearch/flowframework/transport/WorkflowRequest.java#L197-L206

Similarly, the constructor using `StreamInput` doesn't read them
https://github.com/opensearch-project/flow-framework/blob/cf1016fd72e0f9196db498cbb52f4a57bafadade/src/main/java/org/opensearch/flowframework/transport/WorkflowRequest.java#L128-L136

This was missed in unit testing because the tests incorrectly checked the original `workflowRequest` and not the `streamInputRequest` after passing through transport. (Note the last 4 lines.)
https://github.com/opensearch-project/flow-framework/blob/cf1016fd72e0f9196db498cbb52f4a57bafadade/src/test/java/org/opensearch/flowframework/transport/WorkflowRequestResponseTests.java#L155-L166

Fix the unit test to properly test the workflow created from transport. Change lines 163-166 above to:
```java
        assertNull(streamInputRequest.validate());
        assertFalse(streamInputRequest.isProvision());
        // FAILS, `getDefaultParams()` is null
        assertTrue(streamInputRequest.getDefaultParams().isEmpty()); 
        // FAILS, `getUseCase()` is null
        assertEquals(streamInputRequest.getUseCase(), "cohere-embedding_model_deploy");
```

If these fields are required to be part of the `WorkflowRequest` they should be properly serialized.  Note that adding these fields would break compatibility across versions and would require version-checking for the additional fields.

However, these fields are never accessed in the code, other than in the (failing) tests.  If they are not needed, they should be removed.

The fields were added in #583.

The tests have been fixed in #757, with the failing tests commented out until this issue is resolved.

junweid62:
### Description
- Removed `useCase` And `defaultParams` since the WorkflowRequest class doesn't serialize the use case and default parameters over transport. 

- Also removed the constructor that has never been used in our code base

### Related Issues
Resolves #758 

### Check List
- ~[ ] New functionality has been documented.~
- ~[ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).~

### DIFF
--- a/src/test/java/org/opensearch/flowframework/transport/WorkflowRequestResponseTests.java
+++ b/src/test/java/org/opensearch/flowframework/transport/WorkflowRequestResponseTests.java
@@ -153,69 +153,10 @@ public class WorkflowRequestResponseTests extends OpenSearchTestCase {
         assertEquals("bar", streamInputRequest.getParams().get("foo"));
     }
 
-    public void testWorkflowRequestWithUseCase() throws IOException {
-        WorkflowRequest workflowRequest = new WorkflowRequest("123", template, "cohere-embedding_model_deploy", Collections.emptyMap());
-        assertNotNull(workflowRequest.getWorkflowId());
-        assertEquals(template, workflowRequest.getTemplate());
-        assertNull(workflowRequest.validate());
-        assertFalse(workflowRequest.isProvision());
-        assertFalse(workflowRequest.isUpdateFields());
-        assertTrue(workflowRequest.getDefaultParams().isEmpty());
-        assertEquals(workflowRequest.getUseCase(), "cohere-embedding_model_deploy");
-
-        BytesStreamOutput out = new BytesStreamOutput();
-        workflowRequest.writeTo(out);
-        BytesStreamInput in = new BytesStreamInput(BytesReference.toBytes(out.bytes()));
-
-        WorkflowRequest streamInputRequest = new WorkflowRequest(in);
-
-        assertEquals(workflowRequest.getWorkflowId(), streamInputRequest.getWorkflowId());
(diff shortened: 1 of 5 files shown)


joshpalis (approved):

src/test/java/org/opensearch/flowframework/transport/WorkflowRequestResponseTests.java
```diff
@@ -153,69 +153,10 @@ public void testWorkflowRequestWithParams() throws IOException {
         assertEquals("bar", streamInputRequest.getParams().get("foo"));
     }
 
-    public void testWorkflowRequestWithUseCase() throws IOException {
```
dbwiddis, line 156: Should these tests be fully removed, or just remove the parts related to the use case and params?  

Since that particular method of provisioning produces a different code path, it's useful to make sure the existing params match the expected

dbwiddis (commented):
LGTM, but not sure if the removed tests are completely duplicated somewhere else or provide at least one interesting change worth testing (even if not the new params)

dbwiddis (approved):
> Also removed the constructor that has never been used in our code base

Although this comment seems to be the "new thing" that was tested so probably good to remove.

PR #980, opened 2024-12-13
Implement multi tenancy in Flow Framework

dbwiddis:
### Description

Uses the multitenant remote metadata client developed in ML Commons `feature/multi_tenancy` branch to migrate Flow Framework system indices to use the same pattern.

Submitting in logical steps to ease maintainer review.  Suggestions:
1. To start, review each of the first several commits to understand how the tenant id is implemented and passed around
2. Then visit the tenant aware integ test classes to understand what it is we're trying to allow/prohibit

### Check List


.github/workflows/test_security.yml
```diff
@@ -34,6 +34,22 @@ jobs:
       options: --user root
 
     steps:
+      # TEMPORARY until this is on Maven
+      - name: Set up JDK 21
```
dhrubo-os, line 38: just curious why do we need to setup JDK again when are doing this at line 56?

dbwiddis: I hope that all the code between those TEMPORARY comments will be removed before this is merged.

The one on line 56 will stay, the one in here will be removed.  I didn't want to otherwise change the order of things that will remain.

src/main/java/org/opensearch/flowframework/indices/FlowFrameworkIndicesHandler.java
```diff
@@ -486,135 +504,201 @@ public void updateTemplateInGlobalContext(
...
+                            ensureExpectedToken(XContentParser.Token.START_OBJECT, parser.nextToken(), parser);
+                            WorkflowState workflowState = WorkflowState.parse(parser);
+                            listener.onResponse(workflowState);
+                        } catch (Exception e) {
+                            String errorMessage = ParameterizedMessageFactory.INSTANCE.newMessage(
+                                "Failed to parse workflowState: {}",
+                                getResponse.getId()
+                            ).getFormattedMessage();
+                            logger.error(errorMessage, e);
+                            listener.onFailure(new FlowFrameworkException(errorMessage, RestStatus.BAD_REQUEST));
```
owaiskazi19, line 602: Can be a INTERNAL_SERVER_ERROR here

src/main/java/org/opensearch/flowframework/indices/FlowFrameworkIndicesHandler.java
```diff
@@ -628,13 +712,38 @@ public void updateFlowFrameworkSystemIndexDoc(
...
+                    .dataObject(updatedDocument)
+                    .retryOnConflict(RETRIES)
+                    .build();
+                sdkClient.updateDataObjectAsync(updateRequest).whenComplete((r, throwable) -> {
+                    context.restore();
+                    if (throwable == null) {
+                        UpdateResponse response;
+                        try {
+                            response = UpdateResponse.fromXContent(r.parser());
+                            logger.info("Deleted workflow state doc: {}", documentId);
```
owaiskazi19, line 728: Aren't we updating here?

owaiskazi19 (commented):
Great to see multi tenancy support in FF. Started with the initial review

build.gradle
```diff
@@ -203,6 +208,12 @@ dependencies {
         resolutionStrategy {
             force("com.google.guava:guava:33.4.0-jre") // CVE for 31.1, keep to force transitive dependencies
             force("org.apache.httpcomponents.core5:httpcore5:5.3.2") // Dependency Jar Hell
+            // Force to prevent Jar Hell
+            force("org.apache.httpcomponents.core5:httpcore5:${apacheHttpVersion}")
+            force "org.apache.httpcomponents.core5:httpcore5-h2:${apacheHttpVersion}"
+            force("org.apache.httpcomponents.client5:httpclient5:${apacheHttpClientVersion}")
+            force "org.apache.logging.log4j:log4j-api:${log4jVersion}"
+            force "org.apache.logging.log4j:log4j-core:${log4jVersion}"
```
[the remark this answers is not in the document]
dbwiddis: It's a dependency of the Remote metadata client: 
https://github.com/opensearch-project/opensearch-remote-metadata-sdk/blob/c86d3c5c59819f17b397141551d9d80588f7d611/build.gradle#L90-L91

We have three choices:
 - force our version here
 - add an exclusion for log4j in the implementation line above for the client and any other dependency that uses the client transitively
 - remove our own log4j dependency and trust the upstream (but then other upstream dependencies may conflict)

build.gradle
```diff
@@ -262,10 +273,19 @@ integTest {
...
+        }
+        systemProperty "plugins.flow_framework.multi_tenancy_enabled", "true"
+    }
+
     // Only rest case can run with remote cluster
-    if (System.getProperty("tests.rest.cluster") != null) {
+    if (System.getProperty("tests.rest.cluster") != null && System.getProperty("tests.rest.tenantaware") == null) {
         filter {
             includeTestsMatching "org.opensearch.flowframework.rest.*IT"
+            excludeTestsMatching "org.opensearch.ml.rest.*TenantAwareIT"
```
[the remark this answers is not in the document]
dbwiddis: They are fine to run on secure integ tests, they run with tenant awareness false, same as normal integ tests that also run them.

They have to be excluded from the remote because we haven't set up the remote cluster, which we do below on line 330 that you commented on.

build.gradle
```diff
@@ -293,6 +313,28 @@ integTest {
...
+                    writer.write("\n# Set multitenancy\n")
+                    writer.write("plugins.flow_framework.multi_tenancy_enabled: true\n")
+                }
+                // TODO this properly uses the remote client factory but needs a remote cluster set up
+                // TODO get the endpoint from a system property
+                if (System.getProperty("tests.rest.cluster") != null) {
+                    ymlFile.withWriterAppend { writer ->
+                        writer.write("\n# Use a remote cluster\n")
+                        writer.write("plugins.flow_framework.remote_metadata_type: RemoteOpenSearch\n")
+                        writer.write("plugins.flow_framework.remote_metadata_endpoint: https://127.0.0.1:9200\n")
```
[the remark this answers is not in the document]
dbwiddis: See TODO.

src/main/java/org/opensearch/flowframework/FlowFrameworkPlugin.java
```diff
@@ -121,9 +137,28 @@ public Collection<Object> createComponents(
...
+            xContentRegistry,
+            // Here we assume remote metadata client is only used with tenant awareness.
+            // This may change in the future allowing more options for this map
+            FLOW_FRAMEWORK_MULTI_TENANCY_ENABLED.get(settings)
+                ? Map.ofEntries(
+                    Map.entry(REMOTE_METADATA_TYPE_KEY, REMOTE_METADATA_TYPE.get(settings)),
+                    Map.entry(REMOTE_METADATA_ENDPOINT_KEY, REMOTE_METADATA_ENDPOINT.get(settings)),
+                    Map.entry(REMOTE_METADATA_REGION_KEY, REMOTE_METADATA_REGION.get(settings)),
+                    Map.entry(REMOTE_METADATA_SERVICE_NAME_KEY, REMOTE_METADATA_SERVICE_NAME.get(settings)),
+                    Map.entry(TENANT_AWARE_KEY, "true"),
```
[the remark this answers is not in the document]
dbwiddis: The value of `TENANT_AWARE_KEY` is how we communicate to the remote metdata client that our plugin variable `FLOW_FRAMEWORK_MULTI_TENANCY_ENABLED` is set true.  SdkClient has this line:
```java
Boolean multiTenancy = Boolean.parseBoolean(metadataSettings.get(TENANT_AWARE_KEY));
```

Per the comment above this, we're assuming for simplicity that we're only using a remote client with multitenancy. This isn't strictly required, we could use a single tenant remote client, and we could pass the whole map in all cases and put the value of `FLOW_FRAMEWORK_MULTI_TENANCY_ENABLED.get(settings)` here instead of `"true"`.  I expect this to change in the future; trying to reduce the combinatorial explosion of tests required for GA.

src/main/java/org/opensearch/flowframework/FlowFrameworkPlugin.java
```diff
@@ -121,9 +137,28 @@ public Collection<Object> createComponents(
...
+                    Map.entry(TENANT_ID_FIELD_KEY, TENANT_ID_FIELD)
+                )
+                : Collections.emptyMap(),
+            // TODO: Find a better thread pool or make one
+            client.threadPool().executor(ThreadPool.Names.GENERIC)
```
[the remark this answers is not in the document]
dbwiddis: We don't need it for the local client (NodeClient) implementation as this executor is completely ignored in that case.  It only impacts remote.  I'm actually trying to remove the requirement for an executor for remote as well so this might just go away entirely.

IF there were more than one person trying to do all these things it'd probably already be gone :)

src/main/java/org/opensearch/flowframework/FlowFrameworkPlugin.java
```diff
@@ -121,9 +137,28 @@ public Collection<Object> createComponents(
...
+        );
```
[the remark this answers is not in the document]
dbwiddis: Which other file?  IT doesn't look that complex here.

src/main/java/org/opensearch/flowframework/FlowFrameworkPlugin.java
```diff
@@ -206,21 +253,21 @@ public List<ExecutorBuilder<?>> getExecutorBuilders(Settings settings) {
             new ScalingExecutorBuilder(
                 WORKFLOW_THREAD_POOL,
                 1,
-                Math.max(2, OpenSearchExecutors.allocatedProcessors(settings) - 1),
+                Math.max(10, OpenSearchExecutors.allocatedProcessors(settings) - 1),
```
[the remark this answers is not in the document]
dbwiddis: Our previous settings (2/4/2) were insufficient and tests were failing with blocking threads, particularly the provisioning.  This current (10/20/10) was part of my most recent commit to test if it fixed the flakiness of integ tests (it didn't so I'll probably reduce to 4/8/4 again), and the flakiness was do to using search to verify indexing without forcing a refresh (on my local, not yet committed...)

We probably need to make this a setting in any case, it will need to be much higher for multitenancy.

src/main/java/org/opensearch/flowframework/common/FlowFrameworkSettings.java
```diff
@@ -83,6 +91,53 @@ public class FlowFrameworkSettings {
...
+    // ./bin/opensearch -Eplugins.flow_framework.multi_tenancy_enabled=true
+    // 2. As a system property using OPENSEARCH_JAVA_OPTS (overrides opensearch.yml):
+    // export OPENSEARCH_JAVA_OPTS="-Dplugins.flow_framework.multi_tenancy_enabled=true"
+    // ./bin/opensearch
+    // Or inline when starting OpenSearch:
+    // OPENSEARCH_JAVA_OPTS="-Dplugins.flow_framework.multi_tenancy_enabled=true" ./bin/opensearch
+    // 3. In the opensearch.yml configuration file:
+    // plugins.flow_framework.multi_tenancy_enabled: true
+    // After setting it, a full cluster restart is required for the changes to take effect.
+    /** This setting determines if multitenancy is enabled */
```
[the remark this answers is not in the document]
dbwiddis: Actually it should be in the javadoc.  ML Commons has it that way after review; I copied this over from there. before the change.  Insert this in the long list of things I'm trying to do solo.

src/main/java/org/opensearch/flowframework/indices/FlowFrameworkIndicesHandler.java
```diff
@@ -371,39 +392,51 @@ public void initializeConfigIndex(ActionListener<Boolean> listener) {
...
-
-            ) {
-                request.source(state.toXContent(builder, ToXContent.EMPTY_PARAMS)).setRefreshPolicy(WriteRequest.RefreshPolicy.IMMEDIATE);
-                request.id(workflowId);
-                client.index(request, ActionListener.runBefore(listener, context::restore));
-            } catch (Exception e) {
-                String errorMessage = "Failed to put state index document";
-                logger.error(errorMessage, e);
-                listener.onFailure(new FlowFrameworkException(errorMessage, ExceptionsHelper.status(e)));
+            PutDataObjectRequest putRequest = PutDataObjectRequest.builder()
```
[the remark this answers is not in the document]
dbwiddis: Probably, yes.  Here's where I'm trying to strike a balance between major refactoring of our already duplicated code, and minimizing the diff to ease review.

src/main/java/org/opensearch/flowframework/indices/FlowFrameworkIndicesHandler.java
dbwiddis: Already stumbled on this and changed it on my local.

src/main/java/org/opensearch/flowframework/FlowFrameworkPlugin.java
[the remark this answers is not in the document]
amitgalitz: I think a minimum of 10 is pretty high? we might not have enough processors. We usually implemented in my past experience a Math.max(1, OpenSearchExecutors.allocatedProcessors(settings) /2 ) to take up only at most half the processors. we can indeed change that to -1 depending how we feel.

src/main/java/org/opensearch/flowframework/model/WorkflowState.java
```diff
@@ -314,6 +336,9 @@ public static WorkflowState updateExistingWorkflowState(WorkflowState existingSt
         if (stateWithNewFields.resourcesCreated() != null) {
             builder.resourcesCreated(stateWithNewFields.resourcesCreated());
         }
+        if (stateWithNewFields.getTenantId() != null) {
```
amitgalitz, line 339: should this cover bwc issues? Should we test for that manually or with a test as a TODO at least?

src/main/java/org/opensearch/flowframework/transport/handler/SearchHandler.java
```diff
@@ -83,16 +102,40 @@ public void validateRole(
...
+                    SearchResponse searchResponse = SearchResponse.fromXContent(r.parser());
+                    logger.info(Arrays.toString(request.indices()) + " search complete: {}", searchResponse.getHits().getTotalHits());
+                    listener.onResponse(searchResponse);
+                } catch (Exception e) {
+                    logger.error("Failed to parse search response", e);
+                    listener.onFailure(new FlowFrameworkException("Failed to parse search response", INTERNAL_SERVER_ERROR));
+                }
+            } else {
+                Exception cause = SdkClientUtils.unwrapAndConvertToException(throwable);
+                logger.error(Arrays.toString(request.indices()) + " search failed", cause);
```
amitgalitz, line 136: nit:
logger.error("Search failed for indices: {}", Arrays.toString(request.indices()), cause); instead of concatenation

src/main/java/org/opensearch/flowframework/util/EncryptorUtils.java
```diff
@@ -66,38 +75,46 @@ public class EncryptorUtils {
     // https://github.com/aws/aws-encryption-sdk-java/issues/1879
     private static final String WRAPPING_ALGORITHM = "AES/GCM/NOPADDING";
 
+    // concurrent map can't have null as a key. This key is to support single tenancy
+    public static final String DEFAULT_TENANT_ID = "03000200-0400-0500-0006-000700080009";
```
amitgalitz, line 79: What does this random string represent? do we just need to have one single random ID that is the same when we don't have a tenant ID? should this be generated differently for each cluster?

src/main/java/org/opensearch/flowframework/util/EncryptorUtils.java
```diff
@@ -175,12 +192,23 @@ private Template processTemplateCredentials(Template template, Function<String,
...
+     * @param tenantId The tenant id. If null, encrypts for the default tenant id.
      * @return the encrypted credential
      */
-    String encrypt(final String credential) {
-        initializeMasterKeyIfAbsent();
+    String encrypt(final String credential, @Nullable String tenantId) {
+        CountDownLatch latch = new CountDownLatch(1);
+        initializeMasterKeyIfAbsent(tenantId).whenComplete((v, throwable) -> latch.countDown());
+        try {
+            if (!latch.await(5, TimeUnit.SECONDS)) {
```
amitgalitz, line 202: can be moved to a constant

src/main/java/org/opensearch/flowframework/util/EncryptorUtils.java
```diff
@@ -248,99 +285,146 @@ public Template redactTemplateSecuredFields(User user, Template template) {
...
-                        listener.onResponse(true);
-                    } catch (FlowFrameworkException e) {
-                        listener.onFailure(e);
-                    }
+                    // Set generated key to master
+                    logger.info("Config has been initialized successfully");
+                    setMasterKey(tenantId, config.masterKey());
+                    listener.onResponse(true);
+                } else {
+                    Exception exception = SdkClientUtils.unwrapAndConvertToException(throwable);
```
amitgalitz, line 334: we should also do context.restore() if theres an exception right?

src/main/java/org/opensearch/flowframework/util/EncryptorUtils.java
```diff
@@ -248,99 +285,146 @@ public Template redactTemplateSecuredFields(User user, Template template) {
...
+                    resultFuture.completeExceptionally(
+                        new FlowFrameworkException("Failed to get master key from config index", ExceptionsHelper.status(exception))
+                    );
+                }
+            });
+            return resultFuture;
+        }
+    }
+
+    private String hashString(String input) {
```
amitgalitz, line 415: Could add input validation that its not null

src/main/java/org/opensearch/flowframework/util/EncryptorUtils.java
```diff
@@ -248,99 +285,146 @@ public Template redactTemplateSecuredFields(User user, Template template) {
...
+        });
+    }
+
+    private void generateAndIndexNewMasterKey(String tenantId, ActionListener<Boolean> listener) {
+        String masterKeyId = tenantId == null ? MASTER_KEY : MASTER_KEY + "_" + hashString(tenantId);
+        Config config = new Config(generateMasterKey(), Instant.now());
+        PutDataObjectRequest putRequest = PutDataObjectRequest.builder()
+            .index(CONFIG_INDEX)
+            .id(masterKeyId)
+            .tenantId(tenantId)
```
amitgalitz, line 322: There's no issue here when tenantID is null? we are passing the original tenantId which could be null in

src/main/java/org/opensearch/flowframework/util/EncryptorUtils.java
```diff
@@ -248,99 +285,146 @@ public Template redactTemplateSecuredFields(User user, Template template) {
...
-                    } catch (FlowFrameworkException e) {
-                        listener.onFailure(e);
-                    }
+                    // Set generated key to master
+                    logger.info("Config has been initialized successfully");
+                    setMasterKey(tenantId, config.masterKey());
+                    listener.onResponse(true);
+                } else {
+                    Exception exception = SdkClientUtils.unwrapAndConvertToException(throwable);
+                    logger.error("Failed to index new master key in config", exception);
```
amitgalitz, line 335: nit: add tenant id to error message maybe?

src/main/java/org/opensearch/flowframework/util/RestActionUtils.java
```diff
@@ -0,0 +1,54 @@
...
+        if (isMultiTenancyEnabled) {
+            Map<String, List<String>> headers = restRequest.getHeaders();
+            if (headers.containsKey(CommonValue.TENANT_ID_HEADER)) {
+                List<String> tenantIdList = headers.get(CommonValue.TENANT_ID_HEADER);
+                if (tenantIdList != null && !tenantIdList.isEmpty()) {
+                    String tenantId = tenantIdList.get(0);
+                    if (tenantId != null) {
+                        return tenantId;
+                    } else {
+                        throw new FlowFrameworkException("Tenant ID can't be null", RestStatus.FORBIDDEN);
```
amitgalitz, line 42: Should this be BAD_REQUEST instead?

build.gradle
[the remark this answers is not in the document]
dbwiddis: Choice 4: update remote metadata client to pull from OpenSearch version catalog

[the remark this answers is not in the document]
dbwiddis: Added a TODO, I'll handle this along with #992

src/main/java/org/opensearch/flowframework/FlowFrameworkPlugin.java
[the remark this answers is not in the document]
dbwiddis: > Do you mean instead of an executor we just make the call like in other locations which auto uses the plugin threadpool?

I mean chaining action listeners which use the SAME threadpool and are throttled by Netty/IO.

TLDR: I hope to get rid of the need for the executor in the upstream client. I started working on it but have other priorities.   If you're bored you can help.

src/main/java/org/opensearch/flowframework/FlowFrameworkPlugin.java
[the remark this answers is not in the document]
dbwiddis: > I think a minimum of 10 is pretty high?

We're not actually using CPU for these, we're waiting on `actionGet()` blocking calls.

src/main/java/org/opensearch/flowframework/indices/FlowFrameworkIndicesHandler.java
```diff
@@ -332,37 +337,53 @@ public void putTemplateToGlobalContext(Template template, ActionListener<IndexRe
...
+        PutDataObjectRequest request = PutDataObjectRequest.builder()
+            .index(GLOBAL_CONTEXT_INDEX)
+            .id(documentId)
+            .tenantId(template.getTenantId())
+            .dataObject(encryptorUtils.encryptTemplateCredentials(template))
+            .build();
+        try (ThreadContext.StoredContext context = client.threadPool().getThreadContext().stashContext()) {
+            sdkClient.putDataObjectAsync(request).whenComplete((r, throwable) -> {
+                context.restore();
+                if (throwable == null) {
```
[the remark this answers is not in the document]
dbwiddis: It's essentially the same test. Either `r` or `throwable` will be null.  I've been using this pattern everywhere.

In the future I may update the abstract superclass of the metadata client to remove this entirely and put it back to "action listener" style.  If only I had more time.

build.gradle
```diff
@@ -37,6 +37,9 @@ buildscript {
         isSameMajorVersion = opensearch_version.split("\\.")[0] == bwcVersionShort.split("\\.")[0]
         swaggerVersion = "2.1.25"
         swaggerCoreVersion = "2.2.28"
+        apacheHttpVersion = "5.3.1"
+        apacheHttpClientVersion = "5.4.1"
+        log4jVersion = "2.24.2"
```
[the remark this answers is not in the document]
dbwiddis: yes we should, I've added a TODO here and will address this with #992

src/main/java/org/opensearch/flowframework/model/WorkflowState.java
dbwiddis: I don't understand the question here.  without tenant awareness this will probably always be null anyway.  Our unit tests use null nearly everywhere.  Tenant aware integ tests test this thoroughly when it's set.

src/main/java/org/opensearch/flowframework/util/EncryptorUtils.java
dbwiddis: It's not random! 😁 

https://arstechnica.com/civis/threads/uuid-you-keep-using-that-word-i-do-not-think-it-means-what-you-think-it-means.1491688/

It's intended to be a constant, and can be any constant we want, including the empty string.  As the comment says, a concurrent hash map can not have a null key.

When multitenancy is not enabled, `tenantId` will be null and will then invoke this code:
```java
    /**
     * Sets the master key
     * @param tenantId The tenant id. If null, sets the key for the default id.
     * @param masterKey the master key
     */
    void setMasterKey(@Nullable String tenantId, String masterKey) {
        this.tenantMasterKeys.put(Objects.requireNonNullElse(tenantId, DEFAULT_TENANT_ID), masterKey);
    }

    /**
     * Returns the master key
     * @param tenantId The tenant id. If null, gets the key for the default id.
     * @return the master key
     */
    String getMasterKey(@Nullable String tenantId) {
        return tenantMasterKeys.get(Objects.requireNonNullElse(tenantId, DEFAULT_TENANT_ID));    
```

The other way to handle it is to have a separate tenantMasterKey `String` variable, and do if checks on tenant ID being null to decide whether to use it or the map.  in which case one of either the separate variable or the map will be unused and we'll have to add if/then checks.

I prefer this method to reduce the number of `null` variables floating around and unused fields in a class that someone writing new code may reference thinking they're doing the right thing.

src/main/java/org/opensearch/flowframework/util/EncryptorUtils.java
dbwiddis: the number 5?

src/main/java/org/opensearch/flowframework/util/EncryptorUtils.java
dbwiddis: Moved it outside the conditional.  It's in a try-with-resources block anyway so is technically unneeded in either.

src/main/java/org/opensearch/flowframework/util/EncryptorUtils.java
dbwiddis: if tenant awareness isn't enabled, null is the expected value.  The SdkClient validates that tenant ID is non null if multitenancy is enabled, so it's required in that case even if not used otherwise.

src/main/java/org/opensearch/flowframework/util/EncryptorUtils.java
amitgalitz: I see, I guess it seemed very out of place in the code, maybe it can be a non dynamic setting but this is okay also?

src/main/java/org/opensearch/flowframework/util/RestActionUtils.java
dbwiddis: I want to retain consistency with ML Commons, who implemented this first and I copied:  https://github.com/opensearch-project/ml-commons/blob/06a2b40b07df086c5474574c7a63128e4521c7d7/plugin/src/main/java/org/opensearch/ml/utils/TenantAwareHelper.java#L77-L98

src/main/java/org/opensearch/flowframework/util/EncryptorUtils.java
```diff
@@ -248,99 +285,146 @@ public Template redactTemplateSecuredFields(User user, Template template) {
...
+            } else {
+                // Shouldn't get here
+                listener.onFailure(exception);
+            }
+            return null;
+        });
+    }
 
+    private void generateAndIndexNewMasterKey(String tenantId, ActionListener<Boolean> listener) {
+        String masterKeyId = tenantId == null ? MASTER_KEY : MASTER_KEY + "_" + hashString(tenantId);
```
owaiskazi19, line 317: This is new code. Trying to understand more here. Why are we changing the way master key is stored currently?

src/main/java/org/opensearch/flowframework/util/TenantAwareHelper.java
```diff
@@ -0,0 +1,59 @@
...
+     * Validates the tenant ID based on the multi-tenancy feature setting.
+     *
+     * @param isMultiTenancyEnabled whether the multi-tenancy feature is enabled.
+     * @param tenantId The tenant ID to validate.
+     * @param listener The action listener to handle failure cases.
+     * @return true if the tenant ID is valid or if multi-tenancy is not enabled; false if the tenant ID is invalid and multi-tenancy is enabled.
+     */
+    public static boolean validateTenantId(boolean isMultiTenancyEnabled, String tenantId, ActionListener<?> listener) {
+        if (isMultiTenancyEnabled && tenantId == null) {
+            listener.onFailure(new FlowFrameworkException("You don't have permission to access this resource", RestStatus.FORBIDDEN));
```
owaiskazi19, line 32:
```suggestion
            listener.onFailure(new FlowFrameworkException("No permission to access this resource", RestStatus.FORBIDDEN));
```

owaiskazi19 (commented):
There are discrepancies at few places between `INTERNAL_SERVER_ERROR` and `BAD_REQUEST` rest status for the request calls

build.gradle
```diff
@@ -262,10 +273,19 @@ integTest {
     systemProperty('user', user)
     systemProperty('password', password)
 
+    // Only tenant aware test if set
+    if (System.getProperty("tests.rest.tenantaware") == "true") {
+        filter {
+            includeTestsMatching "org.opensearch.flowframework.*TenantAwareIT"
+        }
+        systemProperty "plugins.flow_framework.multi_tenancy_enabled", "true"
```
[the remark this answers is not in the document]
owaiskazi19: Can we make the setting as dynamic to avoid a full cluster restart?

build.gradle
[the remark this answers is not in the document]
owaiskazi19: In secure integ tests we just run `FlowFrameworkSecureRestApiIT` and exclude normal integ tests as well. Should be the same for tenant integ tests

src/main/java/org/opensearch/flowframework/FlowFrameworkPlugin.java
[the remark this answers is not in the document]
owaiskazi19: A new file to initialize SDKClient but it's fine to do that later.

src/main/java/org/opensearch/flowframework/util/EncryptorUtils.java
```diff
@@ -175,12 +192,23 @@ private Template processTemplateCredentials(Template template, Function<String,
     /**
      * Encrypts the given credential
      * @param credential the credential to encrypt
+     * @param tenantId The tenant id. If null, encrypts for the default tenant id.
      * @return the encrypted credential
      */
-    String encrypt(final String credential) {
-        initializeMasterKeyIfAbsent();
+    String encrypt(final String credential, @Nullable String tenantId) {
+        CountDownLatch latch = new CountDownLatch(1);
```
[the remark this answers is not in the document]
dbwiddis: I tried to copy the ML Commons implementation as much as possible for the encryption parts.

src/main/java/org/opensearch/flowframework/util/EncryptorUtils.java
dbwiddis: Not sure -1 is more or less confusing. I'd prefer the empty string (which is guaranteed to be interned as a JVM source code constant) or a descriptive string like "DEFAULT_TENANT_ID".  I put empty for now.  

With regard to the handling of the master key:

Current state in single tenancy is a single `String` constant for the master key.

In multi tenancy we want one master key per tenant, as each one will want different encryption.  This is implemented using key-value pairs, where the key is the tenant ID, and the value is the encryption key.

In memory, this is stored as a map, with the tenant ID as the key to the map.

However, the map is temporary/in-memory in the (stateless) multitenant node, so somewhat serves as a cache.  We need a permanent storage of the key-value (tenant-masterkey) relationship in the config index.  We already used the config index with the `MASTER_KEY` document ID storing the key and its creation time.  We are just going to create more documents, one for each tenant.

When the tenant ID is null (no multitenancy) we continue to use the same, pre-existing `MASTER_KEY` document ID, so we have backward compatibility.   Under multitenancy we will have additional document IDs of the master key appended with a unique string encoding the tenant ID.   Since our tenant IDs are likely to include characters such as `:` which are illegal in document IDs, we use URL-safe Base64 encoding to hash the tenant ID.

So one tenant will have a document ID of "master_key_jdIWel8Tr_3" and another one "master_key_LTw9d_eWr" and so forth.

So when looking up the key to encrypt/decrypt, the code will first check the map to see if the key exists. If so, great.  If not it will fetch it by document ID from the config index (and then cache it in the map for future use).  When creating the key in the first place it's both put in the map and the index.

src/main/resources/mappings/workflow-state.json
```diff
@@ -1,7 +1,7 @@
 {
   "dynamic": false,
   "_meta": {
-    "schema_version": 3
+    "schema_version": 4
```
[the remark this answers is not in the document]
dbwiddis: I don't think it's necessary or relevant, or really even possible.  

We do not envision any scenario where a user would upgrade from 2.18- to 2.19+ and expect the Tenant ID to be anything but null.
 - TenantID is forced to null unless multitenancy is enabled 
 - Enabling multitenancy has to happen before the cluster starts up; it's all part of the tests where we pass `-Dtests.rest.multitenancy=true`.  So generic BWC won't work.  And those tests only exist in 2.19+.

For any scenario where we want backwards compatibility, the value will be null.  It's possible to test whether an index mapping contains a field but that's not really associated with backwards compatibility.  What we would really want to know is if we have a 2.18 cluster and update it to 2.19, whether the new mapping "works".  But the tenant ID will be null, and any normal functions/parsers to extract the tenant ID will have the same value.

src/main/java/org/opensearch/flowframework/util/RestActionUtils.java
```diff
@@ -0,0 +1,54 @@
...
+     * Finds the tenant id in the REST Headers
+     * @param isMultiTenancyEnabled whether multitenancy is enabled
+     * @param restRequest the RestRequest
+     * @return The tenant ID from the headers or null if multitenancy is not enabled
+     */
+    public static String getTenantID(Boolean isMultiTenancyEnabled, RestRequest restRequest) {
+        if (isMultiTenancyEnabled) {
+            Map<String, List<String>> headers = restRequest.getHeaders();
+            if (headers.containsKey(CommonValue.TENANT_ID_HEADER)) {
+                List<String> tenantIdList = headers.get(CommonValue.TENANT_ID_HEADER);
```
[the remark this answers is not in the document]
dbwiddis: I just re-copied the new ML Commons version and also moved it to the TenantAwareHelper class.

src/main/java/org/opensearch/flowframework/model/WorkflowState.java
amitgalitz: I see you covered this bwc in other comments

src/main/java/org/opensearch/flowframework/util/EncryptorUtils.java
amitgalitz: thought it was used all over, I see it was just 2 uses, can be ignored

src/main/java/org/opensearch/flowframework/transport/GetWorkflowStateTransportAction.java
```diff
@@ -108,41 +124,20 @@ protected void doExecute(Task task, GetWorkflowStateRequest request, ActionListe
...
-                String errorMessage = ParameterizedMessageFactory.INSTANCE.newMessage("Failed to get workflow status of: {}", workflowId)
-                    .getFormattedMessage();
-                logger.error(errorMessage, e);
-                listener.onFailure(new FlowFrameworkException(errorMessage, RestStatus.NOT_FOUND));
-            }
-        }), context::restore));
+        flowFrameworkIndicesHandler.getWorkflowState(workflowId, tenantId, ActionListener.wrap(workflowState -> {
+            GetWorkflowStateResponse workflowStateResponse = new GetWorkflowStateResponse(workflowState, request.getAll());
+            listener.onResponse(workflowStateResponse);
+        }, listener::onFailure), context);
```
amitgalitz, line 141: I think we should restore context here after making a search to system index

dbwiddis: The previous code used an `ActionListener.runBefore()` call which restored the context immediately prior to completing the action listener. In this case we are making a call to `flowFrameworkIndicesHandler.getWorkflowState()` passing the context along.

That call immediately restores the context after performing the GET, regardless of how the action listener is completed.  
```java
    public void getWorkflowState(String workflowId, String tenantId, ActionListener<WorkflowState> listener, StoredContext context) {
        GetDataObjectRequest getRequest = GetDataObjectRequest.builder()
            .index(WORKFLOW_STATE_INDEX)
            .id(workflowId)
            .tenantId(tenantId)
            .build();
        sdkClient.getDataObjectAsync(getRequest).whenComplete((r, throwable) -> {
            context.restore();
```
This is actually earlier than the previous `runBeflore()` behavior which would have waited to restore it until just before the one `onResponse()` call or the multiple `onFailure()` calls.

build.gradle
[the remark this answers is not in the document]
amitgalitz: what happens on B/G, if we have multitenancy set to true and people have tenant ID and then they toggle it off and on cluster restart everyone can see everyone's data? Is that just something that is expected if its toggeled off.
Also otherway around if you turn it on, would all the other indices have tenant ID set to the default so it would all just be considered one tenant and new requests would have different tenants?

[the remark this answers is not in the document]
dbwiddis: > what happens on B/G, if we have multitenancy set to true and people have tenant ID and then they toggle it off and on cluster restart everyone can see everyone's data?

That would happen, if "they" toggled it off.  But the setting will be on the server, so "they" will need write access to `opensearch.yml` to do so.

> Is that just something that is expected if its toggeled off. Also otherway around if you turn it on, would all the other indices have tenant ID set to the default so it would all just be considered one tenant and new requests would have different tenants?

If toggled on, any previous data that has a `null` tenant ID in the documents will not be accessible to any tenant.

This setting is designed to be set and always on/always off for a particular configuration, and never toggled.

owaiskazi19 (approved):
Thanks for addressing the comments. Not blocking the PR for the ones that can be addressed later

amitgalitz (approved):

PR #990, opened 2025-01-08
Add synchronous execution option to workflow provisioning

### LINKED ISSUE #967: [FEATURE] Add option to provision synchronously

Presently, when provisioning a workflow (via either the [provision API](https://opensearch.org/docs/latest/automating-configurations/api/provision-workflow/), [create API with provision or param](https://opensearch.org/docs/latest/automating-configurations/api/create-workflow/), the REST call returns immediately with a 200 (OK) response, but the caller must then poll the [Workflow Status API](https://opensearch.org/docs/latest/automating-configurations/api/get-workflow-status/) to monitor the status of provisioning.

This asynchronous execution of provisioning was intentional to provide the ability for a front end to obtain status throughout provisioning, possibly including a progress bar or similar, and because some provisioning processes take longer than the expected time for a REST response.

However, there are some use cases where the user may be willing to wait for a completed response, and not have to poll.  This would be particularly useful in cases similar to the ML Commons Remote Model deployment which provides such a synchronous API.

Add optional parameters to the create and provision work flow APIs to wait for the request to complete, with a timeout.  Other OpenSearchAPIs use [`wait_for_completion`](https://opensearch.org/docs/latest/api-reference/snapshots/create-snapshot/) and [`wait_for_completion_timeout`](https://opensearch.org/docs/latest/search-plugins/async/index/) so I'd suggest these names.

Alter the Provision Workflow Transport action, when this parameter is present, to wait to return until provisioning is complete (or the timeout).

A separate wrapper API that does the retries internally.

This would be a much simpler approach for automation tools, that would not require them to code all the polling themselves.

junweid62:
Jan 14 Revision
Added synchronous execution option to reprovision 

### Description
This PR introduces a new `wait_for_completion_timeout ` feature to the Provision Workflow API in the OpenSearch Flow Framework. The feature allows users to control whether the API call waits for the entire workflow provisioning process to complete before returning a response.

#### What’s Changed:
1. Added support for the wait_for_completion_timeout parameter in the REST layer (RestProvisionWorkflowAction).

- Accepts a time duration value (e.g., 30s, 1m).
- If the workflow is provisioned within the specified timeout, the API returns the created resources (same response as GetWorkflowStatus).
- If the timeout is reached before provisioning completes, the API returns the workflow state without waiting further.

3. Updated the transport layer (ProvisionWorkflowTransportAction) to handle the timeout logic and ensure correct behavior during synchronous provisioning.

Success Response:
```
{
    "workflow_id": "K13IR5QBEpCfUu_-AQdU",
    "state": "COMPLETED",
    "resources_created": [
        {
            "workflow_step_id": "create_connector_1",
            "resource_id": "LF3IR5QBEpCfUu_-Awd_",
            "resource_type": "connector_id"
        },
        {
            "workflow_step_id": "register_model_2",
            "resource_id": "L13IR5QBEpCfUu_-BQdI",
            "resource_type": "model_id"
        },
        {
            "workflow_step_name": "deploy_model",
            "workflow_step_id": "deploy_model_3",
            "resource_id": "L13IR5QBEpCfUu_-BQdI",
            "resource_type": "model_id"
        }
    ]
}
```
TimeOut Response:
```
{
    "workflow_id": "SmACR5QBdrR0lYdqgHa9",
    "state": "PROVISIONING",
    "resources_created": [
        {
            "workflow_step_id": "create_connector_1",
            "resource_type": "connector_id",
            "resource_id": "S2ACR5QBdrR0lYdqgXYK"
        },
        {
            "workflow_step_id": "register_model_2",
            "resource_type": "model_id",
            "resource_id": "TWACR5QBdrR0lYdqgXZ-"
        }
    ]
}
```
#### Areas of Concern:
I have a few parts of the implementation that I believe can be further improved, particularly in ProvisionWorkflowTransportAction. Some of the logic feels a bit verbose and might not be the most efficient way to handle the timeout and synchronous execution. I’d appreciate the feedback from reviewers.
### Related Issues
Resolves #967 

### Check List
- [x] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
- [x] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).

### DIFF
--- a/src/main/java/org/opensearch/flowframework/transport/CreateWorkflowTransportAction.java
+++ b/src/main/java/org/opensearch/flowframework/transport/CreateWorkflowTransportAction.java
@@ -52,7 +52,6 @@ import java.util.Map;
 import static java.lang.Boolean.FALSE;
 import static org.opensearch.flowframework.common.CommonValue.GLOBAL_CONTEXT_INDEX;
 import static org.opensearch.flowframework.common.CommonValue.PROVISIONING_PROGRESS_FIELD;
-import static org.opensearch.flowframework.common.CommonValue.PROVISION_TIMEOUT_FIELD;
 import static org.opensearch.flowframework.common.CommonValue.STATE_FIELD;
 import static org.opensearch.flowframework.common.CommonValue.WAIT_FOR_COMPLETION_TIMEOUT;
 import static org.opensearch.flowframework.common.FlowFrameworkSettings.FILTER_BY_BACKEND_ROLES;
--- a/src/main/java/org/opensearch/flowframework/transport/CreateWorkflowTransportAction.java
+++ b/src/main/java/org/opensearch/flowframework/transport/CreateWorkflowTransportAction.java
@@ -251,7 +251,8 @@ public class CreateWorkflowTransportAction extends HandledTransportAction<Workfl
                                                     WorkflowRequest workflowRequest = new WorkflowRequest(
                                                         globalContextResponse.getId(),
                                                         null,
-                                                        request.getParams()
+                                                        request.getParams(),
+                                                        request.getWaitForCompletionTimeout()
                                                     );
                                                     logger.info(
                                                         "Provisioning parameter is set, continuing to provision workflow {}",
@@ -261,7 +262,18 @@ public class CreateWorkflowTransportAction extends HandledTransportAction<Workfl
                                                         ProvisionWorkflowAction.INSTANCE,
                                                         workflowRequest,
                                                         ActionListener.wrap(provisionResponse -> {
-                                                            listener.onResponse(new WorkflowResponse(provisionResponse.getWorkflowId()));
+                                                            if (request.getWaitForCompletionTimeout() != null) {
+                                                                listener.onResponse(
+                                                                    new WorkflowResponse(
+                                                                        provisionResponse.getWorkflowId(),
+                                                                        provisionResponse.getWorkflowState()
+                                                                    )
+                                                                );
+                                                            } else {
+                                                                listener.onResponse(
+                                                                    new WorkflowResponse(provisionResponse.getWorkflowId())
+                                                                );
+                                                            }
                                                         }, exception -> {
                                                             String errorMessage = "Provisioning failed.";
                                                             logger.error(errorMessage, exception);
--- a/src/main/java/org/opensearch/flowframework/transport/CreateWorkflowTransportAction.java
+++ b/src/main/java/org/opensearch/flowframework/transport/CreateWorkflowTransportAction.java
@@ -53,6 +53,7 @@ import static java.lang.Boolean.FALSE;
 import static org.opensearch.flowframework.common.CommonValue.GLOBAL_CONTEXT_INDEX;
 import static org.opensearch.flowframework.common.CommonValue.PROVISIONING_PROGRESS_FIELD;
 import static org.opensearch.flowframework.common.CommonValue.STATE_FIELD;
+import static org.opensearch.flowframework.common.CommonValue.WAIT_FOR_COMPLETION_TIMEOUT;
 import static org.opensearch.flowframework.common.FlowFrameworkSettings.FILTER_BY_BACKEND_ROLES;
 import static org.opensearch.flowframework.util.ParseUtils.checkFilterByBackendRoles;
 import static org.opensearch.flowframework.util.ParseUtils.getUserContext;
@@ -248,11 +249,14 @@ public class CreateWorkflowTransportAction extends HandledTransportAction<Workfl
                                             ActionListener.wrap(stateResponse -> {
                                                 logger.info("Creating state workflow doc: {}", globalContextResponse.getId());
                                                 if (request.isProvision()) {
+                                                    String waitForTimeCompletion = request.getParams()
+                                                        .getOrDefault(WAIT_FOR_COMPLETION_TIMEOUT, TimeValue.MINUS_ONE.toString());
                                                     WorkflowRequest workflowRequest = new WorkflowRequest(
                                                         globalContextResponse.getId(),
                                                         null,
                                                         request.getParams(),
-                                                        request.getWaitForCompletionTimeout()
+                                                        // todo : what is this setting name represent?
+                                                        TimeValue.parseTimeValue(waitForTimeCompletion, "provision.timout")
                                                     );
                                                     logger.info(
                                                         "Provisioning parameter is set, continuing to provision workflow {}",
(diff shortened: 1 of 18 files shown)


dbwiddis (commented):
Generally looks good.
 - You need to handle -1 time value; my recommendation is you use that for the default "async" rather than null
 - You need to do stream version checks for the new (optional) workflow state in the response, and the new timeout parameter in the workflow request (unless you want to just keep it in the params map).

src/main/java/org/opensearch/flowframework/util/WorkflowTimeoutUtility.java
```diff
@@ -34,6 +33,7 @@ public class WorkflowTimeoutUtility {
     private static final Logger logger = LogManager.getLogger(WorkflowTimeoutUtility.class);
     private static final TimeValue MAX_TIMEOUT_MILLIS = TimeValue.timeValueSeconds(300);
     private static final TimeValue MIN_TIMEOUT_MILLIS = TimeValue.timeValueSeconds(1);
```
dbwiddis, line 35: No reason to force that long of a timeout, people may want half a second or less, for example.  I don't see a need for anything other than making sure it isn't negative since that has special meaning.

junweid62: Thanks for the feedback! I’ve made the changes to ensure that only non-negative values are allowed for timeouts. There’s no longer any upper limit enforced on the timeout duration.

src/main/java/org/opensearch/flowframework/transport/CreateWorkflowTransportAction.java
```diff
@@ -248,10 +250,13 @@ private void createExecute(WorkflowRequest request, User user, ActionListener<Wo
                                             ActionListener.wrap(stateResponse -> {
                                                 logger.info("Creating state workflow doc: {}", globalContextResponse.getId());
                                                 if (request.isProvision()) {
+                                                    String waitForTimeCompletion = request.getParams()
+                                                        .getOrDefault(WAIT_FOR_COMPLETION_TIMEOUT, TimeValue.MINUS_ONE.toString());
```
dbwiddis, line 254: In the case this param doesn't exist (probably most of the time), we shouldn't convert this to a string just to parse it back.   This is probably a good opportunity for a `containsKey()` conditional.

src/main/java/org/opensearch/flowframework/util/WorkflowTimeoutUtility.java
```diff
@@ -0,0 +1,198 @@
...
+
+/**
+ * Utility class for managing timeout tasks in workflow execution.
+ * This class provides methods to schedule timeout handlers, wrap listeners with timeout cancellation logic,
+ * and fetch workflow states after timeouts.
+ */
+public class WorkflowTimeoutUtility {
+
+    private static final Logger logger = LogManager.getLogger(WorkflowTimeoutUtility.class);
+    private static final TimeValue MIN_TIMEOUT_MILLIS = TimeValue.timeValueSeconds(0);
```
dbwiddis, line 36: This works but you're just doing a fancy wrapper for 0.

Can probably delete this constant and just use 0 where you are calling `MIN_TIMEOUT_MILLIS.millis()`.

src/main/java/org/opensearch/flowframework/util/WorkflowTimeoutUtility.java
```diff
@@ -0,0 +1,198 @@
...
+     * @param isResponseSent An atomic boolean to ensure the response is sent only once.
+     * @param listener       The listener to notify of the workflow failure.
+     */
+    public static void handleFailure(
+        String workflowId,
+        Exception e,
+        AtomicBoolean isResponseSent,
+        ActionListener<WorkflowResponse> listener
+    ) {
+        if (isResponseSent.compareAndSet(false, true)) {
```
dbwiddis, line 161: For those of us who don't use AtomicBooleans often, a comment above this saying what we're doing would be helpful.   (I think I can guess from the else clause.)

dbwiddis (approved):
LGTM with a few suggestions.

src/main/java/org/opensearch/flowframework/transport/CreateWorkflowTransportAction.java
```diff
@@ -248,10 +249,19 @@ private void createExecute(WorkflowRequest request, User user, ActionListener<Wo
                                             ActionListener.wrap(stateResponse -> {
                                                 logger.info("Creating state workflow doc: {}", globalContextResponse.getId());
                                                 if (request.isProvision()) {
+                                                    // default to minus one indicate async execution
+                                                    TimeValue waitForTimeCompletion = TimeValue.MINUS_ONE;
```
joshpalis, line 253: I see this block is repeated for both provision/reprovision cases. We can probably retrieve this once before any additional processing of the request

src/main/java/org/opensearch/flowframework/transport/WorkflowRequest.java
```diff
@@ -62,13 +65,20 @@ public class WorkflowRequest extends ActionRequest {
      */
     private Map<String, String> params;
 
+    /**
+     * The timeout duration to wait for workflow completion.
+     * If null, the request will respond immediately with the workflowId.
+     */
+    @Nullable
```
joshpalis, line 72: I see the timeout value defaults to -1, so this wont ever be null. We can remove the nullable annotation here and update the javadoc

src/main/java/org/opensearch/flowframework/transport/WorkflowRequest.java
```diff
@@ -133,6 +187,11 @@ public WorkflowRequest(StreamInput in) throws IOException {
             this.params = Collections.emptyMap();
         }
         this.reprovision = !provision && Boolean.parseBoolean(params.get(REPROVISION_WORKFLOW));
+        // todo:change to 2.19
+        if (in.getVersion().onOrAfter(Version.CURRENT)) {
```
joshpalis, line 191: Version.V_2_19_0 already exists in 3.x and 2.x Can we use that instead of CURRENT?

src/main/java/org/opensearch/flowframework/transport/WorkflowResponse.java
```diff
@@ -44,6 +49,11 @@ public WorkflowResponse(String workflowId) {
     public WorkflowResponse(StreamInput in) throws IOException {
         super(in);
         this.workflowId = in.readString();
+        // todo : change version to 2_19_0
+        if (in.getVersion().onOrAfter(Version.CURRENT)) {
```
joshpalis, line 53: same here for the version number

src/main/java/org/opensearch/flowframework/transport/WorkflowResponse.java
```diff
@@ -54,14 +64,50 @@ public String getWorkflowId() {
...
+            .resourcesCreated(workflowState.resourcesCreated())
+            .build();
+
+    }
+
     @Override
     public void writeTo(StreamOutput out) throws IOException {
         out.writeString(workflowId);
+        // todo : change version to 2_19_0
+        if (out.getVersion().onOrAfter(Version.CURRENT)) {
```
joshpalis, line 99: same here

joshpalis (commented):
Overall looks good to me, great work implementing this feature @junweid62 . A few comments

src/main/java/org/opensearch/flowframework/transport/WorkflowRequest.java
dbwiddis: @joshpalis the issue isn't the version it's the BWC tests.

Until this updated `WorkflowRequest` class exists on 2.x, then a mixed 2.x / 3.x cluster would fail on serialization/deserialization because 2.19 would pass that check.

That said I don't think we actually test 2.x vs. 3.x so we are probably safe to just put that as 2_19_0 now (or accept any failing BWC in review)

junweid62: Change to 2.19 now :)

joshpalis (approved):

dbwiddis:
@junweid62 can you please manually backport this?

Issue #992, opened 2025-01-13
[FEATURE] Use version catalog for jackson and httpcore5 dependencies


dbwiddis:
### Is your feature request related to a problem?

We are forcing dependency resolution on imports that conflict with OpenSearch:
https://github.com/opensearch-project/flow-framework/blob/0284554eb2ed0e58d702298a6342bcdec962c843/build.gradle#L212-L213

### What solution would you like?

Thanks to https://github.com/opensearch-project/OpenSearch/pull/16284 we can now just reference the upstream version.

See for example:
 - https://github.com/opensearch-project/opensearch-remote-metadata-sdk/pull/38
 - https://github.com/opensearch-project/opensearch-remote-metadata-sdk/pull/40

### What alternatives have you considered?

Keeping track of our own version, forcing it like we do, and continually bumping it when prompted.

### Do you have any additional context?

Should probably do two separate PRs as the core5 one won't get backported.

dbwiddis:
Completed as part of #980

PR #997, opened 2025-01-15
Change Jackson dependency to get version from core to avoid jar hell

dbwiddis:
### Description

Following https://github.com/opensearch-project/OpenSearch/pull/16707, OpenSearch now has a Version Catalog where we can just import the same version of dependencies that are used in OpenSearch without causing Jar Hell or having to force our own versions.

That will also reduce the churn of version bump PRs for the same dependencies.  

This PR uses the catalog for Jackson dependencies (which will be backported).  After this is merged I'll do another one for http client (which will not be backported).

### Issues Resolved

First part of #992 

### Check List

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -2,6 +2,7 @@ import java.nio.file.Files
 import org.opensearch.gradle.testclusters.OpenSearchCluster
 import org.opensearch.gradle.testclusters.StandaloneRestIntegTestTask
 import org.opensearch.gradle.test.RestIntegTestTask
+import org.opensearch.gradle.VersionProperties
 import java.util.concurrent.Callable
 import java.nio.file.Paths
 
@@ -35,7 +36,6 @@ buildscript {
 
         isSameMajorVersion = opensearch_version.split("\\.")[0] == bwcVersionShort.split("\\.")[0]
         swaggerVersion = "2.1.24"
-        jacksonVersion = "2.18.1"
         swaggerCoreVersion = "2.2.27"
 
     }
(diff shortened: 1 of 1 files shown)


amitgalitz (approved):

build.gradle
```diff
@@ -101,6 +101,8 @@ allprojects {
     // Default to the apache license
     project.ext.licenseName = 'The Apache Software License, Version 2.0'
     project.ext.licenseUrl = 'http://www.apache.org/licenses/LICENSE-2.0.txt'
+    // Get OpenSearch version catalog
+    project.getExtensions().getExtraProperties().set("versions", VersionProperties.getVersions())
```
amitgalitz, line 105: nit: not 100% sure this is a must as AD uses versions.<package name> toget versions from opensearch core but doesn't have this line https://github.com/opensearch-project/anomaly-detection/blob/main/build.gradle#L215

dbwiddis: That said, I just copied from a working config and it seems removing that line it still builds.

owaiskazi19 (approved):

build.gradle
```diff
@@ -35,9 +36,7 @@ buildscript {
 
         isSameMajorVersion = opensearch_version.split("\\.")[0] == bwcVersionShort.split("\\.")[0]
         swaggerVersion = "2.1.24"
-        jacksonVersion = "2.18.1"
-        swaggerCoreVersion = "2.2.27"
-
+        swaggerCoreVersion = "2.2.28"
```
owaiskazi19, line 39: Can remove this line as https://github.com/opensearch-project/flow-framework/pull/1005 is merged.

Issue #986, opened 2024-12-28
[BUG] Can't get or delete workflow state without template if filtering by user


dbwiddis:
### What is the bug?

In most circumstances, we pair a workflow state document with a template document, sharing the same document id.

Our API permits users to delete the template while keeping the state document (by default; users can specify to clear status to delete both).

However, when access control is enabled, only the template contains the `User` that can be used to validate permission for various APIs including:
 - Provision workflow
 - Reprovision workflow
 - Get workflow
 - Delete workflow
 - Deprovision workflow
 - Get workflow state

For the Provision, Reprovision, and Get workflow cases, failure is expected if the template has been deleted.  However, for the other three APIs, the lack of template means there is no way to even view, or delete the state document:
 - Delete workflow may not seem needed since the template is already deleted, but this is one way to delete the state document using the clear status parameter
 - Deprovisioning deletes the state on success if the template doesn't exist, but you can't even execute deprovisioning
 - Get workflow state has no concept of the user so no check can be conducted

In summary, we have an orphaned document that the user can't either view or delete.

### How can one reproduce the bug?

1. Enable filtering by user
2. Create and provision a workflow
3. Delete the workflow _without_ specifying `clear_status=true`
4. Attempt any of the above APIs
   - Delete workflow _with_ `clear_status=true` fails as access can't be checked
   - Deprovision workflow fails as as access can't be checked
   - Get workflow state fails as access can't be checked

### What is the expected behavior?

User can clean up all documents in their cluster without direct access to system indices

### Do you have any additional context?

Discovered this in the course of integ tests for #980, which adds additional complexity of tenant id validation.  As tenant id is also added to the workflow state document, it's possible to work around it by falling back to tenant id check of workflow state in case of a template 404.  This suggests we could/should also add the User to the workflow state document in order to permit the same fallbacks.

PR #998, opened 2025-01-15
Fixed RBAC fetching from workflow state when template is not present

owaiskazi19:
### Description
In the current implementation of Role Based Access Control, the user is fetched only from Template. After invoking delete API, the template gets deleted but that should not stop from fetching the state of the same workflow or deprovisioning it. This PR fixes it with reading the user from workflow state index rather global context.

### Related Issues
Resolves https://github.com/opensearch-project/flow-framework/issues/986

### Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 ### Enhancements
 ### Bug Fixes
 - Remove useCase and defaultParams field in WorkflowRequest ([#758](https://github.com/opensearch-project/flow-framework/pull/758))
+- Fix RBAC fetching from workflow state when template is not present ([#998](https://github.com/opensearch-project/flow-framework/pull/998))
 
 ### Infrastructure
 ### Documentation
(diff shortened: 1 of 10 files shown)


dbwiddis (approved):
LGTM!

amitgalitz:
What happens if the user for the template and workflow state are different? If one user creates and another provisions, should we maybe get the user from whichever index corresponds to the action (action on template vs state) index. Or are we deciding that if user has access to the state they by default get access to template and vice versa
Also if we do fine grained updates do we change the user?

joshpalis (commented):
Overall the changes look good to me, though I do share similar concerns with @amitgalitz. Im curious if it would make more sense to retrieve the user from the template when invoking delete and similarly retrieving the user from the workflow state when invoking deprovision/reprovision.

owaiskazi19:
> Also if we do fine grained updates do we change the user?

Discussed this with @amitgalitz offline and missed posting here. The user remains the same in workflow state and template. Since when creating the builder for WorflowState we just use the same user of the template.

joshpalis (approved):

owaiskazi19:
Need to merge https://github.com/opensearch-project/flow-framework/pull/1013 before creating 2.x backport PR

PR #1074, opened 2025-03-07
Add per-tenant provisioning throttling

dbwiddis:
### Description

Adds settings to control per-tenant limits on simultaneous provision, reprovision, and deprovision.

### Check List
- ~[ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).~

### DIFF
--- a/src/main/java/org/opensearch/flowframework/common/FlowFrameworkSettings.java
+++ b/src/main/java/org/opensearch/flowframework/common/FlowFrameworkSettings.java
@@ -33,6 +33,10 @@ public class FlowFrameworkSettings {
     protected volatile TimeValue requestTimeout;
     /** Whether multitenancy is enabled */
     private final Boolean isMultiTenancyEnabled;
+    /** Max simultaneous provision requests */
+    private volatile Integer maxActiveProvisionsPerTenant;
+    /** Max simultaneous deprovision requests */
+    private volatile Integer maxActiveDeprovisionsPerTenant;
 
     /** The upper limit of max workflows that can be created  */
     public static final int MAX_WORKFLOWS_LIMIT = 10000;
@@ -129,6 +133,25 @@ public class FlowFrameworkSettings {
         Setting.Property.NodeScope
     );
 
+    /** This setting sets max workflows that can be simultaneously provisioned, or reprovisioned by the same tenant */
+    public static final Setting<Integer> MAX_ACTIVE_PROVISIONS_PER_TENANT = Setting.intSetting(
+        "plugins.flow_framework.max_active_provisions_per_tenant",
+        2,
+        1,
+        4,
+        Setting.Property.NodeScope,
+        Setting.Property.Dynamic
+    );
+
+    /** This setting sets max workflows that can be simultaneously deprovisioned by the same tenant */
+    public static final Setting<Integer> MAX_ACTIVE_DEPROVISIONS_PER_TENANT = Setting.intSetting(
+        "plugins.flow_framework.max_active_deprovisions_per_tenant",
+        1,
+        1,
+        2,
+        Setting.Property.NodeScope,
+        Setting.Property.Dynamic
+    );
     /** This setting sets the remote metadata type */
     public static final Setting<String> REMOTE_METADATA_TYPE = Setting.simpleString(
         "plugins.flow_framework." + REMOTE_METADATA_TYPE_KEY,
@@ -172,11 +195,17 @@ public class FlowFrameworkSettings {
         this.maxWorkflows = MAX_WORKFLOWS.get(settings);
         this.requestTimeout = WORKFLOW_REQUEST_TIMEOUT.get(settings);
         this.isMultiTenancyEnabled = FLOW_FRAMEWORK_MULTI_TENANCY_ENABLED.get(settings);
+        this.maxActiveProvisionsPerTenant = MAX_ACTIVE_PROVISIONS_PER_TENANT.get(settings);
+        this.maxActiveDeprovisionsPerTenant = MAX_ACTIVE_DEPROVISIONS_PER_TENANT.get(settings);
         clusterService.getClusterSettings().addSettingsUpdateConsumer(FLOW_FRAMEWORK_ENABLED, it -> isFlowFrameworkEnabled = it);
         clusterService.getClusterSettings().addSettingsUpdateConsumer(TASK_REQUEST_RETRY_DURATION, it -> retryDuration = it);
         clusterService.getClusterSettings().addSettingsUpdateConsumer(MAX_WORKFLOW_STEPS, it -> maxWorkflowSteps = it);
         clusterService.getClusterSettings().addSettingsUpdateConsumer(MAX_WORKFLOWS, it -> maxWorkflows = it);
         clusterService.getClusterSettings().addSettingsUpdateConsumer(WORKFLOW_REQUEST_TIMEOUT, it -> requestTimeout = it);
+        clusterService.getClusterSettings()
+            .addSettingsUpdateConsumer(MAX_ACTIVE_PROVISIONS_PER_TENANT, it -> maxActiveProvisionsPerTenant = it);
+        clusterService.getClusterSettings()
+            .addSettingsUpdateConsumer(MAX_ACTIVE_DEPROVISIONS_PER_TENANT, it -> maxActiveDeprovisionsPerTenant = it);
     }
 
     /**
@@ -225,5 +254,21 @@ public class FlowFrameworkSettings {
      */
     public boolean isMultiTenancyEnabled() {
         return isMultiTenancyEnabled;
+    }
+
+    /**
+     * Getter for max active provisions per tenant
+     * @return max active provisions
+     */
+    public Integer getMaxActiveProvisionsPerTenant() {
+        return maxActiveProvisionsPerTenant;
+    }
+
+    /**
+     * Getter for max active deprovisions per tenant
+     * @return max active deprovisions
+     */
+    public Integer getMaxActiveDeprovisionsPerTenant() {
+        return maxActiveDeprovisionsPerTenant;
     }
 }
--- a/src/main/java/org/opensearch/flowframework/transport/ReprovisionWorkflowTransportAction.java
+++ b/src/main/java/org/opensearch/flowframework/transport/ReprovisionWorkflowTransportAction.java
@@ -429,9 +429,11 @@ public class ReprovisionWorkflowTransportAction extends HandledTransportAction<R
             }
 
             logger.info("Reprovisioning completed successfully for workflow {}", workflowId);
+            // Need to call TenantAwareHelper.releaseProvision in cases listener.onFailure is not called
+            String tenantId = template.getTenantId();
             flowFrameworkIndicesHandler.updateFlowFrameworkSystemIndexDoc(
                 workflowId,
-                template.getTenantId(),
+                tenantId,
                 Map.ofEntries(
                     Map.entry(STATE_FIELD, State.COMPLETED),
                     Map.entry(PROVISIONING_PROGRESS_FIELD, ProvisioningProgress.DONE),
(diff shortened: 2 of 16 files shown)


dbwiddis:
tests failing because of jar hell caused by https://github.com/opensearch-project/ml-commons/pull/3475

src/main/java/org/opensearch/flowframework/util/TenantAwareHelper.java
```diff
@@ -88,4 +94,156 @@ public static String getTenantID(Boolean isMultiTenancyEnabled, RestRequest rest
...
+     * @param executionsMap The map tracking the number of active executions per tenant.
+     */
+    private static void release(String tenantId, ConcurrentHashMap<String, AtomicInteger> executionsMap) {
+        if (tenantId == null) {
+            return; // No throttling for null tenantId
+        }
+        AtomicInteger count = executionsMap.get(tenantId);
+        if (count != null) {
+            count.decrementAndGet();
+        }
```
owaiskazi19, line 192: Don't we have to remove tenantId from the map here?
```suggestion
        if (count != null && count.decrementAndGet() <= 0) {
        executionsMap.remove(tenantId);
    }
```

src/main/java/org/opensearch/flowframework/common/FlowFrameworkSettings.java
```diff
@@ -129,6 +133,25 @@ public class FlowFrameworkSettings {
         Setting.Property.NodeScope
     );
 
+    /** This setting sets max workflows that can be simultaneously provisioned, or reprovisioned by the same tenant */
+    public static final Setting<Integer> MAX_ACTIVE_PROVISIONS_PER_TENANT = Setting.intSetting(
+        "plugins.flow_framework.max_active_provisions_per_tenant",
+        2,
```
owaiskazi19, line 139: Why are we keeping the default value as 2 here?

src/main/java/org/opensearch/flowframework/transport/ReprovisionWorkflowTransportAction.java
```diff
@@ -452,8 +460,15 @@ private void executeWorkflow(
...
                             })
                         );
+                    } else {
+                        // We've completed provisioning asynchronously
+                        TenantAwareHelper.releaseProvision(tenantId);
                     }
-                }, exception -> { logger.error("Failed to update workflow state for workflow {}", workflowId, exception); })
+                }, exception -> {
+                    // We've completed provisioning asynchronously but failed state update
+                    TenantAwareHelper.releaseProvision(tenantId);
```
owaiskazi19, line 469: In the concurrent map should we also store the status of provisioning along with the tenantId? To keep a track for which tenantId we had a successful provisioning

src/main/java/org/opensearch/flowframework/util/TenantAwareHelper.java
```diff
@@ -88,4 +94,156 @@ public static String getTenantID(Boolean isMultiTenancyEnabled, RestRequest rest
...
+
+            @Override
+            public void onFailure(Exception e) {
+                try {
+                    releaseDeprovision(tenantId);
+                } finally {
+                    delegate.onFailure(e);
+                }
+            }
+        });
```
owaiskazi19, line 247: Can use runAfter to reduce the method completely
```java
return ActionListener.notifyOnce(ActionListener.runAfter(delegate, () -> releaseDeprovision(tenantId)));
```

src/main/java/org/opensearch/flowframework/util/TenantAwareHelper.java
```diff
@@ -88,4 +94,156 @@ public static String getTenantID(Boolean isMultiTenancyEnabled, RestRequest rest
...
+
+            @Override
+            public void onFailure(Exception e) {
+                try {
+                    releaseProvision(tenantId);
+                } finally {
+                    delegate.onFailure(e);
+                }
+            }
+        });
```
owaiskazi19, line 219: Same as below

src/test/java/org/opensearch/flowframework/rest/RestWorkflowProvisionTenantAwareIT.java
```diff
@@ -0,0 +1,376 @@
...
+        response = makeRequest(tenantRequest, DELETE, WORKFLOW_PATH + workflowId2);
+        assertOK(response);
+        map = responseToMap(response);
+        assertEquals(workflowId2, map.get(DOC_ID).toString());
+
+        // Delete from other tenant
+        response = makeRequest(otherTenantRequest, DELETE, WORKFLOW_PATH + otherWorkflowId);
+        assertOK(response);
+        map = responseToMap(response);
+        assertEquals(otherWorkflowId, map.get(DOC_ID).toString());
```
owaiskazi19, line 341: We are doing a bunch of operations in this test case. We could split the tests with `whenThrottlingEnabled`, `whenMultitenantIsEnabled`, `whenThrottlingisDisabled` and  `whenMultitenantIsDisabled`

owaiskazi19 (commented):
Looks good overall with comments

src/main/java/org/opensearch/flowframework/util/TenantAwareHelper.java
```diff
@@ -88,4 +94,156 @@ public static String getTenantID(Boolean isMultiTenancyEnabled, RestRequest rest
...
+        if (tenantId == null) {
+            return true; // No throttling for null tenantId
+        }
+        AtomicInteger count = executionsMap.computeIfAbsent(tenantId, k -> new AtomicInteger(0));
+        if (count.incrementAndGet() <= maxExecutions) {
+            return true;
+        } else {
+            count.decrementAndGet(); // Rollback the increment
+            return false;
+        }
```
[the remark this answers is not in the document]
dbwiddis: > We could use `compareAndSet` to avoid any race condition due to increment and decrement

I considered almost exactly your suggestion but there's still a race condition between your `.get()` on line 171 and the comparison on line 173.  In which case `compareAndSet()` might fail even if the change in `count` was a decrement or an increment below maxExecutions.

src/main/java/org/opensearch/flowframework/common/FlowFrameworkSettings.java
```diff
@@ -129,6 +133,25 @@ public class FlowFrameworkSettings {
...
+        1,
+        4,
+        Setting.Property.NodeScope,
+        Setting.Property.Dynamic
+    );
+
+    /** This setting sets max workflows that can be simultaneously deprovisioned by the same tenant */
+    public static final Setting<Integer> MAX_ACTIVE_DEPROVISIONS_PER_TENANT = Setting.intSetting(
+        "plugins.flow_framework.max_active_deprovisions_per_tenant",
+        1,
```
amitgalitz, line 149: I think we should keep it the same as provision and at 2 at least

src/main/java/org/opensearch/flowframework/util/TenantAwareHelper.java
dbwiddis: The code logic would work the same here.  But I can see how filling up a static map with a bunch of 0's could leave a bunch of zero map entries.

src/main/java/org/opensearch/flowframework/common/FlowFrameworkSettings.java
```diff
@@ -129,6 +133,25 @@ public class FlowFrameworkSettings {
         Setting.Property.NodeScope
     );
 
+    /** This setting sets max workflows that can be simultaneously provisioned, or reprovisioned by the same tenant */
+    public static final Setting<Integer> MAX_ACTIVE_PROVISIONS_PER_TENANT = Setting.intSetting(
+        "plugins.flow_framework.max_active_provisions_per_tenant",
+        2,
+        1,
+        4,
```
amitgalitz, line 141: Should we also make our threadpool size dynamic and the max higher here incase we have larger nodes that can accommodate that?

src/main/java/org/opensearch/flowframework/common/FlowFrameworkSettings.java
dbwiddis: I want the default value to be less than our thread pool size. 

What other default value do you suggest?

src/main/java/org/opensearch/flowframework/util/TenantAwareHelper.java
dbwiddis: Your code suggestion still leaves a race condition between the decrement and get, and the removal.

I think we could use computeIfPresent, though, and it removes the entry on null.

src/main/java/org/opensearch/flowframework/transport/ReprovisionWorkflowTransportAction.java
dbwiddis: Not sure what you're asking. This is intended to be a very lightweight throttling mechanism, not a stats collector.

src/test/java/org/opensearch/flowframework/rest/RestWorkflowProvisionTenantAwareIT.java
dbwiddis: I used the same pattern as my other ITs, which rely on a long sequence of operations, most of which are the same and only some of which are different.   Splitting them up would increase duplication.

src/main/java/org/opensearch/flowframework/common/FlowFrameworkSettings.java
dbwiddis: Yes.

src/main/java/org/opensearch/flowframework/util/TenantAwareHelper.java
amitgalitz: +1 on computeIfPresent so if count is 0 then we should remove entry from map

amitgalitz: Additionally could a race condition here cause a negative count?

dbwiddis: > could a race condition here cause a negative count?

I don't think so?  The code to decrement is only run after there was an increment.

src/main/java/org/opensearch/flowframework/util/TenantAwareHelper.java
dbwiddis: In theory, yes, we could use `runBefore()` or `runAfter()` here because we are calling `releaseDeprovision()` in both cases.  In fact, the code here is exactly what happens on `runBefore()`.  

However, this is just a lucky break because we use synchronous code for deprovisioning, so our action listener handling is simpler.  We can _not_ do it for the provision and reprovision code because we only call `releaseProvision()` on failure and _not_ on success.

In which case I'd prefer to keep this method implementation similar to the adjacent method in the same class.

src/main/java/org/opensearch/flowframework/util/TenantAwareHelper.java
dbwiddis: We only release on failure here so we can't wrap it with the same method in both cases.

dbwiddis:
@owaiskazi19 @amitgalitz Thanks for your review!  

I've addressed some comments in https://github.com/opensearch-project/flow-framework/pull/1074/commits/6fbf71842500eb02680e92dc4e2ba5b2bca756f6:
 - used `computeIfPresent()` to remove from the map when it hits 0
 - used `ActionListener.runBefore()` in the deprovision case; it is a lot simpler

I'd like to handle changing the value of settings/limits and defaults in another PR (followed by a discussion issue).

I'll consider refactoring all of the multitenant ITs at a later date.

Other comments I've responded that I don't think changes are needed; I've added some comments in code explaining why.  In particular I don't think there's a easy way to replace the increment/rollback logic with compareAndSet without really overcomplicating things.

owaiskazi19 (approved):
Thanks for addressing the comments

amitgalitz (approved):

Issue #1072, opened 2025-02-27
[BUG] Jar hell from org.bouncycastle:bcprov-jdk18on


dbwiddis:
### What is the bug?

Build is failing on Jar Hell.
```
FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':dependencies'.
> Could not resolve all dependencies for configuration ':testCompileClasspath'.
   > Conflict found for the following module:
       - org.bouncycastle:bcprov-jdk18on between versions 1.80 and 1.78
```

### How can one reproduce the bug?

 -  `./gradlew compileTestJava`

### What is the expected behavior?

No Jar Hell

### Do you have any additional context?

Our build.gradle specifies version 1.8.0:
https://github.com/opensearch-project/flow-framework/blob/88fe98e561a68337fbe1188e0189a87a01fa4f50/build.gradle#L180

The OpenSearch Version Catalog specifies 1.7.8:
https://github.com/opensearch-project/OpenSearch/blob/0ffed5e8b743a075f5d66c4fe6e9b8371eacfa14/gradle/libs.versions.toml#L61

Quick fix:
 - Change the `build.gradle` line above to get the version from the version catalog (`${versions.bouncycastle}`)

Better fix:
 - Audit all the dependencies to see if they are in the version catalog and update appropriately.
 - Check whether they are actually required on the main build or whether they can be `testImplementation` as this one appears to probably be

PR #1087, opened 2025-03-21
Migrate from BC to BCFIPS libraries

cwperks:
### Description

`bouncycastle` was removed from the gradle version catalog in https://github.com/opensearch-project/OpenSearch/pull/17507/files#diff-697f70cdd88ba88fe77eebda60c7e143f6ad1286bca75017421e93ad84fb87df.

This PR migrates from BC to BCFIPS libraries for this repo

### Check List
- [ ] Commits are signed per the DCO using `--signoff`.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -177,6 +177,8 @@ dependencies {
     implementation "software.amazon.cryptography:aws-cryptographic-material-providers:1.9.0"
     implementation "org.dafny:DafnyRuntime:4.10.0"
     implementation "software.amazon.smithy.dafny:conversion:0.1.1"
+    // needed by aws-encryption-sdk-java
+    implementation "org.bouncycastle:bc-fips:${versions.bouncycastle_jce}"
     api "org.apache.httpcomponents.core5:httpcore5:${versions.httpcore5}"
     implementation "jakarta.json.bind:jakarta.json.bind-api:3.0.1"
     implementation "org.glassfish:jakarta.json:2.0.1"
--- a/build.gradle
+++ b/build.gradle
@@ -177,7 +177,7 @@ dependencies {
     implementation "software.amazon.cryptography:aws-cryptographic-material-providers:1.9.0"
     implementation "org.dafny:DafnyRuntime:4.10.0"
     implementation "software.amazon.smithy.dafny:conversion:0.1.1"
-    implementation "org.bouncycastle:bcprov-jdk18on:1.80"
+    implementation "org.bouncycastle:bcprov-jdk18on:1.78"
     api "org.apache.httpcomponents.core5:httpcore5:${versions.httpcore5}"
     implementation "jakarta.json.bind:jakarta.json.bind-api:3.0.1"
     implementation "org.glassfish:jakarta.json:2.0.1"
--- a/build.gradle
+++ b/build.gradle
@@ -8,8 +8,8 @@ import java.nio.file.Paths
 
 buildscript {
     ext {
-        opensearch_version = System.getProperty("opensearch.version", "3.0.0-alpha1-SNAPSHOT")
-        buildVersionQualifier = System.getProperty("build.version_qualifier", "alpha1")
+        opensearch_version = System.getProperty("opensearch.version", "3.0.0-beta1-SNAPSHOT")
+        buildVersionQualifier = System.getProperty("build.version_qualifier", "beta1")
         isSnapshot = "true" == System.getProperty("build.snapshot", "true")
         version_tokens = opensearch_version.tokenize('-')
         opensearch_build = version_tokens[0] + '.0'
@@ -178,7 +178,7 @@ dependencies {
     implementation "org.dafny:DafnyRuntime:4.10.0"
     implementation "software.amazon.smithy.dafny:conversion:0.1.1"
     // needed by aws-encryption-sdk-java
-    implementation "org.bouncycastle:bc-fips:2.0.0"
+    implementation "org.bouncycastle:bc-fips:${versions.bouncycastle_jce}"
     api "org.apache.httpcomponents.core5:httpcore5:${versions.httpcore5}"
     implementation "jakarta.json.bind:jakarta.json.bind-api:3.0.1"
     implementation "org.glassfish:jakarta.json:2.0.1"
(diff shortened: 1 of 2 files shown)


dbwiddis:
Hey @cwperks we previously had 1.8.0 which caused jar hell.  We downgraded to the version catalog to avoid that.  

But I expect our dependency manager will suggest an upgrade again. 

Can you be more clear on the reason for this downgrade if it is not for consistency with other dependencies?  Can we not make this 1.8.0 again by reverting #1078?

cwperks:
@dbwiddis I can try to update the version, but the purpose of this PR is to put the version inline instead of referencing the version catalog because `versions.bouncycastle` was removed from the version catalog.

@beanuwave has been working on a large effort around [FIPS-140-3](https://github.com/opensearch-project/security/issues/3420) compliance for OpenSearch which includes swapping BC non-FIPS jars with FIPS jars so non-FIPS references were removed from the core.

Currently the jar swap is only for the core and has not been performed in the plugins. Plugins that referenced the version of bouncycastle from core's version catalog now need to put the version inline.

cwperks:
@dbwiddis I think the update will work, but core also changed the qualifier to beta1 and we now need to wait for all of FF's dependent plugins to update accordingly.

Switching back to 1.78 in the interim

cwperks:
> @dbwiddis I think the update will work, but core also changed the qualifier to beta1 and we now need to wait for all of FF's dependent plugins to update accordingly.

Confirmed it would work.

When checking out core you need to specify the qualifier when building. i.e. `./gradlew publishToMavenLocal -Dbuild.version_qualifier=alpha1`

dbwiddis:
The bigger question is why we are not aligning on a common version. I’d like to prevent future churn.

cwperks:
One option could be to do the swap in this repo from non-FIPS -> FIPS deps for BC dependencies.

CC @beanuwave

dbwiddis:
> One option could be to do the swap in this repo from non-FIPS -> FIPS deps for BC dependencies.
> 
> CC @beanuwave

Sure, happy to look into this.  Still trying to understand what happened here.
 - we'd been bumping versions normally and had gone from 1.78 to 1.80 over the last few months
 - our pipeline suddenly broke with 1.80 due to jar hell with 1.78 (#1072)
 - we updated to the version catalog to hopefully eliminate that
 - if we merge this PR we'll be back where we were a few months ago and we'll get a version bump to 1.80
 - rather than merging this PR I'd prefer to just revert #1078 if it's no longer needed
 - but I don't understand yet why it was a problem to begin with. 
   - Was whatever was causing the jar hell removed and 1.80 will work fine? (In which case let's revert #1078)
   - Or is there some problem with this dependency where we should switch to something else?

cwperks:
> Was whatever was causing the jar hell removed and 1.80 will work fine?

It used to collide with the dependencies in opensearch-test-framework [here](https://github.com/opensearch-project/OpenSearch/pull/17507/files#diff-63e45f04f9e0184f1f10ba17bbeb51e13abe858369a206acfa01fd8dd4be58abL52-L54). With this PR those deps changed to BCFIPS. 

```
> Task :dependencyInsight
Dependency resolution failed because of conflict on the following module:
   - org.bouncycastle:bcprov-jdk18on between versions 1.80 and 1.78

org.bouncycastle:bcprov-jdk18on:1.80
  Variant compile:
    | Attribute Name                 | Provided | Requested    |
    |--------------------------------|----------|--------------|
    | org.gradle.status              | release  |              |
    | org.gradle.category            | library  | library      |
    | org.gradle.libraryelements     | jar      | classes      |
    | org.gradle.usage               | java-api | java-api     |
    | org.gradle.dependency.bundling |          | external     |
    | org.gradle.jvm.environment     |          | standard-jvm |
    | org.gradle.jvm.version         |          | 21           |
   Selection reasons:
      - By conflict resolution: between versions 1.80 and 1.78

org.bouncycastle:bcprov-jdk18on:1.80
\--- testCompileClasspath

org.bouncycastle:bcprov-jdk18on:1.78 -> 1.80
\--- org.opensearch.test:framework:3.0.0-alpha1-SNAPSHOT:20250319.180915-157
     \--- testCompileClasspath
```

dbwiddis:
> ncies in opensearch-test-framework [here](https://github.com/opensearch-project/OpenSearch/pull/17507/files#diff-63e45f04f9e0184f1f10ba17bbeb51e13abe858369a206acfa01fd8dd4be58abL52-L54). With this [PR](https://github.com/opensearch-project/OpenSearch/pull/17507) those deps changed to BCFIPS.

So I think the best action for us is to do the same

build.gradle
```diff
@@ -177,7 +177,8 @@ dependencies {
     implementation "software.amazon.cryptography:aws-cryptographic-material-providers:1.9.0"
     implementation "org.dafny:DafnyRuntime:4.10.0"
     implementation "software.amazon.smithy.dafny:conversion:0.1.1"
-    implementation "org.bouncycastle:bcprov-jdk18on:${versions.bouncycastle}"
+    // needed by aws-encryption-sdk-java
+    implementation "org.bouncycastle:bc-fips:2.0.0"
```
cwperks, line 181: Switch this to `implementation "org.bouncycastle:bc-fips:${versions.bouncycastle_jce}"` when FF updates to beta1 qualifier.

This is added in the [version catalog](https://github.com/opensearch-project/OpenSearch/blob/main/gradle/libs.versions.toml#L63) after the cut to beta1 and is not available in the alpha1 artifacts.

dbwiddis:
@cwperks Thanks for updating the PR!  We'll merge this and switch to beta1 once all our upstream repos do...

One curious question remains:

> It used to collide with the dependencies in opensearch-test-framework [here](https://github.com/opensearch-project/OpenSearch/pull/17507/files#diff-63e45f04f9e0184f1f10ba17bbeb51e13abe858369a206acfa01fd8dd4be58abL52-L54).

Yeah, but we had bumped to 1.8.0 on Jan 15, and from 1.78 to 1.79 [back on Nov 1 last year](https://github.com/opensearch-project/flow-framework/pull/940).  But the Jar hell only started relatively recently.  Maybe it was whatever security fix we needed that caused the 2.19.1 release since it happened around that time...

dbwiddis (approved):

amitgalitz (approved):

PR #1096, opened 2025-04-04
Fix Config parser does not handle tenant_id field

### LINKED ISSUE #1095: [BUG] Config parser does not handle tenant_id field
### What is the bug?

When using a remote DDB client with tenant awareness set, calling a REST request which caches a tenant_id key that already exists times out.
 - The `Config` class is hardcoded for only two fields, `master_key` and `create_time`.
 - The mapping declaration for the config index matches and is set `"dynamic": false`.
 - With tenant awareness set, the SdkClient automatically adds a `tenant_id` field to each object.
 - On NodeClient, the non-dynamic setting of the index means this extra field is ignored.  This may also be the case for the remote client depending on the mapping assigned.
 - On remote clients such as DDB the tenant ID is saved as part of the object (essentially allowing a dynamic setting) and thus is read in as the `source` from a get call when the item exists.

In addition to parsing the `Config` causing an exception, the exception is not caught nor logged, making it difficult to track.

1. Create a template which uses remote storage (such as Dynamo DB), with tenant awareness enabled.
2. Restart the local OpenSearch instance to clear the map cache.
3. Create another template with the same remote storage, using the same tenant ID as in step 1.

REST calls complete as expected, and when they don't, they provide useful debugging information.

The `tenant_id` field isn't actually used so there is no need to add it to the mappings; however, parsing the return JSON value should be robust to the presence of a tenant ID.

Possible fixes:
 - Don't parse returned JSON into `Config`, instead use the parse-to-map feature and fetch the key from the map.  (This is what ML Commons does.)
 - When parsing, add a case to the switch statement read the tenant_id field and call `parser.text()` but ignore it (probably the fastest, I'd recommend)
 - Add the field to the `Config` object and change the mapping to include it as well (not recommended, we don't actually need this field as the Config index is not user-facing and the tenant ID is already encoded in the document id).

In addition to fixing the parsing error, update the try-with-resoruces block where `Config.parse()` is called to catch exceptions and return a 500 error.

CC: @junweid62 @dhrubo-os @arjunkumargiri

junweid62:
Used this recommend option to fix this edge case.

### Related Issues
Resolves #1095 

### Check List

### DIFF
--- a/src/main/java/org/opensearch/flowframework/model/Config.java
+++ b/src/main/java/org/opensearch/flowframework/model/Config.java
@@ -8,6 +8,8 @@
  */
 package org.opensearch.flowframework.model;
 
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 import org.opensearch.core.rest.RestStatus;
 import org.opensearch.core.xcontent.ToXContentObject;
 import org.opensearch.core.xcontent.XContentBuilder;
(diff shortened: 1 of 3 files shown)


src/main/java/org/opensearch/flowframework/model/Config.java
```diff
@@ -59,12 +78,17 @@ public XContentBuilder toXContent(XContentBuilder builder, Params params) throws
...
+        String tenantId = null;
 
         ensureExpectedToken(XContentParser.Token.START_OBJECT, parser.currentToken(), parser);
         while (parser.nextToken() != XContentParser.Token.END_OBJECT) {
             String fieldName = parser.currentName();
             parser.nextToken();
+            logger.info("Parsing field: {}, token: {}, value: {}", fieldName, parser.currentToken(), parser.text());
             switch (fieldName) {
+                case TENANT_ID_FIELD:
+                    tenantId = parser.text();
```
dbwiddis, line 90: This could be `null`.  Use `parser.objectText()` here which will either be `null` or a `String` you can either cast or `toString()` the object to.

src/main/java/org/opensearch/flowframework/model/Config.java
```diff
@@ -21,29 +23,46 @@
...
-     * @param masterKey The encryption master key
+     * @param masterKey  The encryption master key
      * @param createTime The config creation time
+     * @param tenantId   The tenantId
      */
-    public Config(String masterKey, Instant createTime) {
+    public Config(String masterKey, Instant createTime, String tenantId) {
         this.masterKey = masterKey;
         this.createTime = createTime;
+        this.tenantId = tenantId;
```
dbwiddis, line 49: Given all the complexities of potentially null tenant ID and the fact that we don't really ever use the value, can we just simplify here and only handle the case where the field exists, then read the value (`.objectTex()`) and ignore it?

dbwiddis: Also https://github.com/opensearch-project/flow-framework/blob/main/src/test/java/org/opensearch/flowframework/model/ConfigTests.java needs to be updated to test for present, null, and absent tenant Id

PR #1098, opened 2025-04-07
Complete action listener on failed synchronous workflow provisioning

### LINKED ISSUE #1097: [BUG] Action Listener is not completed on failed synchronous provisioning
### What is the bug?

When provisioning fails in a workflow step, the `actionGet()` call here throws an `ExecutionException`:
https://github.com/opensearch-project/flow-framework/blob/ceffb7bdc262f89f1dd5983736052691c5bc8d6b/src/main/java/org/opensearch/flowframework/transport/ProvisionWorkflowTransportAction.java#L402-L406

However, in the `catch` block, the action listener is never completed. 

1. Enable tenant awareness
2. Provision synchronously multiple times with failed provisioning
3. Quickly reach the limit of in-process provisioning from tenant-aware throttling

Action listener is completed in all cases.

This was not originally required when provisioning was asynchronous, because we would always complete the listener regardless of whether the provisioning failed; we would simply monitor the state.

However, with synchronous provisioning we need to track failures at the time the provisioning fails.

Probably best to complete the listener (`onFailure()`) in both cases of this wrapper:
https://github.com/opensearch-project/flow-framework/blob/ceffb7bdc262f89f1dd5983736052691c5bc8d6b/src/main/java/org/opensearch/flowframework/transport/ProvisionWorkflowTransportAction.java#L470-L472

We should also add a test for exception during synchronous provisioning to `ProvisionWorkflowTransportActionTests`.

dbwiddis:
### Description

During synchronous provisioning (and reprovisioning) the actionListener was never completed on an exception in the workflow steps (which throw an exception on `actionGet()`).  
 - For asynchronous provisioning this was expected, as the listener was already completed.  In the sync case we ended up just failing on the timeout in this case.
 - Also had to handle the case of synchronous provisioning timeout

### Related Issues

Resolves #1097

### Check List

### DIFF
--- a/src/main/java/org/opensearch/flowframework/util/WorkflowTimeoutUtility.java
+++ b/src/main/java/org/opensearch/flowframework/util/WorkflowTimeoutUtility.java
@@ -14,6 +14,7 @@ import org.opensearch.ExceptionsHelper;
 import org.opensearch.common.unit.TimeValue;
 import org.opensearch.core.action.ActionListener;
 import org.opensearch.flowframework.exception.FlowFrameworkException;
+import org.opensearch.flowframework.model.WorkflowState;
 import org.opensearch.flowframework.transport.GetWorkflowStateAction;
 import org.opensearch.flowframework.transport.GetWorkflowStateRequest;
 import org.opensearch.flowframework.transport.WorkflowResponse;
@@ -164,25 +165,13 @@ public class WorkflowTimeoutUtility {
      *
      * @param workflowId    The unique identifier of the workflow.
      * @param e             The exception that occurred during workflow execution.
-     * @param isResponseSent An atomic boolean to ensure the response is sent only once.
      * @param listener       The listener to notify of the workflow failure.
      */
-    public static void handleFailure(
-        String workflowId,
-        Exception e,
-        AtomicBoolean isResponseSent,
-        ActionListener<WorkflowResponse> listener
-    ) {
-        // Check if the failure has already been reported, and report it only if it hasn't been reported yet.
-        if (isResponseSent.compareAndSet(false, true)) {
-            FlowFrameworkException exception = new FlowFrameworkException(
-                "Failed to execute workflow " + workflowId,
-                ExceptionsHelper.status(e)
-            );
-            listener.onFailure(exception);
-        } else {
-            logger.info("Ignoring onFailure for workflowId: {} as timeout already occurred", workflowId);
-        }
+    public static void handleFailure(String workflowId, Exception e, ActionListener<WorkflowResponse> listener) {
+        FlowFrameworkException exception = e instanceof FlowFrameworkException
+            ? (FlowFrameworkException) e
+            : new FlowFrameworkException("Failed to execute workflow " + workflowId, ExceptionsHelper.status(e));
+        listener.onFailure(exception);
     }
 
     /**
@@ -207,8 +196,16 @@ public class WorkflowTimeoutUtility {
             new GetWorkflowStateRequest(workflowId, false, tenantId),
             ActionListener.wrap(
                 response -> listener.onResponse(new WorkflowResponse(workflowId, response.getWorkflowState())),
-                exception -> listener.onFailure(
-                    new FlowFrameworkException("Failed to get workflow state after timeout", ExceptionsHelper.status(exception))
+                // we don't want to fail the listener as provisioning is still ongoing
+                exception -> listener.onResponse(
+                    new WorkflowResponse(
+                        workflowId,
+                        WorkflowState.builder()
+                            .workflowId(workflowId)
+                            .error("Workflow timed out, failed to fetch current state")
+                            .state("UNKNOWN")
+                            .build()
+                    )
                 )
             )
         );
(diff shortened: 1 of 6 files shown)


joshpalis (approved):

src/main/java/org/opensearch/flowframework/util/WorkflowTimeoutUtility.java
```diff
@@ -205,8 +208,10 @@ public static void fetchWorkflowStateAfterTimeout(
         client.execute(
             GetWorkflowStateAction.INSTANCE,
             new GetWorkflowStateRequest(workflowId, false, tenantId),
-            ActionListener.wrap(
-                response -> listener.onResponse(new WorkflowResponse(workflowId, response.getWorkflowState())),
+            ActionListener.wrap(response -> {
+                TenantAwareHelper.releaseProvision(tenantId);
```
amitgalitz, line 212: Are we sure we want to release provisioning, incase provisioning is still ongoing?

junweid62:
LGTM overall!

One question similar to @amitgalitz  – are we sure we want to allow releasing while provisioning might still be ongoing? Since our timeout logic doesn't cancel provisioning, it could still be running in the background. Would that cause issues if we release early?

src/main/java/org/opensearch/flowframework/util/WorkflowTimeoutUtility.java
dbwiddis: Good callout. No. I just reverted the last commit, will post a comment below outlining entire code path.

dbwiddis:
Tenant-aware listener:  releases on failure, we need to release (exactly once) on success.   
Possible code paths:
- We wrap the original listener in this, so the `listener` throughout the code is this one.
  - If we fail on lack of permission, listener is failed. Done. Otherwise getWorkflow().
  - If we fail to get the workflow, listener is failed. Done. Otherwise onGetWorkflowResponse().
  - If we fail parsing etc. listener is failed. Done. Otherwise run the function executeProvisionRequest().
  - Get the template, update state, etc. If we fail in any, listener is failed. Done. Otherwise:
    - If wait for completion timeout is MINUS_ONE (line 235) async provisioning.  Also (line 252) Complete the listener successfully and return.  (This does not release the provisioning lock which must be released on successful or failed provisioning, one of the fixes in this PR.)
    - Else sync provisioning

- Asynchronous provisioning (existing behavior).  Occurs when wait_for_completion_timeout is -1.
   - call executeWorkflow with isSyncExecution false
- Synchronous provisioning. TImeout >= 0.
   - call executeWorkflow with a new action listener. 
     - On success calls timeout utility response which completes the original listener
     - on failure calls timeout utility failure which completes the original listener

- executeWorkflow
   - Provisions asynchronously (but tracks return-to-user with isSyncExecution value
     - If successful completion of the workflow (all workflow steps complete) we update the state
       - if isSyncExecution we fetch the workflow state to return
         - we either release provision lock and complete onResponse on update, which goes to timeout utility handleResponse
         - or fail on update which goes to handleFailure which completes original listener and releases the lock
       - if not isSyncExecution we just release the lock (we have already completed action listener)
     - If an exception occurs in a workflow, actionGet() throws exception at line 405 entering the catch block at line 448
       - we update the failed state.
         - on success state update
           - with isSyncExecution we fail which goes to timeout utility handle failure and completes the original listener (this is the main bug fix of this PR)
           - otherwise we release the lock (original listener was already completed)
         - on failed state update
           - with isSyncExecution we fail which goes to timeout utility handle failure and completes the original listener (this is the main bug fix of this PR)
           - otherwise we release the lock (original listener was already completed)

- Other possible outcome: we are sync provisioning but timeout expires.  This is similar to async provisioning.
  - we will complete the original action listener on success but continue "async" provisioning in the background under the isSyncExecution code paths above.  WE do not want to release the provision lock.
  - on failed return value we still do not want to release the provision lock but we DO want to fail the original action listener.
  - Code path here starts with scheduleTimeoutHandler() on line 363
  - Eventually goes to fetchWorkflowState after timeout
  - We do NOT want to fail the original listener here because this would release the provision lock
    - so needed change, exception path on workflow timeout utility line 210 should NOT complete the "tenant aware lock" listener but rather complete on response but with a failed state.

junweid62 (approved):

dbwiddis:
> One question similar to @amitgalitz – are we sure we want to allow releasing while provisioning might still be ongoing? Since our timeout logic doesn't cancel provisioning, it could still be running in the background. Would that cause issues if we release early?

Yes, both @junweid62 and @amitgalitz called out correctly that for timeout we want to continue with the provisioning as we normally would for the "async" case.  I've gone through all the code paths and updated the PR.
 - all of the releasing is handled either by the onFailure listener (which can only release once) or manually in the (Re)Provision transport actions on the success cases.  No releasing occurs in the timeout utility.
 - the timeout utility keeps track of whether the listener onResponse was sent, so in the success case it only sends it once, but the provision release still always occurs even if the provision success onResponse occurs after timeout; it releases the lock and then the onResponse ends up being a no-op.

amitgalitz (approved):

PR #1139, opened 2025-05-09
Make thread pool sizes configurable

dbwiddis:
### Description

Allows configuration of thread pool sizes at startup using a static setting.

### Check List
- [X] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).
  - https://github.com/opensearch-project/documentation-website/pull/9881

Reviewers please provide tech review of above doc PR.

### DIFF
--- a/src/main/java/org/opensearch/flowframework/common/FlowFrameworkSettings.java
+++ b/src/main/java/org/opensearch/flowframework/common/FlowFrameworkSettings.java
@@ -33,8 +33,14 @@ public class FlowFrameworkSettings {
     protected volatile TimeValue requestTimeout;
     /** Whether multitenancy is enabled */
     private final Boolean isMultiTenancyEnabled;
+    /** Size of the threadpool used for retryable tasks in workflows */
+    private volatile Integer workflowThreadPoolSize;
+    /** Size of the threadpool for provisioning */
+    private volatile Integer provisionThreadPoolSize;
     /** Max simultaneous provision requests */
     private volatile Integer maxActiveProvisionsPerTenant;
+    /** Size of the threadpool for deprovisioning */
+    private volatile Integer deprovisionThreadPoolSize;
     /** Max simultaneous deprovision requests */
     private volatile Integer maxActiveDeprovisionsPerTenant;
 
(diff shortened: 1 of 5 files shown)


junweid62:
LGTM! Just curious — why do we need this feature?

dbwiddis:
> LGTM! Just curious — why do we need this feature?

The existing sizes are assuming a single tenant cluster.  For Multi-tenant implementation with many more tenants, we need to be able to have more simultaneous tenant provisioning.

I honestly think our thread pool sizes are too small as we're mostly just blocking waiting on the result of async REST calls.

dbwiddis:
Tests failing due to https://github.com/opensearch-project/job-scheduler/pull/773, need upstream dependencies who depend on Guava to update.

junweid62 (approved):

src/main/java/org/opensearch/flowframework/common/FlowFrameworkSettings.java
```diff
@@ -133,22 +139,49 @@ public class FlowFrameworkSettings {
         Setting.Property.NodeScope
     );
 
+    /** This setting sets max retryable tasks during workflow execution that polls results */
+    public static final Setting<Integer> WORKFLOW_THREAD_POOL_SIZE = Setting.intSetting(
```
owaiskazi19, line 143: Do we want to keep this setting as dynamic or final?

dbwiddis: It should not be dynamic, as it is used in Thread Pool initialization and it is not possible to resize thread pools once started.

It is currently static (there is no `Dynamic` property), where it can be changed on a node restart.

I do not think it needs to be final (where it would survive a node restart).

owaiskazi19 (approved):

amitgalitz (approved):

src/main/java/org/opensearch/flowframework/common/FlowFrameworkSettings.java
dbwiddis: > We are not making thread pool sizes configurable then :)

We are making them configurable in `opensearch.yml` or command line parameters, which is the intent. :)

PR #1141, opened 2025-05-12
Conditionally include ddb-client dependency only if env variable set 

dbwiddis:
### Description

Adds a configuration to include the remote metadata ddb-client dependency if an environment variable is set.

Allows build-time inclusion of the dependency in locations where it is required, and doesn't include extra dependencies in other cases.

### Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@ All notable changes to this project are documented in this file.
 
 Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 
-## [Unreleased 3.1](https://github.com/opensearch-project/flow-framework/compare/3.0...HEAD)
+## [Unreleased 3.x](https://github.com/opensearch-project/flow-framework/compare/3.0...HEAD)
 ### Features
 ### Enhancements
 - Make thread pool sizes configurable ([#1139](https://github.com/opensearch-project/flow-framework/issues/1139))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,13 +3,15 @@ All notable changes to this project are documented in this file.
 
 Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 
-## [Unreleased 3.x](https://github.com/opensearch-project/flow-framework/compare/3.0...HEAD)
+## [Unreleased 3.1](https://github.com/opensearch-project/flow-framework/compare/3.0...HEAD)
 ### Features
 ### Enhancements
 - Make thread pool sizes configurable ([#1139](https://github.com/opensearch-project/flow-framework/issues/1139))
 
 ### Bug Fixes
 ### Infrastructure
+- Conditionally include ddb-client dependency only if env variable set ([#1141](https://github.com/opensearch-project/flow-framework/issues/1141))
+
 ### Documentation
 ### Maintenance
 ### Refactoring
--- a/build.gradle
+++ b/build.gradle
@@ -35,6 +35,18 @@ buildscript {
         bwcFlowFrameworkPath = bwcFilePath + "flowframework/"
 
         isSameMajorVersion = opensearch_version.split("\\.")[0] == bwcVersionShort.split("\\.")[0]
+
+        // Version constants for dependencies
+        awsEncryptionSdkVersion = "3.0.1"
+        awsCryptoMaterialProvidersVersion = "1.10.1"
+        dafnyRuntimeVersion = "4.10.0"
+        smithyDafnyVersion = "0.1.1"
+        gsonVersion = "2.13.1"
+        junitJupiterVersion = "5.12.2"
+        jakartaJsonBindVersion = "3.0.1"
+        jakartaJsonVersion = "2.0.1"
+        yassonVersion = "3.0.4"
+        parssonVersion = "1.1.7"
         swaggerVersion = "2.1.28"
         swaggerCoreVersion = "2.2.32"
     }
@@ -181,16 +193,16 @@ dependencies {
     }
     implementation("org.apache.commons:commons-lang3:${versions.commonslang}")
     implementation("org.opensearch:common-utils:${common_utils_version}")
-    implementation("com.amazonaws:aws-encryption-sdk-java:3.0.1")
-    implementation("software.amazon.cryptography:aws-cryptographic-material-providers:1.10.1")
-    implementation("org.dafny:DafnyRuntime:4.10.0")
-    implementation("software.amazon.smithy.dafny:conversion:0.1.1")
+    implementation("com.amazonaws:aws-encryption-sdk-java:${awsEncryptionSdkVersion}")
+    implementation("software.amazon.cryptography:aws-cryptographic-material-providers:${awsCryptoMaterialProvidersVersion}")
+    implementation("org.dafny:DafnyRuntime:${dafnyRuntimeVersion}")
+    implementation("software.amazon.smithy.dafny:conversion:${smithyDafnyVersion}")
     implementation("org.bouncycastle:bc-fips:${versions.bouncycastle_jce}")
     api("org.apache.httpcomponents.core5:httpcore5:${versions.httpcore5}")
-    implementation("jakarta.json.bind:jakarta.json.bind-api:3.0.1")
-    implementation("org.glassfish:jakarta.json:2.0.1")
-    implementation("org.eclipse:yasson:3.0.4")
-    implementation("com.google.code.gson:gson:2.13.1")
+    implementation("jakarta.json.bind:jakarta.json.bind-api:${jakartaJsonBindVersion}")
+    implementation("org.glassfish:jakarta.json:${jakartaJsonVersion}")
+    implementation("org.eclipse:yasson:${yassonVersion}")
+    implementation("com.google.code.gson:gson:${gsonVersion}")
 
     // Swagger-Parser dependencies for API consistency tests
     implementation("io.swagger.core.v3:swagger-models:${swaggerCoreVersion}")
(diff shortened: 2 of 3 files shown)


build.gradle
```diff
@@ -35,8 +35,24 @@ buildscript {
...
+        smithyDafnyVersion = "0.1.1"
+        gsonVersion = "2.13.1"
+        junitJupiterVersion = "5.12.2"
+        jakartaJsonBindVersion = "3.0.1"
+        jakartaJsonVersion = "2.0.1"
+        yassonVersion = "3.0.4"
+        parssonVersion = "1.1.7"
         swaggerVersion = "2.1.26"
         swaggerCoreVersion = "2.2.30"
+        httpclient5Version = "5.4.4"
```
dhrubo-os, line 52: can we use version from core: https://github.com/opensearch-project/OpenSearch/blob/main/gradle/libs.versions.toml#L45

build.gradle
```diff
@@ -211,13 +232,32 @@ dependencies {
 
     configurations.all {
         resolutionStrategy {
-            force("com.google.guava:guava:33.4.8-jre") // CVE for 31.1, keep to force transitive dependencies
+            force("com.google.guava:guava:${versions.guava}")
+
+            // Jackson dependencies
+            force("com.fasterxml.jackson.core:jackson-core:${versions.jackson}")
```
dhrubo-os, line 238: Just curious why do we need to force all of these libraries. What other version are they getting? Also from where?

dbwiddis: OpenSearch Java client brings in 2.17.0.

DDB Client brings in 2.18.2.

They come in via the DDB client (depending ultimately on AOS client) via jackson BOM.
```
com.fasterxml.jackson.core:jackson-core:2.18.2 FAILED
+--- com.fasterxml.jackson:jackson-bom:2.18.2
|    +--- testRuntimeClasspath
|    +--- com.fasterxml.jackson.datatype:jackson-datatype-jsr310:2.18.2
|    |    +--- testRuntimeClasspath
|    |    +--- org.opensearch:opensearch-remote-metadata-sdk-ddb-client:3.1.0.0-SNAPSHOT:20250507.054903-3
```

dhrubo-os: what about other libraries like: `opensearch-rest-client`, `jakarta`, `yasson`, `parsson`, `json`, `httpclient5`. Do we need to force versions for all of them?

dbwiddis: > what about other libraries ... Do we need to force versions for all of them?

All of these come from the `opensearch-java` client (which is 2.x compatible) and conflict with `opensearch` 3.x.  Until we have a 3.x version of the java client we have conflicts.

Options are to either force or exclude from the transitive bringing them in, if we are certain we're getting them.

The problem with excluding is that we only bring them in here as test dependencies (for integ tests which use clients) so we'd need to upgrade those to runtime dependencies (how they are used in SDK).

We could force to 3.x versions at SDK Client, but that means that upstream client would be responsible for maintaining everyone's versions, and I don't want to assume that extra management burden.

dbwiddis: Possible long-term solutions:
1. For jackson, see if using the version catalog `aws` dependency resolves it at SDK level: https://github.com/opensearch-project/OpenSearch/blob/3118fb621e82c6c8ed036ef3a83ea93946c1a0fb/gradle/libs.versions.toml#L56
2. For OpenSearch Java Client, re-evaluate when there is a 3.x version of the client

dhrubo-os (approved):

dhrubo-os:
Do dependent plugins like ml-commons/flow framework need to do anything based on this change?

build.gradle
dbwiddis: > For jackson, see if using the version catalog aws dependency resolves it at SDK level

That's a good change to make but it's still the same `opensearch-java` issue and the same "exclude and upgrade test dependencies to runtime" vs. "force version" tradeoff.  I'll add a comment on them to make it more clear and track revisiting it later.

dbwiddis:
> Do dependent plugins like ml-commons/flow framework need to do anything based on this change?

The only change upstream is aligning bom version with OpenSearch which is getting version bumps that were otherwise frozen. I think ML Commons also uses BOM-related dependencies and I'd recommend using the version catalog if you aren't already.

But otherwise no change.   It's your call at ML commons if you want to conditionally include DDB dependencies, or just have it there all the time even if unused.   But adding it may (as seen here) introduce some jar hell with the java client runtime vs. test dependencies.

build.gradle
dbwiddis: Looks like OpenSearch Java v3 is out. Let me try to see if that resolves all this.

junweid62 (approved):

CHANGELOG.md
```diff
@@ -3,20 +3,13 @@ All notable changes to this project are documented in this file.
 
 Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 
-## [Unreleased 3.x](https://github.com/opensearch-project/flow-framework/compare/3.0...HEAD)
+## [Unreleased 3.1](https://github.com/opensearch-project/flow-framework/compare/3.0...HEAD)
```
owaiskazi19, line 6: Don't we keep this as 3.x and pick up the changes to a release note files separately?

build.gradle
```diff
@@ -193,11 +205,16 @@ dependencies {
     implementation("io.swagger.parser.v3:swagger-parser-v3:${swaggerVersion}")
 
     // Multi-tenant SDK Client
-    implementation("org.opensearch:opensearch-remote-metadata-sdk:${opensearch_build}") {
-        exclude group: "org.apache.httpcomponents.client5", module: "httpclient5"
+    if (System.getenv('REMOTE_METADATA_SDK_IMPL') == 'ddb-client') {
+        api("org.opensearch:opensearch-remote-metadata-sdk:${opensearch_build}")
```
owaiskazi19, line 209: Can create a constant for this dependency
```
def opensearchSdk = "org.opensearch:opensearch-remote-metadata-sdk:${opensearch_build}"
```
and use it then 
`api(opensearchSdk) `
in the else block
`implementation(opensearchSdk)`

build.gradle
```diff
@@ -211,13 +228,37 @@ dependencies {
...
+                force("org.eclipse:yasson:${yassonVersion}")
+                force("org.eclipse.parsson:parsson:${parssonVersion}")
+                force("org.glassfish:jakarta.json:${jakartaJsonVersion}")
+
+                // Apache and Commons dependencies
+                force("org.apache.httpcomponents.client5:httpclient5:${versions.httpclient5}")
+                force("org.apache.httpcomponents:httpcore:${versions.httpcore}")
+                force("commons-codec:commons-codec:${versions.commonscodec}")
+                force("commons-logging:commons-logging:${versions.commonslogging}")
+            }
```
owaiskazi19, line 257: Can we have an issue to remove these force dependencies later?

owaiskazi19 (commented):
Looks good overall with minor nitpicks

CHANGELOG.md
dbwiddis: The `x` or `1` is just a label. The important part is what we are comparing against on the link, which needs to refer to the latest release (currently 3.0).

We will have to change this line anyway when we reset the changelog after a release (after 3.1.0 we will change the link to reference `3.1...HEAD`) so I don't think it matters much whether we put x or 1.  I'm happy to put this back if there's a strong opinion to do so (how are other plugins doing it?)

build.gradle
dbwiddis: We can, but this seems to complicate it a bit.   I had considered making the next line dynamic based on the value of `REMOTE_METADATA_SDK_IMPL` (so `opensearchSdk + "-" + System.getenv('REMOTE_METADATA_SDK_IMPL')) for example.   If we're going to create constants let's go all the way.

For now with only one option, I think it's reasonable to just keep the single repetition.  IF we ever get to more than one branch, we can get fancy.

build.gradle
dbwiddis: Unfortunately not.  I have updated SDK to `openSearch-java` version 3.0 and there are still version conflicts as the java client does not keep in sync with the version catalog.

CHANGELOG.md
owaiskazi19: Kept it as 3.x in neural search https://github.com/opensearch-project/neural-search/blob/main/CHANGELOG.md?plain=1#L6

build.gradle
owaiskazi19: Should we open an issue on java client side to keep the dependency in sync then? That would avoid forcing dependencies on the plugins side

build.gradle
owaiskazi19: I am fine keeping it as it is. But trying to understand more here what's complicated in creating the constant and using it for api and implementation?

build.gradle
dbwiddis: I don't think we can keep them in sync.  The OpenSearch Java Client does not share version/release coupling with OpenSearch itself.  Unlike OpenSearch dependencies (where we use -SNAPSHOT versions) we are using a fixed release version of the Java Client.

Assuming it was possible to sync Java Client to the version catalog (I'm not sure if it is but play along for a moment) we'd have to rely on a SNAPSHOT client for normal CI builds but we would use a non-SNAPSHOT version for the release build; which since we don't release at the same time would be out of date and break the RC build.

Some discussion at https://github.com/opensearch-project/flow-framework/pull/1141#discussion_r2085332193 on another option, forcing the versions at the SDK level but that removes the independence of plugins to use their own versions so I don't think it's a good idea.  (It may be reasonable to do for version catalog dependencies, though.)

build.gradle
dbwiddis: > trying to understand more here what's complicated in creating the constant and using it for api and implementation

It's not, but why would we do it differently than any other repeated string?  

1. Why obfuscate a dependency's maven coordinates just to prevent a repetition?
   - Why not make a `httpclient5` constant too?
   - And would that mess up dependabot/mend resolution?
2. It makes a lot of sense to have the dependencies for "foo" and "foo-ddb-client" right next to each other, rather than substituting only one of them with a constant.
   - The actual repeated string that should be a constant is "org.opensearch:opensearch-remote-metadata-sdk" because then we can append "ddb-client" (or whatever the environment variable is) to it to get the client constant, and then we can have both abbreviated "sdk" and "sdk-client" constants next to each other if that's what we want
   - I actually did consider doing this but think it overcomplicates a simple if/else right now
   - I actually would do that if we end up adding conditionals for more clients, which we probably will in the future, but for now I'm just trying to handle the immediate need
3. We don't really make constants for very many other strings other than versions, and intentionally follow the same pattern for consistency.  We do have `org.opensearch` as a group constant but don't substitute it in dependency strings.  It reduces readability.  People are used to seeing maven coordinates in implementation/api statements, not constants they need to look up elsewhere in the file
4. There is a lot more repetition in the `integTest` section in various inclusion/exclusion test paths if we really want to make a lot of things constants.  I don't think it's really necessary.

owaiskazi19 (approved):

PR #1151, opened 2025-05-29
Fixing llm field processing in RegisterAgentStep

joshpalis:
### Description
Resolves llm field input processing for the RegisterAgentStep. Current behavior fails when a model ID is provided, the following json is an example input for the llm field :
```
"llm": {
    "model_id": "<model_id>",
    "parameters": {
      "max_iteration": 5,
      "system_instruction": "You are a helpful assistant.",
      "prompt": "${parameters.question}"
    }
```

### Related Issues
n/a

### Check List

### DIFF
--- a/src/main/java/org/opensearch/flowframework/workflow/RegisterAgentStep.java
+++ b/src/main/java/org/opensearch/flowframework/workflow/RegisterAgentStep.java
@@ -55,7 +55,6 @@ import static org.opensearch.flowframework.common.CommonValue.PARAMETERS_FIELD;
 import static org.opensearch.flowframework.common.CommonValue.TOOLS_FIELD;
 import static org.opensearch.flowframework.common.CommonValue.TOOLS_ORDER_FIELD;
 import static org.opensearch.flowframework.common.CommonValue.TYPE;
-import static org.opensearch.flowframework.common.WorkflowResources.MODEL_ID;
 import static org.opensearch.flowframework.exception.WorkflowStepException.getSafeException;
 import static org.opensearch.flowframework.util.ParseUtils.getStringToStringMap;
 
@@ -170,7 +169,7 @@ public class RegisterAgentStep implements WorkflowStep {
                     Object llmParams = llmFieldMap.get(PARAMETERS_FIELD);
 
                     if (llmParams != null) {
-                        llmParameters.putAll(getStringToStringMap(llmParams, PARAMETERS_FIELD));
+                        validateLLMParametersMap(llmParams);
                     }
                 } catch (IllegalArgumentException ex) {
                     String errorMessage = "Failed to parse llm field: " + ex.getMessage();
@@ -310,5 +309,18 @@ public class RegisterAgentStep implements WorkflowStep {
     private Map<String, Object> getParseFieldMap(String llmFieldMapString) throws OpenSearchParseException {
         BytesReference llmFieldBytes = new BytesArray(llmFieldMapString.getBytes(StandardCharsets.UTF_8));
         return XContentHelper.convertToMap(llmFieldBytes, false, MediaTypeRegistry.JSON).v2();
+    }
+
+    private void validateLLMParametersMap(Object llmParams) {
+        String errorMessage = "llm field [" + PARAMETERS_FIELD + "] must be a string to string map";
+        if (!(llmParams instanceof Map)) {
+            throw new IllegalArgumentException(errorMessage);
+        }
+        Map<String, Object> llmParamsMap = (Map<String, Object>) llmParams;
+        for (Map.Entry<String, Object> entry : llmParamsMap.entrySet()) {
+            if (!(entry.getValue() instanceof String)) {
+                throw new IllegalArgumentException(errorMessage);
+            }
+        }
     }
 }
--- a/src/main/java/org/opensearch/flowframework/workflow/RegisterAgentStep.java
+++ b/src/main/java/org/opensearch/flowframework/workflow/RegisterAgentStep.java
@@ -174,7 +174,7 @@ public class RegisterAgentStep implements WorkflowStep {
                         llmParameters.putAll(getStringToStringMap(llmParams, PARAMETERS_FIELD));
                     }
                 } catch (Exception ex) {
-                    String errorMessage = "Failed to create llm configuration";
+                    String errorMessage = "Invalid llm field format";
                     logger.error(errorMessage, ex);
                     registerAgentModelFuture.onFailure(new WorkflowStepException(errorMessage, RestStatus.BAD_REQUEST));
                 }
--- a/src/main/java/org/opensearch/flowframework/workflow/RegisterAgentStep.java
+++ b/src/main/java/org/opensearch/flowframework/workflow/RegisterAgentStep.java
@@ -11,6 +11,7 @@ package org.opensearch.flowframework.workflow;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.opensearch.ExceptionsHelper;
+import org.opensearch.OpenSearchParseException;
 import org.opensearch.action.support.PlainActionFuture;
 import org.opensearch.common.Nullable;
 import org.opensearch.common.xcontent.XContentHelper;
@@ -164,17 +165,18 @@ public class RegisterAgentStep implements WorkflowStep {
             if (llmField != null) {
                 try {
                     // Convert llm field string to map
-                    Map<String, Object> llmFieldMap = getLLMFieldMap(llmField);
+                    Map<String, Object> llmFieldMap = getParseFieldMap(llmField);
                     llmModelId = (String) llmFieldMap.get(MODEL_ID);
                     Object llmParams = llmFieldMap.get(PARAMETERS_FIELD);
 
                     if (llmParams != null) {
                         llmParameters.putAll(getStringToStringMap(llmParams, PARAMETERS_FIELD));
                     }
-                } catch (Exception ex) {
-                    String errorMessage = "Invalid register agent llm field format";
+                } catch (IllegalArgumentException ex) {
+                    String errorMessage = "Failed to parse llm field: " + ex.getMessage();
                     logger.error(errorMessage, ex);
-                    registerAgentModelFuture.onFailure(new WorkflowStepException(errorMessage, RestStatus.BAD_REQUEST));
+                    registerAgentModelFuture.onFailure(new WorkflowStepException(ex.getMessage(), RestStatus.BAD_REQUEST));
+                    return registerAgentModelFuture;
                 }
             }
 
@@ -305,7 +307,7 @@ public class RegisterAgentStep implements WorkflowStep {
         return builder.build();
     }
 
-    private Map<String, Object> getLLMFieldMap(String llmFieldMapString) {
+    private Map<String, Object> getParseFieldMap(String llmFieldMapString) throws OpenSearchParseException {
         BytesReference llmFieldBytes = new BytesArray(llmFieldMapString.getBytes(StandardCharsets.UTF_8));
         return XContentHelper.convertToMap(llmFieldBytes, false, MediaTypeRegistry.JSON).v2();
     }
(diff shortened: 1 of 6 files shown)


dbwiddis (commented):
LGTM so far.  Can we have a test class?

Does this need to be backported to 2.19.3?

src/main/java/org/opensearch/flowframework/workflow/RegisterAgentStep.java
```diff
@@ -158,6 +159,27 @@ public void onFailure(Exception ex) {
...
+                    Map<String, Object> llmFieldMap = XContentHelper.convertToMap(llmFieldBytes, false, MediaTypeRegistry.JSON).v2();
+
+                    llmModelId = (String) llmFieldMap.get(MODEL_ID);
+                    Object llmParams = llmFieldMap.get(PARAMETERS_FIELD);
+
+                    if (llmParams != null) {
+                        llmParameters.putAll(getStringToStringMap(llmParams, PARAMETERS_FIELD));
+                    }
+                } catch (Exception ex) {
+                    String errorMessage = "Failed to create llm configuration";
```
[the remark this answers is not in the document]
joshpalis: will do

joshpalis:
> Can you add corresponding UT to ensure the parsing works?

@ohltyler I have updated the existing unit and integration tests to ensure parsing works

> LGTM so far. Can we have a test class?
Does this need to be backported to 2.19.3?

@dbwiddis not sure if need to backport this, @ohltyler can you confirm

ohltyler:
> @dbwiddis not sure if need to backport this, @ohltyler can you confirm

Backport not required for UI, will leave it up to your discretion to include in 2.19.3. If easy enough, I'd suggest yes.

src/main/java/org/opensearch/flowframework/workflow/RegisterAgentStep.java
```diff
@@ -158,6 +159,27 @@ public void onFailure(Exception ex) {
             Instant lastUpdateTime = createdTime;
             String appType = (String) inputs.get(APP_TYPE_FIELD);
 
+            String llmModelId = null;
+            Map<String, String> llmParameters = new HashMap<>();
+            if (llmField != null) {
+                try {
+                    // Convert model interface string to map
+                    BytesReference llmFieldBytes = new BytesArray(llmField.getBytes(StandardCharsets.UTF_8));
+                    Map<String, Object> llmFieldMap = XContentHelper.convertToMap(llmFieldBytes, false, MediaTypeRegistry.JSON).v2();
```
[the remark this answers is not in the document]
joshpalis: done

src/main/java/org/opensearch/flowframework/workflow/RegisterAgentStep.java
```diff
@@ -284,4 +304,9 @@ private MLMemorySpec getMLMemorySpec(Object mlMemory) {
 
         return builder.build();
     }
+
+    private Map<String, Object> getLLMFieldMap(String llmFieldMapString) {
+        BytesReference llmFieldBytes = new BytesArray(llmFieldMapString.getBytes(StandardCharsets.UTF_8));
+        return XContentHelper.convertToMap(llmFieldBytes, false, MediaTypeRegistry.JSON).v2();
```
dbwiddis, line 310: This method throws an `OpenSearchParseException`.   This has useful parsing information and a `status()` that can be used by the calling method.

While technically optional to add `throws` here, probably would help readability as we're doing the same as the method we're calling.

src/main/java/org/opensearch/flowframework/workflow/RegisterAgentStep.java
```diff
@@ -158,6 +159,25 @@ public void onFailure(Exception ex) {
             Instant lastUpdateTime = createdTime;
             String appType = (String) inputs.get(APP_TYPE_FIELD);
 
+            String llmModelId = null;
+            Map<String, String> llmParameters = new HashMap<>();
+            if (llmField != null) {
+                try {
+                    // Convert llm field string to map
+                    Map<String, Object> llmFieldMap = getLLMFieldMap(llmField);
```
dbwiddis, line 167: We should more narrowly catch the `ParseException` (or `OpenSearchException`) here to keep the same parse failure message rather than a generic "invalid format" from the more broad catch block.

src/main/java/org/opensearch/flowframework/workflow/RegisterAgentStep.java
```diff
@@ -284,4 +304,9 @@ private MLMemorySpec getMLMemorySpec(Object mlMemory) {
 
         return builder.build();
     }
+
+    private Map<String, Object> getLLMFieldMap(String llmFieldMapString) {
```
dbwiddis, line 308: Probably can name this more generally.  Perhaps `parseFieldMap()`?

src/main/java/org/opensearch/flowframework/workflow/RegisterAgentStep.java
```diff
@@ -158,6 +159,25 @@ public void onFailure(Exception ex) {
...
+                    llmModelId = (String) llmFieldMap.get(MODEL_ID);
+                    Object llmParams = llmFieldMap.get(PARAMETERS_FIELD);
+
+                    if (llmParams != null) {
+                        llmParameters.putAll(getStringToStringMap(llmParams, PARAMETERS_FIELD));
+                    }
+                } catch (Exception ex) {
+                    String errorMessage = "Invalid register agent llm field format";
+                    logger.error(errorMessage, ex);
+                    registerAgentModelFuture.onFailure(new WorkflowStepException(errorMessage, RestStatus.BAD_REQUEST));
```
dbwiddis, line 177: as the codecov warning notes, need a test case for the exception catching

dbwiddis:
Why are we not making a change log entry? This seems like a bug fix.

src/main/java/org/opensearch/flowframework/workflow/RegisterAgentStep.java
joshpalis: This actually throws an illegal argument exception rather than an OpenSearchParseException, I will update the error handling in the next commit so we can pass the error back to the user. 
```
[2025-05-29T20:18:16,193][ERROR][o.o.f.w.RegisterAgentStep] [integTest-0] Failed to parse llm field: [parameters] must be a key-value map.
java.lang.IllegalArgumentException: [parameters] must be a key-value map.
        at org.opensearch.flowframework.util.ParseUtils.getStringToStringMap(ParseUtils.java:554) ~[?:?]
        at org.opensearch.flowframework.workflow.RegisterAgentStep.execute(RegisterAgentStep.java:173) ~[?:?]
        at org.opensearch.flowframework.workflow.ProcessNode.lambda$execute$0(ProcessNode.java:182) ~[?:?]
        at java.base/java.util.concurrent.CompletableFuture$AsyncRun.run(CompletableFuture.java:1804) ~[?:?]
        at org.opensearch.common.util.concurrent.ThreadContext$ContextPreservingRunnable.run(ThreadContext.java:916) ~[opensearch-3.1.0-SNAPSHOT.jar:3.1.0-SNAPSHOT]
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144) ~[?:?]
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642) ~[?:?]
        at java.base/java.lang.Thread.run(Thread.java:1583) [?:?]
```

dbwiddis: > This actually throws an illegal argument exception 

That's at our own parsing level of "is this even a map" but then there's further XContent validation in the helper method that could fail (e.g., an extra closing `}` might trip it)

joshpalis: Ive attempted to do so but any invalid json format (such as extra `}`) is already caught during creation of the workflow template before its indexed. I believe this level of error handling is sufficient for this RegisterAgentStep

dbwiddis: Try putting an integer value above Integer.MAX_VALUE in a field.  See https://github.com/opensearch-project/OpenSearch/issues/18131

joshpalis: So I attempted to create an agent with an llm parameter entry of int with value greater than max and this fails to create an agent. 

```
"llm":{"model_id":"J7DfGJcBjIjt_6rk_MOd","parameters":{"test":"test", "number": 2147483648}}
```

The issue however doesnt seem to be parsing but the Register Agent response from ML Commons is failing with 5xx due to : 
```
class java.lang.Long cannot be cast to class java.lang.String (java.lang.Long and java.lang.String are in module java.base of loader 'bootstrap')
[2025-05-29T23:53:22,187][ERROR][o.o.f.w.RegisterAgentStep] [integTest-0] Failed to register the agent
[2025-05-29T23:53:22,187][ERROR][o.o.f.t.ProvisionWorkflowTransportAction] [integTest-0] Provisioning failed for workflow LQ92HpcBHeQerNNc9JHy during step chat_agent.
org.opensearch.flowframework.exception.WorkflowStepException: Failed to register the agent
        at org.opensearch.flowframework.workflow.RegisterAgentStep$1.onFailure(RegisterAgentStep.java:122) ~[?:?]
```

dbwiddis: > So I attempted to create an agent with an llm parameter entry of int with value greater than max and this fails to create an agent.

Yeah but that's on provisioning.  And should be 400 not 500.

Would be ideal if we didn't even get that far, see #1152 ...

I got an NPE using `"parameters",null` in your test... still trying to get that parse exception.

joshpalis: > In your unit test if I replace line 198 with Map.entry("llm", "{invalid}"), I get:

In practice, I don't this this situation would ever occur since we would not be able to parse and index a workflow that has an invalid json structure. I have tested this scenario below:
```
curl -i -XPOST "localhost:9200/_plugins/_flow_framework/workflow?provision=true" -H "Content-Type:application/json" --data '{"name":"chatbot","use_case":"REGISTER_AGENT","description":"","version":{"template":"1.0.0","compatibility":["3.0.0"]},"workflows":{"provision":{"nodes":[{"id":"chat_agent","type":"register_agent","previous_node_inputs":{},"user_inputs":{"name":"default_agent","type":"conversational","description":"","llm":{invalid},"memory":{"type":"conversation_index"},"parameters":{},"tools":[],"app_type":"os_chat"}}],"edges":[],"user_params":{}}}}'
HTTP/1.1 400 Bad Request
X-OpenSearch-Version: OpenSearch/3.1.0-SNAPSHOT (opensearch)
content-length: 77

{"error":"IOException: template content invalid for specified Content-Type."}
``` 

> > So I attempted to create an agent with an llm parameter entry of int with value greater than max and this fails to create an agent.
> 
> Yeah but that's on provisioning. And should be 400 not 500.
> 
Looking into the MLCommons LLM Spec, seems it expects the field map to strictly be string to string, hence the error when attempting to create the agent. Since we are creating agents over transport, this input validation doesnt occur which explains the 500 error. To address this, I have added validation for this case in the latest commit. 

 ```
 curl -i -XGET "localhost:9200/_plugins/_flow_framework/workflow/v6qPIJcBbsfGW0-eZOku/_status?pretty&&all=true"
HTTP/1.1 200 OK
X-OpenSearch-Version: OpenSearch/3.1.0-SNAPSHOT (opensearch)
content-length: 304

{
  "workflow_id" : "v6qPIJcBbsfGW0-eZOku",
  "error" : "llm field [parameters] must be a string to string map during step chat_agent, restStatus: BAD_REQUEST",
  "state" : "FAILED",
  "provisioning_progress" : "FAILED",
  "provision_start_time" : 1748597957910,
  "provision_end_time" : 1748597957943
}
 ```

dbwiddis:
Does this bug exist on 2.19? Should we backport?

PR #1168, opened 2025-06-17
feat(util): Add JsonToJson Recommender as a utility function

### LINKED ISSUE #1167: [FEATURE] JsonToJson Recommender
Across different functional components of OpenSearch—such as query parsing, semantic retrieval expansion, machine learning inference, and invoking large language models (LLMs) for analysis and response handling—each component follows its own JSON input and output structure. This structural inconsistency gives rise to a number of practical challenges.

1. Significant differences in data structures between components complicate process integration:
     The JSON outputs of different modules vary greatly in field naming, nesting depth, and semantic representation, making them incompatible for direct consumption by downstream components. Developers are forced to manually implement transformation logic between each pair of modules, resulting in integration processes that are both cumbersome and fragile.
2. High learning and manual maintenance cost with low accuracy:
     Understanding and writing field mapping logic requires deep knowledge of both upstream and downstream data structures. Developers must spend considerable time comparing and converting formats, crafting prompts, and verifying outputs — a process that is error-prone and results in inconsistent accuracy.
3. Mapping rules lack reusability and maintainability, leading to inefficiency:
     Each format conversion typically requires writing mapping logic from scratch. Existing mappings cannot be effectively reused, and even minor structural changes demand significant rework, negatively impacting overall development efficiency and maintainability.

This issue is related to OpenSearch [issue #18507](https://github.com/opensearch-project/OpenSearch/issues/18507) and applies to Flow Framework in the same way.

To address the structural incompatibility and high integration overhead commonly encountered in OpenSearch query processing, we propose the design and implementation of a general-purpose, decoupled transformation function: JsonToJsonRecommender. 
The function is designed with the following key characteristics:

* Automated Mapping Generation: Automatically analyzes the structural and semantic relationships between source and target JSON structures, and produces accurate, interpretable transformation mappings. This enables seamless adaptation across components with heterogeneous data formats.
* Flexible Integration: Operates as a self-contained entity deployable at various points in the query lifecycle, independent of specific upstream or downstream systems, supporting flexible integration within heterogeneous pipelines.
* Generality: Accepts any valid JSON text as input, supporting a wide variety of structures.
* Visibility: Outputs clear field mapping suggestions using JsonPath, making it easy to debug, validate, and reuse.

This utility analyzes a pair of JSON structures — a source (inputJson) and a target (outputJson) — and produces transformation suggestions using JsonPath expressions. These mappings help identify how to extract or rearrange fields from the source JSON to match the structure of the target JSON.

We have previously evaluated using JOLT as a transformation solution. JOLT supports structural reorganization and dynamic key generation, offering advantages in handling complex structural changes. However:
- JOLT has a complex syntax with a steep learning curve;
- It does not support direct data extraction or field comparison;
- It is better suited for restructuring rather than the majority of use cases that only require field remapping.

In contrast, JsonPath is more lightweight and intuitive, making it well-suited for our primary use cases of "reading, filtering, and remapping." Additionally, we already have relevant dependencies in place, resulting in lower integration costs.

Design diagram

Image

Chris-tyt:
### Description
This PR adds a utility class `JsonToJsonRecommender` under `util` that can infer JSONPath mappings based on example input and output JSON objects.

This utility is intended as a core component for implementing automatic field-level mapping in scenarios like inference-based pipeline configuration or input/output transformation logic.

The unit tests include example input-output pairs and cover relevant cases.

### Related Issues
Resolves #1167

### Check List

### DIFF
--- a/src/main/java/org/opensearch/flowframework/util/JsonToJsonRecommender.java
+++ b/src/main/java/org/opensearch/flowframework/util/JsonToJsonRecommender.java
@@ -1,0 +1,694 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+package org.opensearch.flowframework.util;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.core.StreamReadConstraints;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+import org.opensearch.flowframework.common.CommonValue;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+/**
+ * Utility class for analyzing and mapping relationships between input and
+ * output JSON structures.
+ *
+ * Builds an inverted index from input JSON values to their paths, then uses
+ * output JSON values
+ * to find matching input paths, producing a mapping from input paths to output
+ * fields.
+ *
+ * Supports nested objects and arrays, and generates both detailed field
+ * mappings and generalized
+ * JSONPath transformation patterns.
+ *
+ * All methods are static and stateless. This class is thread-safe.
+ */
+public class JsonToJsonRecommender {
+    /**
+     * Private constructor to prevent instantiation of this utility class.
+     * All methods are static and should be accessed directly.
+     */
+    private JsonToJsonRecommender() {
+        // Utility class, not meant to be instantiated
+    }
+
+    /**
+     * ObjectMapper instance used for parsing JSON strings and formatting output.
+     */
+    private static final ObjectMapper MAPPER;
+    static {
+        StreamReadConstraints constraints = StreamReadConstraints.builder()
+            .maxNestingDepth(CommonValue.MAX_JSON_NESTING_DEPTH)
+            .maxStringLength(CommonValue.MAX_JSON_SIZE)
+            .maxNameLength(CommonValue.MAX_JSON_NAME_LENGTH)
+            .build();
+
+        MAPPER = new ObjectMapper();
+        MAPPER.getFactory().setStreamReadConstraints(constraints);
+    }
+
+    /**
+     * Generates mapping data with default settings.
+     * This method generates both detailed field mappings and generalized
+     * JSONPath transformation patterns.
+     *
+     * @param inputJson  Input JSON string to analyze
+     * @param outputJson Output JSON string to map against
+     * @return StringFormatResult containing transformation recommendations
+     * @throws IllegalArgumentException if input validation fails
+     * @throws JsonProcessingException  if JSON parsing fails
+     */
+    public static StringFormatResult getRecommendationInStringFormat(String inputJson, String outputJson) throws IllegalArgumentException,
+        JsonProcessingException {
+        // Input validation
+        validateInputStrings(inputJson, outputJson);
+
+        JsonNode inputNode = MAPPER.readTree(inputJson);
+        JsonNode outputNode = MAPPER.readTree(outputJson);
+
+        MapFormatResult result = getRecommendationInMapFormat(inputNode, outputNode, false);
+
+        String detailedJsonPathString = WriteMappedPathAsString(result.detailedJsonPath);
+        String generalizedJsonPathString = WriteMappedPathAsString(result.generalizedJsonPath);
+
+        return new StringFormatResult(detailedJsonPathString, generalizedJsonPathString);
+    }
+
+    /**
+     * Core method that generates mapping recommendations from JsonNode inputs.
+     * This is a convenience method that calls
+     * {@link #getRecommendationInMapFormat(JsonNode, JsonNode, boolean)}
+     * with needValidation set to true.
+     *
+     * @param inputNode  Input JSON node to analyze
+     * @param outputNode Output JSON node to map against
+     * @return MapFormatResult containing both detailed and generalized mappings
+     * @throws IllegalArgumentException if input validation fails
+     */
+    public static MapFormatResult getRecommendationInMapFormat(JsonNode inputNode, JsonNode outputNode) throws IllegalArgumentException {
+        return getRecommendationInMapFormat(inputNode, outputNode, true);
+    }
+
+    /**
+     * Core method that generates mapping recommendations from JsonNode inputs.
+     * This is the main internal method for getting transformation recommendations
+     * from input and output JSON structures.
+     *
+     * @param inputNode      Input JSON node to analyze
+     * @param outputNode     Output JSON node to map against
+     * @param needValidation If true, validates the nesting depth and key lengths of JsonNodes;
+     *                       otherwise skips validation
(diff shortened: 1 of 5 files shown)


junweid62 (commented):
LGTM! The implementation is solid and well-structured, but the class is quite long (550+ lines) and takes on multiple responsibilities (indexing, mapping, formatting, validation). To improve maintainability and readability, I suggest splitting it into smaller focused components

joshpalis:
I would suggest rebasing the PR against main again since this branch is out of date

src/main/java/org/opensearch/flowframework/util/JsonToJsonRecommender.java
```diff
@@ -0,0 +1,590 @@
...
+
+    /**
+     * Formats the detailed mapping in a tabular format.
+     * Creates a well-formatted table showing the relationship between output and
+     * input fields.
+     *
+     * @param mapping The detailed mapping to format
+     * @return A formatted string representation of the mapping table
+     */
+    private static String formatJsonPath(Map<String, String> mapping) {
```
joshpalis, line 482: Human readability shouldnt be the goal here as this functionality will be used internally. Consider just outputing the json path itself so it can be consumed directly by downstream components

src/main/java/org/opensearch/flowframework/util/JsonToJsonRecommender.java
```diff
@@ -0,0 +1,590 @@
...
+     * from input and output JSON structures.
+     *
+     * @param inputJson   Input JSON string to analyze
+     * @param outputJson  Output JSON string to map against
+     * @param showDetails If true, includes formatted detailed output in the verbose
+     *                    field;
+     *                    otherwise verbose field will be empty
+     * @return MappingOutput containing:
+     *         - detailedJsonPathString: field-to-field mapping as JSON string
+     *         - generalizedJsonPathString: suggested JsonPaths as JSON string
```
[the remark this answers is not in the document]
joshpalis: +1, @Chris-tyt it would be beneficial to apply the json path to the provided input and comparing the generated output to the provided output. If they don't match, we should throw an error

src/main/java/org/opensearch/flowframework/util/JsonToJsonRecommender.java
```diff
@@ -0,0 +1,558 @@
...
+     * Generates mapping data with default settings (detailed output is empty).
+     * This is a convenience method that calls {@link #getRecommendation(String, String, boolean)}
+     * with showDetails set to false.
+     *
+     * @param inputJson   Input JSON string to analyze
+     * @param outputJson  Output JSON string to map against
+     * @return MappingOutput containing transformation recommendations
+     * @throws Exception if JSON parsing fails or processing encounters errors
+     */
+    public static MappingOutput getRecommendation(String inputJson, String outputJson) throws Exception {
```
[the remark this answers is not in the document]
Chris-tyt: You're right — it's done now.

src/main/java/org/opensearch/flowframework/util/JsonToJsonRecommender.java
```diff
@@ -0,0 +1,635 @@
...
+     * @param inputJson Input JSON string to analyze
+     * @param outputJson Output JSON string to map against
+     * @return Generalized JSONPath suggestions as a formatted JSON string
+     * @throws IllegalArgumentException if input validation fails
+     * @throws JsonProcessingException if JSON parsing fails
+     */
+    public static String getRecommendationGeneralized(String inputJson, String outputJson) throws IllegalArgumentException,
+        JsonProcessingException {
+        return getRecommendation(inputJson, outputJson, false).generalizedJsonPathString;
+    }
```
dbwiddis, line 198: FYI, I used this class to try to implement an API-based call to use this.  See my POC here (not intended for actual use, but could be cool for a demo): https://github.com/opensearch-project/flow-framework/commit/31db9d43d586867c51eb4e6ecf75453295d6be3b

Something to think about is that while Strings are fine low-level, we almost always have the data in a different format than a string.   Sure, you need a wrapping function like this which takes String input and gives String output, but the inner getRecommendation() might benefit from a more flexibile/modular structure.

The first thing you do is convert both strings to a `JsonNode`.  So you should have a method that takes those `JsonNode` arguments as input and output, and returns an appropriate generalized Mapper, before you do all of the pretty-printing and wrapping in your own object.

src/main/java/org/opensearch/flowframework/util/JsonToJsonRecommender.java
joshpalis: The front end will only be able to interact with this functionality via an API, so we can handle formatting once we create it. I think for now its best to just output the json path itself so that we can avoid additional format processing when using the jsonpath in a search request processor for example

src/main/java/org/opensearch/flowframework/util/JsonToJsonRecommender.java
```diff
@@ -0,0 +1,635 @@
...
+        validateInputs(inputJson, outputJson);
+
+        JsonNode inputNode;
+        JsonNode outputNode;
+
+        try {
+            inputNode = MAPPER.readTree(inputJson);
+        } catch (Exception e) {
+            throw new JsonProcessingException("Failed to parse input JSON: " + e.getMessage(), e) {
+            };
```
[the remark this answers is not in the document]
Chris-tyt: Thanks, that makes a lot of sense — there's indeed no special handling needed here. I've removed the catch blocks so the Exception is now allowed to propagate naturally. If we later need to convert it (e.g. to a Exception with line/column info), we can do so by parsing.

src/main/java/org/opensearch/flowframework/util/JsonToJsonRecommender.java
Chris-tyt: Understand. I'll remove it, it's reasonable to handle formatting when we create an api.

dbwiddis: > Understand. I'll remove it, it's reasonable to handle formatting when we create an api.

You can write a helper method that does the formatting stuff to preserve the work you've done, but keep the main method simpler for machine (and unit testing!) use.

src/main/java/org/opensearch/flowframework/util/JsonToJsonRecommender.java
Chris-tyt: Thanks a lot for your implementation — really appreciate it!

You're absolutely right: while String inputs and outputs work for basic usage, in most real-world cases we're working with structured data, not raw strings. Based on your suggestion, I've updated the design to be more modular and flexible.

Now provide two main methods:
- `getRecommendationInStringFormat`: accepts JSON as a `String`, and returns a result wrapper containing the output as a `String` (`StringFormatResult`).
- `getRecommendationInMapFormat`: accepts JSON as a `JsonNode`, and returns a result wrapper containing a `Map<String, Object>` (`MapFormatResult`), which can be further consumed or parsed using `ObjectMapper`.

If you only need the generalized mapping result, there are also:
- `getRecommendationGeneralizedInStringFormat`
- `getRecommendationGeneralizedInMapFormat`

To support output formatting when needed, we've also added a shared utility method `writeMappedPathAsString`, which converts a Map<String, Object> into a formatted String.

joshpalis (approved):
Amazing work Yanting, thank you for taking the time to address my comments, looks good to me

junweid62 (approved):

joshpalis:
Seems the only failing tests are API specs for ML related steps (create connector, register models, etc)
CC : @junweid62

junweid62:
The build failure is unrelated to this PR. I'll proceed with the merge and address the issue in a separate follow-up PR.

PR #1176, opened 2025-06-23
feat(util): Add JsonToJson Transformer

### LINKED ISSUE #1167: [FEATURE] JsonToJson Recommender

Chris-tyt:
### Description
This PR introduces a new utility component: `JsonToJsonTransformer`.

The JsonToJsonTransformer is designed to generate a target outputJson by applying JsonPath mappings to a given inputJson (includes support for wildcard processing). It enables flexible and reusable JSON transformation logic that can adapt to various schema requirements.

Example:
```
inputJson:
{
  "level1": {
    "level2": {
      "level3": {
        "level4": {
          "value": "deep_value",
          "items": [
            {"id": 1, "data": {"info": "item1"}},
            {"id": 2, "data": {"info": "item2"}}
          ]
        }
      }
    }
  }
}

mappingJson:
{
  "deepValue": "$.level1.level2.level3.level4.value",
  "deepItems[*]": {
    "itemId": "$.level1.level2.level3.level4.items[*].id",
    "itemInfo": "$.level1.level2.level3.level4.items[*].data.info"
  }
}


expectedOutput:
{
  "deepValue": "deep_value",
  "deepItems": [
    {
      "itemId": 1,
      "itemInfo": "item1"
    },
    {
      "itemId": 2,
      "itemInfo": "item2"
    }
  ]
}
```

### Related Issues
Resolves #1167 

Related #1168 

### Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,8 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 
 ## [Unreleased 3.2](https://github.com/opensearch-project/flow-framework/compare/3.1...HEAD)
 ### Features
-- Add JsonToJson Recommender as a utility function ([#1168](https://github.com/opensearch-project/flow-framework/issues/1168))
+- Add JsonToJson Recommender as a utility function ([#1167](https://github.com/opensearch-project/flow-framework/issues/1167))([#1168](https://github.com/opensearch-project/flow-framework/pull/1168))
+- Add JsonToJson Transformer as a utility function ([#1167](https://github.com/opensearch-project/flow-framework/issues/1167)) ([#1176](https://github.com/opensearch-project/flow-framework/pull/1176))
 
 ### Enhancements
 ### Bug Fixes
--- a/src/main/java/org/opensearch/flowframework/util/JsonToJsonTransformer.java
+++ b/src/main/java/org/opensearch/flowframework/util/JsonToJsonTransformer.java
@@ -1,0 +1,384 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+package org.opensearch.flowframework.util;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.core.StreamReadConstraints;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ArrayNode;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+
+import org.opensearch.core.common.Strings;
+import org.opensearch.flowframework.common.CommonValue;
+
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.regex.Pattern;
+
+import com.jayway.jsonpath.Configuration;
+import com.jayway.jsonpath.JsonPath;
+import com.jayway.jsonpath.Option;
+import com.jayway.jsonpath.spi.json.JacksonJsonProvider;
+import com.jayway.jsonpath.spi.mapper.JacksonMappingProvider;
+
+/**
+ * Transforms an input JSON into the desired output JSON using
+ * a mapping produced by JsonToJsonRecommender.
+ *
+ * The mapping JSON must have the same structure that the transformer should
+ * emit; leaf nodes contain a valid JsonPath expression that references a value
+ * inside the input JSON. Both fully-qualified paths (e.g. $.foo.bar)
+ * and generalised paths with the [*] wildcard are supported.
+ *
+ * Example mapping (generalised):
+ * <pre>{@code
+ * {
+ *   "name" : "$.item.name",
+ *   "allocDetails[*]" : {
+ *     "allocation" : "$.item.allocDetails[0].useful.items[*].allocation",
+ *     "team" : {
+ *       "id"   : "$.item.allocDetails[0].useful.items[*].team.id",
+ *       "name" : "$.item.allocDetails[0].useful.items[*].team.name"
+ *     }
+ *   }
+ * }
+ * }</pre>
+ * In this example the transformer will copy name directly and
+ * create an allocDetails array whose length equals the number of
+ * matches produced by the allocation JsonPath. Each array element
+ * will contain the corresponding allocation, team.id
+ * and team.name values.
+ *
+ * Implementation notes:
+ * - Uses Jackson for JSON tree manipulation and as JsonPath provider.
+ * - Uses Jayway JsonPath for value extraction with Jackson as the underlying
+ *   JSON provider instead of json-smart to avoid dependency conflicts.
+ * - Only [*] wildcards are supported (nested wildcards are
+ *   allowed). Explicit numeric indices (e.g. [0]) are honoured
+ *   verbatim.
+ * - Thread-safe and stateless: all public APIs are static and use local
+ *   variables only.
+ *
+ * Call Flow Architecture:
+ * transform()
+ *   └── buildOutputFromMapping() [validates object type and processes properties]
+ *       ├── processSimpleProperty() [handles JsonPath leaf nodes]
+ *       ├── [nested objects] → buildOutputFromMapping() [direct recursion]
+ *       └── processArrayProperty() [handles array syntax and builds elements]
+ *           └── buildOutputFromMapping() [direct recursion for array elements]
+ */
+public final class JsonToJsonTransformer {
+    private static final ObjectMapper MAPPER;
+    private static final Configuration JSON_PATH_CFG;
+
+    // Pre-compiled regex patterns for performance
+    private static final Pattern EXPLICIT_INDEX_PATTERN = Pattern.compile("\\[\\d+\\]");
+    private static final Pattern WILDCARD_PATTERN = Pattern.compile("\\[\\*\\]");
+
+    static {
+        StreamReadConstraints constraints = StreamReadConstraints.builder()
+            .maxNestingDepth(CommonValue.MAX_JSON_NESTING_DEPTH)
+            .maxStringLength(CommonValue.MAX_JSON_SIZE)
+            .maxNameLength(CommonValue.MAX_JSON_NAME_LENGTH)
+            .build();
+
+        MAPPER = new ObjectMapper();
+        MAPPER.getFactory().setStreamReadConstraints(constraints);
+
+        JSON_PATH_CFG = Configuration.builder()
+            .jsonProvider(new JacksonJsonProvider(MAPPER))
+            .mappingProvider(new JacksonMappingProvider(MAPPER))
+            .options(Option.ALWAYS_RETURN_LIST, Option.SUPPRESS_EXCEPTIONS)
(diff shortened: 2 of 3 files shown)


src/main/java/org/opensearch/flowframework/util/JsonToJsonTransformer.java
```diff
@@ -0,0 +1,388 @@
...
+        if (mappingRules == null) {
+            throw new IllegalArgumentException("JsonPath must not be null");
+        }
+
+        JsonNode mappingNode;
+        try {
+            mappingNode = MAPPER.readTree(mappingRules);
+        } catch (Exception e) {
+            throw new JsonProcessingException("Failed to parse mapping rules", e) {
+            };
```
dbwiddis, line 127: Using an anonymous subclass of an external dependency seems odd here.  We're already throwing a `JsonProcessingException` on line 124 so catching it and rethrowing with it wrapped in itself seems doubly (pun intended) odd.   Why not only catch the Mapping exception and let the processing exception just throw as-is?

If we're going to catch and rethrow, use an OpenSearch parsing exception type like [`ParsingException`](https://github.com/opensearch-project/OpenSearch/blob/6bf1a6d9ed118216b2adff9233cb1e69271d8025/libs/core/src/main/java/org/opensearch/core/common/ParsingException.java#L52) (if you can get the line/col position) or `IllegalArgumentException`.  (But make sure whatever you use is in [this list](https://github.com/opensearch-project/OpenSearch/blob/6bf1a6d9ed118216b2adff9233cb1e69271d8025/libs/core/src/main/java/org/opensearch/ExceptionsHelper.java#L98-L105).)

src/main/java/org/opensearch/flowframework/util/JsonToJsonTransformer.java
```diff
@@ -0,0 +1,388 @@
...
+            document = JsonPath.using(JSON_PATH_CFG).parse(inputJson).json();
+        } catch (Exception e) {
+            throw new JsonProcessingException("Failed to parse input JSON", e) {
+            };
+        }
+
+        ObjectNode outputRoot = MAPPER.createObjectNode();
+        buildOutputFromMapping(outputRoot, mappingNode, document);
+
+        return MAPPER.writerWithDefaultPrettyPrinter().writeValueAsString(outputRoot);
```
dbwiddis, line 141: (Style) Probably not the right place to put pretty-printing.  (Which makes it harder to unit test the content.)

We might consider changing the return value here to the `ObjectMapper` type and letting the choice of format (pretty-print or not) be made by another method that calls this one.

src/main/java/org/opensearch/flowframework/util/JsonToJsonTransformer.java
```diff
@@ -0,0 +1,388 @@
...
+    private static boolean isArrayProperty(String key) {
+        return key.contains("[") && key.contains("]");
+    }
+
+    /**
+     * Parses array key into components.
+     *
+     * TERMINATION: Simple parsing logic, no recursion.
+     */
+    private static ArrayKeyInfo parseArrayKey(String arrayKey) {
```
dbwiddis, line 288: This manual reconstruction of only some JsonPath features seems concerning.  Doesn't the JsonPath library already include all possible parsing per its spec in a `JsonPath.parse()` method?  Look at the MLInference processor implementations [like this](https://github.com/opensearch-project/ml-commons/blob/a29750f15b6175fe298ec6528babf1251fd795d1/plugin/src/main/java/org/opensearch/ml/processor/MLInferenceSearchRequestProcessor.java#L269-L271)

src/main/java/org/opensearch/flowframework/util/JsonToJsonTransformer.java
```diff
@@ -0,0 +1,388 @@
...
+        int bracketPos = arrayKey.indexOf('[');
+        String propertyName = arrayKey.substring(0, bracketPos);
+        String indexToken = arrayKey.substring(bracketPos);
+
+        if ("[*]".equals(indexToken)) {
+            return new ArrayKeyInfo(propertyName, true, -1);
+        } else if (indexToken.matches("\\[\\d+\\]")) {
```
dbwiddis, line 295: (Performance) Pre-compile regex patterns as a constant.

Per above comment, though, I'm not sure we should be parsing here.

joshpalis (commented):
initial pass

src/main/java/org/opensearch/flowframework/util/JsonToJsonTransformer.java
```diff
@@ -0,0 +1,388 @@
...
+ * Call Flow Architecture:
+ * transform()
+ *   └── buildOutputFromMapping() [validates object type and processes properties]
+ *       ├── processSimpleProperty() [handles JsonPath leaf nodes]
+ *       ├── [nested objects] → buildOutputFromMapping() [direct recursion]
+ *       └── processArrayProperty() [handles array syntax and builds elements]
+ *           └── buildOutputFromMapping() [direct recursion for array elements]
+ */
+public final class JsonToJsonTransformer {
+    private static final int MAX_JSON_SIZE = XContentContraints.DEFAULT_MAX_STRING_LEN; // 50 MB
```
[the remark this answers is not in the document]
Chris-tyt: Good suggestion! I'll do it in both PRs

src/main/java/org/opensearch/flowframework/util/JsonToJsonTransformer.java
Chris-tyt: Thanks, that makes a lot of sense — there's indeed no special handling needed here. I've removed the catch blocks so the Exception is now allowed to propagate naturally.

src/main/java/org/opensearch/flowframework/util/JsonToJsonTransformer.java
```diff
@@ -0,0 +1,388 @@
...
+     * @throws JsonProcessingException if either input cannot be parsed or the
+     *                                 mapping is malformed
+     */
+    public static String transform(String inputJson, String mappingRules) throws IllegalArgumentException, JsonProcessingException {
+        if (inputJson == null) {
+            throw new IllegalArgumentException("inputJson must not be null");
+        }
+        if (mappingRules == null) {
+            throw new IllegalArgumentException("JsonPath must not be null");
+        }
```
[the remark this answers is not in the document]
Chris-tyt: Modified

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 
 ## [Unreleased 3.1](https://github.com/opensearch-project/flow-framework/compare/3.0...HEAD)
 ### Features
+- Add JsonToJson Transformer as a utility function ([#1167](https://github.com/opensearch-project/flow-framework/issues/1167)) ([#1176](https://github.com/opensearch-project/flow-framework/pull/1176))
```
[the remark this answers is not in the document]
Chris-tyt: Added!

src/main/java/org/opensearch/flowframework/util/JsonToJsonTransformer.java
Chris-tyt: That makes sense! Just return a simple `string`.
All values compared in unit tests are converted to JsonNode.

src/main/java/org/opensearch/flowframework/util/JsonToJsonTransformer.java
```diff
@@ -0,0 +1,384 @@
...
+        if (Strings.isNullOrEmpty(mappingRules)) {
+            throw new IllegalArgumentException("mappingRules must not be null or empty");
+        }
+
+        JsonNode mappingNode = MAPPER.readTree(mappingRules);
+
+        Object document;
+        try {
+            document = JsonPath.using(JSON_PATH_CFG).parse(inputJson).json();
```
joshpalis, line 128: The input json is already a string, we can use .jsonString() which returns a string rather than .json() which is not strongly typed.

src/main/java/org/opensearch/flowframework/util/JsonToJsonTransformer.java
```diff
@@ -0,0 +1,384 @@
...
+     * TERMINATION: No recursion - pure value extraction and assignment.
+     *
+     * @param outputNode the target object
+     * @param propertyKey the property name
+     * @param jsonPathNode the JsonPath expression node
+     * @param inputDocument the source document
+     */
+    private static void processSimpleProperty(ObjectNode outputNode, String propertyKey, JsonNode jsonPathNode, Object inputDocument) {
+        String jsonPath = jsonPathNode.asText();
+        List<?> extracted = JsonPath.using(JSON_PATH_CFG).parse(inputDocument).read(jsonPath);
```
joshpalis, line 200: Additionally we can expect the input document here to be of type string as well

joshpalis (commented):
Second pass, overall looks good to me. Just a few comments regarding handling of the inputDocument that we're parsing

PR #1200, opened 2025-07-30
Avoid race condition setting encryption key

dbwiddis:
### Description

If the first two templates created are created simultaneously (in different threads) they may both attempt to initialize the encryption key.  Without the `.overwriteIfExists(false)` setting on the `PutDataObjectRequest`, it is possible for the second thread to find the key doesn't exist, but then attempt to put its own new key _after_ the first thread has just stored it.  This causes the template created by the first thread to fail decryption as it does not have the correct key.

This fixes the issue by adding `.overwriteIfExists(false)` which will produce a RestStatus.CONFLICT in the case of this race condition; in which case we just try to GET the key again since we know it now exists.

### Related Issues

Related to https://github.com/opensearch-project/opensearch-remote-metadata-sdk/pull/228

### Check List


owaiskazi19 (approved):
Nice one!

PR #1209, opened 2025-08-11
[Backport 2.19] Always succeed hasIndex checks with multiTenancy enabled (#1208)

dbwiddis:
Backport 76b54dc5836378710b24198ac602410dffd382fb from #1208


dbwiddis:
looks like I need to move the checkstyle to run on JDK21

PR #1251, opened 2025-10-13
Onboards flow-framework plugin to resource-sharing and access control framework

DarshitChanpura:
### Description
Implements resource-access-control for workflow and workflow_state.

### Related Issues
- Resolves https://github.com/opensearch-project/flow-framework/issues/1250

### Check List

### DIFF
--- a/src/test/java/org/opensearch/flowframework/rest/FlowFrameworkSecureRestApiIT.java
+++ b/src/test/java/org/opensearch/flowframework/rest/FlowFrameworkSecureRestApiIT.java
@@ -431,17 +431,20 @@ public class FlowFrameworkSecureRestApiIT extends FlowFrameworkRestTestCase {
         response = deleteWorkflow(aliceClient, workflowId);
         assertEquals(RestStatus.OK, TestHelpers.restStatus(response));
 
-        // Invoke status API with failure
-        ResponseException exception = expectThrows(ResponseException.class, () -> getWorkflowStatus(aliceClient, workflowId, false));
-        if (isResourceSharingFeatureEnabled()) {
-            // sample log message: No sharing info found for 'pmaAdZoBL3PmSyJoha0C'. Action
-            // cluster:admin/opensearch/flow_framework/workflow_state/get is not allowed.
-            // since record is deleted corresponding sharing record will also be deleted and thus we show 403 for non-existent sharing
-            // records,
-            assertEquals(RestStatus.FORBIDDEN.getStatus(), exception.getResponse().getStatusLine().getStatusCode());
-        } else {
-            assertEquals(RestStatus.NOT_FOUND.getStatus(), exception.getResponse().getStatusLine().getStatusCode());
-        }
+        // Wait for state doc to be deleted asynchronously
+        assertBusy(() -> {
+            // Invoke status API with failure
+            ResponseException exception = expectThrows(ResponseException.class, () -> getWorkflowStatus(aliceClient, workflowId, false));
+            if (isResourceSharingFeatureEnabled()) {
+                // sample log message: No sharing info found for 'pmaAdZoBL3PmSyJoha0C'. Action
+                // cluster:admin/opensearch/flow_framework/workflow_state/get is not allowed.
+                // since record is deleted corresponding sharing record will also be deleted and thus we show 403 for non-existent sharing
+                // records,
+                assertEquals(RestStatus.FORBIDDEN.getStatus(), exception.getResponse().getStatusLine().getStatusCode());
+            } else {
+                assertEquals(RestStatus.NOT_FOUND.getStatus(), exception.getResponse().getStatusLine().getStatusCode());
+            }
+        }, 30, TimeUnit.SECONDS);
     }
 
     public void testUpdateWorkflowEnabledForAdmin() throws Exception {


build.gradle
```diff
@@ -252,6 +256,8 @@ dependencies {
     testImplementation("org.junit.jupiter:junit-jupiter:${junitJupiterVersion}")
     testImplementation("com.fasterxml.jackson.datatype:jackson-datatype-jsr310:${versions.jackson_databind}")
 
+    testImplementation 'org.awaitility:awaitility:4.3.0'
```
owaiskazi19: Constant for the version

build.gradle
```diff
@@ -538,7 +550,13 @@ testClusters.integTest {
     if (System.getProperty("opensearch.debug") != null) {
         def debugPort = 5005
         nodes.forEach { node ->
-            node.jvmArgs("-agentlib:jdwp=transport=dt_socket,server=n,suspend=y,address=*:${debugPort}")
+            // server=n,suspend=y -> node tries to connect to a debugger and hence test runs fails with
+            //            Exec output and error:
+            //            | Output for ./bin/opensearch-plugin:ERROR: transport error 202: connect failed: Connection refused
+            //            | ERROR: JDWP Transport dt_socket failed to initialize, TRANSPORT_INIT(510)
+            //                    | JDWP exit error AGENT_ERROR_TRANSPORT_INIT(197): No transports initialized [src/jdk.jdwp.agent/share/native/libjdwp/debugInit.c:700].
+            // So instead, we listen to a debugger by saying server=y and suspend=n
```
owaiskazi19: Can we remove the comments?

src/main/java/org/opensearch/flowframework/util/ParseUtils.java
```diff
@@ -251,6 +252,31 @@ public static User getUserContext(Client client) {
...
+        return client != null && client.isFeatureEnabledForType(resourceType);
+    }
+
+    /**
+     * Verifies whether the user has permission to access the resource.
+     * @param resourceType the type of resource to be authorized
+     * @param onSuccess consumer function to execute if resource sharing feature is enabled
+     * @param fallbackOn501 consumer function to execute if resource sharing feature is disabled.
+     */
+    public static void verifyResourceAccessAndProcessRequest(String resourceType, Runnable onSuccess, Runnable fallbackOn501) {
```
owaiskazi19:
```suggestion
    public static void verifyResourceAccessAndProcessRequest(String resourceType, Runnable onSuccess, Runnable onFailure) {
```

src/main/java/org/opensearch/flowframework/transport/PluginClient.java
```diff
@@ -0,0 +1,60 @@
...
+/**
+ * A special client for executing transport actions as this plugin's system subject.
+ */
+public class PluginClient extends FilterClient {
+
+    private static final Logger logger = LogManager.getLogger(PluginClient.class);
+
+    private Subject subject;
+
+    public PluginClient(Client delegate) {
```
owaiskazi19:
```suggestion
    public PluginClient(Client client) {
```

src/main/java/org/opensearch/flowframework/transport/PluginClient.java
```diff
@@ -0,0 +1,60 @@
...
+        super(delegate);
+    }
+
+    public PluginClient(Client delegate, Subject subject) {
```
owaiskazi19: Where this constructor used?

src/main/java/org/opensearch/flowframework/transport/PluginClient.java
```diff
@@ -0,0 +1,60 @@
...
+        this.subject = subject;
+    }
+
+    @Override
+    protected <Request extends ActionRequest, Response extends ActionResponse> void doExecute(
+        ActionType<Response> action,
+        Request request,
+        ActionListener<Response> listener
+    ) {
+        if (subject == null) {
```
owaiskazi19: subject would be null with the first constructor call, right?

src/main/java/org/opensearch/flowframework/util/ResourceSharingClientAccessor.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+package org.opensearch.flowframework.util;
+
+import org.opensearch.security.spi.resources.client.ResourceSharingClient;
+
+/**
+ * Accessor for resource sharing client
+ */
+public class ResourceSharingClientAccessor {
+    private ResourceSharingClient CLIENT;
```
owaiskazi19: Can make it as thread safe?
```suggestion
    private final AtomicReference<ResourceSharingClient> client = new AtomicReference<>();
```

owaiskazi19 (commented):
1st iteration

build.gradle
```diff
@@ -195,6 +196,9 @@ configurations {
 }
 
 dependencies {
+    // For resource access control
+    compileOnly group: 'org.opensearch', name:'opensearch-security-spi', version:"3.4.0-SNAPSHOT"
```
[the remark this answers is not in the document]
DarshitChanpura: yes, once main branch has the version bumped I'll revert it. 

i explicitly set it in last commit: https://github.com/opensearch-project/flow-framework/pull/1251/commits/69009e6819cba77894becab7ff8fa19dc95e0aa0

build.gradle
DarshitChanpura: ack.

build.gradle
DarshitChanpura: no, let's keep it to understand what these changes mean.

src/main/java/org/opensearch/flowframework/util/ParseUtils.java
DarshitChanpura: it is not on failure. it will only fallback when resource-sharing framework is not used.

src/main/java/org/opensearch/flowframework/transport/PluginClient.java
DarshitChanpura: we don't expect subject to be null, this is added to ensure robustness.

src/main/java/org/opensearch/flowframework/util/ParseUtils.java
```diff
@@ -251,6 +252,31 @@ public static User getUserContext(Client client) {
...
+
+    /**
+     * Verifies whether the user has permission to access the resource.
+     * @param resourceType the type of resource to be authorized
+     * @param onSuccess consumer function to execute if resource sharing feature is enabled
+     * @param fallBackIfDisabled consumer function to execute if resource sharing feature is disabled.
+     */
+    public static void verifyResourceAccessAndProcessRequest(String resourceType, Runnable onSuccess, Runnable fallBackIfDisabled) {
+        // Resource access will be auto-evaluated
+        if (shouldUseResourceAuthz(resourceType)) {
```
owaiskazi19: Should be under a try/catch block here

src/main/java/org/opensearch/flowframework/util/ResourceSharingClientAccessor.java
```diff
@@ -0,0 +1,46 @@
...
+        }
+
+        return resourceSharingClientAccessor;
+    }
+
+    /**
+     * Set the resource sharing client
+     */
+    public void setResourceSharingClient(ResourceSharingClient client) {
+        resourceSharingClientAccessor.client.set(client);
```
owaiskazi19: no need to reference static field here
```suggestion
        this.client.set(client);
```

src/main/java/org/opensearch/flowframework/util/ResourceSharingClientAccessor.java
```diff
@@ -0,0 +1,46 @@
...
+
+import java.util.concurrent.atomic.AtomicReference;
+
+/**
+ * Accessor for resource sharing client
+ */
+public class ResourceSharingClientAccessor {
+    private final AtomicReference<ResourceSharingClient> client = new AtomicReference<>();
+
+    private static ResourceSharingClientAccessor resourceSharingClientAccessor;
```
owaiskazi19: For thread safety
```suggestion
    private static final ResourceSharingClientAccessor INSTANCE = new ResourceSharingClientAccessor();
```

src/main/java/org/opensearch/flowframework/util/ResourceSharingClientAccessor.java
```diff
@@ -0,0 +1,46 @@
...
+
+    private ResourceSharingClientAccessor() {}
+
+    public static ResourceSharingClientAccessor getInstance() {
+        if (resourceSharingClientAccessor == null) {
+            resourceSharingClientAccessor = new ResourceSharingClientAccessor();
+        }
+
+        return resourceSharingClientAccessor;
```
owaiskazi19: Now we can just return
```suggestion
        return INSTANCE;
```

owaiskazi19 (commented):
2nd iteration

build.gradle
```diff
@@ -251,6 +256,8 @@ dependencies {
     testImplementation("org.junit.jupiter:junit-jupiter:${junitJupiterVersion}")
     testImplementation("com.fasterxml.jackson.datatype:jackson-datatype-jsr310:${versions.jackson_databind}")
 
+    testImplementation "org.awaitility:awaitility:${awaitlityVersion}"
```
owaiskazi19:
```suggestion
    testImplementation "org.awaitility:awaitility:${awaitilityVersion}"
```

src/main/java/org/opensearch/flowframework/util/ParseUtils.java
```diff
@@ -251,6 +252,31 @@ public static User getUserContext(Client client) {
...
     }
 
+    /**
+     * Checks whether resource authz framework should be used.
+     * @param resourceType the type to check for resource authz
+     * @return true if resource-authz should be used, false otherwise
+     */
+    public static boolean shouldUseResourceAuthz(String resourceType) {
+        var client = ResourceSharingClientAccessor.getInstance().getResourceSharingClient();
+        return client != null && client.isFeatureEnabledForType(resourceType);
```
dbwiddis: Compilation is failing here:
```
> Compilation failed; see the compiler output below.
  Note: Recompile with -Xlint:unchecked for details.
  Note: Some input files use unchecked or unsafe operations.
  /Users/widdisd/git/flow-framework/src/main/java/org/opensearch/flowframework/util/ParseUtils.java:262: error: cannot find symbol
          return client != null && client.isFeatureEnabledForType(resourceType);
                                         ^
    symbol:   method isFeatureEnabledForType(String)
    location: variable client of type ResourceSharingClient
  1 error
```
I see above your comment that
> compilation may fail if security plugin is absent and by extension SPI is absent as well.
(because of the import statement)

but this seems unrelated.  The `ResourceSharingClient` returned by that call does not have an `isFeatureEnabledForType` method.

Possibly related, integ tests are failing because the SPI is apparently used for ML classes:
```
»  fatal error in thread [main], exiting
»  java.lang.InstantiationError: org.opensearch.security.spi.resources.ResourceProvider
»  	at org.opensearch.ml.resources.MLResourceSharingExtension.getResourceProviders(MLResourceSharingExtension.java:22)
»  	at org.opensearch.security.resources.ResourcePluginInfo.setResourceSharingExtensions(ResourcePluginInfo.java:76)
»  	at org.opensearch.security.OpenSearchSecurityPlugin.loadExtensions(OpenSearchSecurityPlugin.java:2433)
»  	at org.opensearch.plugins.PluginsService.loadExtensionsForPlugin(PluginsService.java:647)
»  	at org.opensearch.plugins.PluginsService.loadExtensions(PluginsService.java:627)
»  	at org.opensearch.plugins.PluginsService.<init>(PluginsService.java:229)
»  	at org.opensearch.node.Node.<init>(Node.java:555)
»  	at org.opensearch.node.Node.<init>(Node.java:483)
»  	at org.opensearch.bootstrap.Bootstrap$5.<init>(Bootstrap.java:250)
»  	at org.opensearch.bootstrap.Bootstrap.setup(Bootstrap.java:250)
»  	at org.opensearch.bootstrap.Bootstrap.init(Bootstrap.java:412)
»  	at org.opensearch.bootstrap.OpenSearch.init(OpenSearch.java:168)
»  	at org.opensearch.bootstrap.OpenSearch.execute(OpenSearch.java:159)
»  	at org.opensearch.common.cli.EnvironmentAwareCommand.execute(EnvironmentAwareCommand.java:110)
»  	at org.opensearch.cli.Command.mainWithoutErrorHandling(Command.java:138)
»  	at org.opensearch.cli.Command.main(Command.java:101)
»  	at org.opensearch.bootstrap.OpenSearch.main(OpenSearch.java:125)
»  	at org.opensearch.bootstrap.OpenSearch.main(OpenSearch.java:91)
```

dbwiddis: Seems this depends on https://github.com/opensearch-project/ml-commons/pull/4402

cwperks: There was a recent refactor in https://github.com/opensearch-project/security/pull/5755 to an interface. This will need a new commit to switch this to an anonymous class that implements the interface.

dbwiddis:
```
REPRODUCE WITH: ./gradlew ':integTest' <stuff> -Dtests.security.manager=false 

FlowFrameworkResourceSharingRestApiIT > testProvisionDeprovision_withResourceSharingEnabled FAILED

    java.lang.IllegalArgumentException: Secure Tests are running but HTTPS is not set
```
How do I turn off "secure tests" or are they always secure now?

build.gradle
```diff
@@ -193,6 +195,9 @@ configurations {
 }
 
 dependencies {
+    // For resource access control
+    compileOnly group: 'org.opensearch', name:'opensearch-security-spi', version:"${opensearch_build}"
```
dbwiddis: This format is deprecated and will fail in Gradle 10.   Please match the other declarations in this build.

See https://docs.gradle.org/9.2.0/userguide/upgrading_version_9.html#dependency_multi_string_notation

src/test/java/org/opensearch/flowframework/rest/FlowFrameworkResourceSharingRestApiIT.java
```diff
@@ -0,0 +1,403 @@
...
+    private static final String WORKFLOW_FULL_ACCESS_AG = "workflow_full_access";
+    private static final String WORKFLOW_STATE_READ_ONLY_AG = "workflow_state_read_only";
+    private static final String WORKFLOW_STATE_FULL_ACCESS_AG = "workflow_state_full_access";
+
+    @Before
+    public void setupSecureTests() throws IOException {
+        if (!isHttps()) throw new IllegalArgumentException("Secure Tests are running but HTTPS is not set");
+
+        // If the suite is launched without the flag, just skip these tests cleanly.
+        assumeTrue("RS tests only run when resource_sharing.enabled=true", isResourceSharingFeatureEnabled());
```
dbwiddis: Probably need to reverse the order of these statements so we don't throw IAE when resource sharing isn't enabled, which is currently failing all integ tests.

src/main/java/org/opensearch/flowframework/util/ParseUtils.java
DarshitChanpura: we don't need any all exceptions must be thrown up to the caller.

src/main/java/org/opensearch/flowframework/util/ParseUtils.java
DarshitChanpura: `./gradlew integTest` seems to be working fine for me locally.

src/main/java/org/opensearch/flowframework/util/PluginClient.java
```diff
@@ -0,0 +1,55 @@
...
+        Request request,
+        ActionListener<Response> listener
+    ) {
+        if (subject == null) {
+            throw new IllegalStateException("PluginClient is not initialized.");
+        }
+        try (ThreadContext.StoredContext ctx = threadPool().getThreadContext().newStoredContext(false)) {
+            subject.runAs(() -> {
+                logger.info("Running transport action with subject: {}", subject.getPrincipal().getName());
+                super.doExecute(action, request, ActionListener.runBefore(listener, ctx::restore));
```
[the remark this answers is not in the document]
DarshitChanpura: is it because of using newStoredContext?

src/main/java/org/opensearch/flowframework/util/ParseUtils.java
dbwiddis: > `./gradlew integTest` seems to be working fine for me locally.

Yes, earlier you were throwing the exception before testing for the variable to skip the tests.

dbwiddis:
> i have consistently been able to see success for the test run after addressing the fix.

Nice... unit tests failing on a class cast. Check the argument index if you've added parameters to a method, fixed a few previous tests where action listener was arg 1 and moved to 3...

dbwiddis (approved):

src/main/java/org/opensearch/flowframework/util/PluginClient.java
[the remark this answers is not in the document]
owaiskazi19: it's using try-with-resources so the StoredContext is closed (restored) as soon as the try block exits. Because super.doExecute schedules work asynchronously, closing the context at the end of the try block is too early. Let's address this with the changes suggested

.github/workflows/test_security.yml
```diff
@@ -20,6 +20,7 @@ jobs:
     strategy:
       matrix:
         java: [21]
+        resource_sharing_flag: [ "", "-Dresource_sharing.enabled=true" ]
```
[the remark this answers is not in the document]
owaiskazi19: Will prefer the above cleaner approach here

DarshitChanpura:
> @DarshitChanpura this test is failing again, thought you had fixed it previously :)

I'm no longer able to reproduce this locally so not sure what is going on here. But it is flaky.

gaiksaya:
> @gaiksaya can we update our required tests to include the "true" here?

Done!

PR #1270, opened 2025-11-11
[BUG] Incorrect field map output dimensions in default values for semantic search with local model use case template

### LINKED ISSUE #1254: [BUG] Incorrect field map output dimensions in default values for semantic search with local model use case template
### What is the bug?
The default values in "semantic-search-with-local-model-defaults.json" set the "text_embedding.field_map.output.dimension" to 768. However the model only outputs 384 dimensions. This causes an error in the response when uploading documents to the vector index because the connector expects 768 dimensions, but it is only provided with 384. 

1. Execute the semantic search with local model flow template: `POST /_plugins/_flow_framework/workflow?use_case=semantic_search_with_local_model&provision=true`
2. Try and upload a document to the newly created index
3. You will get an error in the response saying 

```
"caused_by": {
            "type": "illegal_argument_exception",
            "reason": "Vector dimension mismatch. Expected: 768, Given: 384"
```

I believe this is happening because the default value in the flow framework (https://github.com/opensearch-project/flow-framework/blob/main/src/main/resources/defaults/semantic-search-with-local-model-defaults.json) is setting the dimensions feild to 768, when the model only outputs 384. 

I believe this is the root cause of the issue because if you override the dimensions field and set the dimensions to 384, it works as expected. 

The error should not occur. 

### What is your host/environment?
Local - OpenSearch 3.3

akolarkunnu:
### Description
According to documentation in https://huggingface.co/sentence-transformers/paraphrase-MiniLM-L3-v2 , output dimension is 384 and it is working fine with 384.
Tested the behavior and it resolved the issue.

### Related Issues
Resolves #1254

### Check List
- [X] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 ### Enhancements
 ### Bug Fixes
 - Pre-create ML Commons indices for Tenant Aware tests ([#1217](https://github.com/opensearch-project/flow-framework/pull/1217))
+- Incorrect field map output dimensions in default values for semantic search with local model use case template ([#1270](https://github.com/opensearch-project/flow-framework/pull/1270))
 
 ### Infrastructure
 ### Documentation


dbwiddis (approved):
Thanks for the fix. Can you please add a change log entry?

Issue #1201, opened 2025-07-30
[BUG]


dbwiddis:
### What is the bug?

Backend role filtering doesn't handle the case where a template is created/provisioned with an admin (null) user, but accessed with a user with backend roles.

### How can one reproduce the bug?

1. Create a template using a superuser.
2. Observe the `user` field is null in that template.
3. Attempt to edit the template with a user with backend roles.
4. Get a "Failed to parse workflow" error from an NPE.

### What is the expected behavior?

Pick one:
 - User with roles can access admin templates :)
 - User with roles can't access admin templates :(

### Do you have any additional context?

This line is where the logic is undefined:
https://github.com/opensearch-project/flow-framework/blob/6fbe08ca721eb0cc22fe12c1fc0150eb6e7b05fa/src/main/java/org/opensearch/flowframework/util/ParseUtils.java#L507

`checkUserPermissions()` will throw an NPE if either `requestUser` or `resourceUser` is null.  The `requestUser` is effectively null-checked with the `isAdmin(requestUser)` check, but it should be moved to be executed prior to the `checkUserPermissions()` call.  However, this still doesn't address what we should do in the case that the template (or state) doesn't contain a user, when we actually have a user.

Easily addressed with a null check in `checkUserPermissions` but unclear whether `true` or `false` is an appropriate return.

500ocean700-a11y:
Hi is this a good first issue for me

dbwiddis:
> Hi is this a good first issue for me

Hi, @500ocean700-a11y thanks, but @Avinash1423 is already working on it.  There are a few other good issues to work on, though!

PR #1292, opened 2025-12-23
Fixes null Check  where a template is created  with an admin (null) user

Avinash1423:
Fix for #1201

Discription:

Added null Check for checkUserPermissions and moved isAdmin() to be executed prior to the checkUserPermissions().

### DIFF
--- a/src/test/java/org/opensearch/flowframework/util/ParseUtilsTests.java
+++ b/src/test/java/org/opensearch/flowframework/util/ParseUtilsTests.java
@@ -460,4 +460,20 @@ public class ParseUtilsTests extends OpenSearchTestCase {
         assertFalse(ParseUtils.checkUserPermissions(userWithNullRoles, validUser, workflowId));
     }
 
+    public void testCheckUserPermissionsWithMatchingRoles() throws Exception {
+        String workflowId = "testWorkflowId";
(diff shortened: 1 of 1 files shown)


src/main/java/org/opensearch/flowframework/util/ParseUtils.java
```diff
@@ -375,6 +375,9 @@ public static void resolveUserAndExecute(
      * @throws Exception exception
      */
     private static boolean checkUserPermissions(User requestedUser, User resourceUser, String workflowId) throws Exception {
+        if (requestedUser == null || resourceUser == null) {
+            return false;
+        }
         if (resourceUser.getBackendRoles() == null || requestedUser.getBackendRoles() == null) {
             return false;
         }
```
dbwiddis: These lines can probably be combined into one conditional, or maybe we can create a helper method taking varargs `User...` that we can iterate over and just call `if (nullOrNoRoles(requestedUser, resourceUsers)) { return false; }`.

dbwiddis (commented):
Great progress, see comments below.

Avinash1423:
@dbwiddis 
sorry ,was busy right after the new year did not have time to look into the tests until now. Anyways glad u merged it.

Issue #535, opened 2024-02-22
[REFACTOR] Eliminate duplication of required input / output List/Set


dbwiddis:
### Is your feature request related to a problem?

The [WorkflowValidatorTests](https://github.com/opensearch-project/flow-framework/blob/main/src/test/java/org/opensearch/flowframework/model/WorkflowValidatorTests.java) class contains a lot of repetitive code that could be refactored to be more general.

This test class was originally written to validate the JSON file.  Following #523 we now have an Enum and easier access to programmatically perform these tests, so it can be simplified.

### What solution would you like?

The manual creation of a map [here](https://github.com/opensearch-project/flow-framework/blob/24bf51a69f413a0d6e43c3a9d9c8912e9890bd26/src/test/java/org/opensearch/flowframework/model/WorkflowValidatorTests.java#L40-L96) can be reduced to a single statement, which I've done as part of #530 .

The Input/Output size checks [here](https://github.com/opensearch-project/flow-framework/blob/24bf51a69f413a0d6e43c3a9d9c8912e9890bd26/src/test/java/org/opensearch/flowframework/model/WorkflowValidatorTests.java#L102-L156) could be improved.

Taking one entry as an example:
```java
assertTrue(validator.getWorkflowStepValidators().keySet().contains("create_connector"));
assertEquals(7, validator.getWorkflowStepValidators().get("create_connector").getInputs().size());
assertEquals(1, validator.getWorkflowStepValidators().get("create_connector").getOutputs().size());
```
1. The repeated string "create_connector" corresponds to the class name and should be replaced by `CreateConnectorStep.NAME`.
2. There's no need to check the presence of the key in the keyset, the next line will throw an NPE when `get()` returns null and we try to get inputs or outputs.
3. We should refactor to avoid duplication of the required inputs.  Currently each workflow step defines its required inputs inline in the code (as a `Set<String>`) and separately in the `WorkflowStepValidator` class (as a `List<String>`).  This duplication should be removed by making the collection a class field with a getter similarly to how `NAME` is declared.
4. We don't currently define the outputs in the `WorkflowStep` implementation classes, but we should do so to keep that declaration closer to where we actually populate the `WorkflowData` when the step is complete.

### What alternatives have you considered?

Leave the code as-is and keep adding to it.

### Do you have any additional context?

https://en.wikipedia.org/wiki/Don%27t_repeat_yourself

ajayl83:
@dbwiddis can i pick this issue?

[...]

PR #1299, opened 2026-01-13
refactor: Add constants to CreateConnectorStep and update enum usage #535

kokibas:
### Description
Add constants to CreateConnectorStep and update enum usage

### Related Issues
Part of #535 

### Check List
-.- Add REQUIRED_INPUTS, OPTIONAL_INPUTS, and PROVIDED_OUTPUTS constants to CreateConnectorStep
- Update WorkflowSteps.CREATE_CONNECTOR enum to use constants from CreateConnectorStep class
- Refactor WorkflowStepValidatorTests to use constants instead of magic strings
- Remove code duplication between CreateConnectorStep and WorkflowSteps enum


dbwiddis (commented):
Code LGTM!

Some formatting issues (see below).

Also while the Issue describes the create connector as an example, this can actually be done for many of the steps.  Feel free to focus just on the connector step for this PR but we shouldn't "resolve" the issue just yet, several more left to do!

PR #1308, opened 2026-01-28
Refactor RegisterModelGroupStep & ReIndexStep to use const 

kokibas:
### Description
Refactor RegisterModelGroupStep & ReIndexStep to use const 

### Related Issues
#535 

This change is internal and does not require a changelog entry.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 ### Enhancements
 ### Bug Fixes
 - Fix substitution templates for OpenSearch 3.x compatibility by removing `_doc` mapping ([#1301](https://github.com/opensearch-project/flow-framework/pull/1301))
+- Fixed user permission validation to correctly handle null users and backend roles ([#1292](https://github.com/opensearch-project/flow-framework/pull/1292))
 
 ### Infrastructure
 ### Documentation
--- a/build.gradle
+++ b/build.gradle
@@ -61,7 +61,7 @@ buildscript {
 
     dependencies {
         classpath "org.opensearch.gradle:build-tools:${opensearch_version}"
-        classpath "com.diffplug.spotless:spotless-plugin-gradle:8.2.0"
+        classpath "com.diffplug.spotless:spotless-plugin-gradle:8.2.1"
     }
 }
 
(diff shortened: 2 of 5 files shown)


dbwiddis:
> Is this behavior expected, or would it make sense to introduce an explicit timeout or failure path for state index updates to prevent workflow steps from hanging?

I'm traveling but I did ask Q to evaluate your concern and it seems valid.  Please address if you can and thanks for the sharp eyes!

In [RegisterModelGroupStep.java](https://github.com/opensearch-project/flow-framework/tree/main/src/main/java/org/opensearch/flowframework/workflow/RegisterModelGroupStep.java#L70-L87), when mlClient.registerModelGroup succeeds, the onResponse callback calls flowFrameworkIndicesHandler.addResourceToStateIndex() with a resourceListener.

Looking at [FlowFrameworkIndicesHandler.addResourceToStateIndex](fhttps://github.com/opensearch-project/flow-framework/tree/main/src/main/java/org/opensearch/flowframework/indices/FlowFrameworkIndicesHandler.java#L755-L778):

If validateStateIndexExists returns false, the listener is called with onFailure ✓
Otherwise, it calls getAndUpdateResourceInStateDocumentWithRetries which uses sdkClient.getDataObjectAsync().whenComplete()
The async chain in [getAndUpdateResourceInStateDocumentWithRetries](https://github.com/opensearch-project/flow-framework/tree/main/src/main/java/org/opensearch/flowframework/indices/FlowFrameworkIndicesHandler.java#L810-L831) properly handles:

Success path → calls handleStateGetResponse
Failure path → calls handleStateUpdateException
However, there's a potential issue: the CompletableFuture.whenComplete() callback relies on the underlying async operation completing. If:

The SDK client's async operation never completes (network partition, hung connection, etc.)
No timeout is configured at the SDK client level
Then the PlainActionFuture in RegisterModelGroupStep would never be resolved.

Summary:

The concern is valid. The code correctly handles success/failure cases when the async operation completes, but there's no explicit timeout to handle cases where the async operation hangs indefinitely.
The retry logic (5 retries) only handles version conflicts, not timeouts.
If sdkClient.getDataObjectAsync() or sdkClient.updateDataObjectAsync() never completes, the workflow step hangs.
Recommendation: Consider adding:

A timeout wrapper around the CompletableFuture operations (e.g., orTimeout() in Java 9+)
Or configure timeouts at the SDK client level
Or wrap the listener with a timeout mechanism at the workflow step level

